PR #17, opened 2021-12-13
Modify the indentation method to eliminate compilation warnings

AndyQinshy:
Modify the indentation method to eliminate compilation warnings

### DIFF
--- a/contrib/orafce/datefuncs.c
+++ b/contrib/orafce/datefuncs.c
@@ -708,7 +708,7 @@ oraadd_months(PG_FUNCTION_ARGS)
 	y = v.quot;
 	if (y < 0)
 	{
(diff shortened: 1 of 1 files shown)


BingAo-Office:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89549

I think it's gcc’s bug. we can avoid it by add compile option "-Wno-misleading-indentation"

rasifr (approved):
fixed code indentation properly.

rasifr:
> https://gcc.gnu.org/bugzilla/show_bug.cgi?id=89549
> 
> I think it's gcc’s bug. we can avoid it by add compile option "-Wno-misleading-indentation" At least it doesn't feel very practical. In addition, github code comparison does not show or indicate such problems.
> 
It's okay to not add this option. If there are any warnings we can proactively resolve and fix them properly.

PR #19, opened 2021-12-14
compatible oracle string function

yanxuebiao-code:


contrib/orafce/charpad.c
```diff
@@ -287,17 +320,47 @@ orafce_rpad(PG_FUNCTION_ARGS)
...
+
+	nargs = PG_NARGS();
+	if (nargs < 2)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				errmsg("too few args.")));
+	if (nargs > 3)
+		ereport(ERROR,
+				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+				errmsg("more args than function needed.")));
```
rasifr, line 334: ERRCODE_TOO_MANY_ARGUMENTS is more appropriate here. and the error message can also like this.
```suggestion
	if (nargs > 3)
 		ereport(ERROR,
				(errcode(ERRCODE_TOO_MANY_ARGUMENTS),
				 errmsg("too many arguments")));
```

contrib/orafce/charpad.c
```diff
@@ -57,17 +58,49 @@ orafce_lpad(PG_FUNCTION_ARGS)
...
+
+	if (nargs == 3)
+	{
+		argoid3 = get_fn_expr_argtype(fcinfo->flinfo, 2);
+		string2 =  (text *)orafce_sourcetype_to_targetype(PG_GETARG_DATUM(2), argoid3, TEXTOID);
+	}
+	else
+		string2 = cstring_to_text(" ");
+
+	s2blen = VARSIZE_ANY_EXHDR(string2);
```
rasifr, line 96: I think it's better to initialise 'string2' at declaration and have 's2blen' conditional instead of assigning a empty string just for this.

```suggestion
string2 = NULL;
...
if (string2)
	s2blen = VARSIZE_ANY_EXHDR(string2);
```

contrib/orafce/charpad.c
rasifr, line 334: same as previously pointed out.

BingAo-Office: It has been modified. Thank you for your review

rasifr:
committed. Also applied to master.

PR #20, opened 2021-12-16
Let the log collector collect the DDL statements of the packages.

lk18347265415:
when logging_collector = on and log_statement = ddl,There will throw out WARING when create package.And the package statements can't be writen to log file.


rasifr (approved):
enum entry was missing for package, handling log_statement and this patch fixes that.

PR #31, opened 2021-12-29
Add new feature,there is no need to set SEARCH_PATH after switching to oralce mode

lk18347265415:
1. delete set/show  SEARCHPATH in orafce.
2. we can use oracle compatible functions directly after switching to oracle mode.


BingAo-Office (approved):

codeforall:
I have a couple of comments on this pull request.

1. It is moving the orafce functions from pg_catalog to oracle schema without updating the extension version. So effectively the same version of orafce extension will have its objects in the different schema on IvorySQL 1.0 and 1.1, which doesn't seem like a correct way to go. So IMHO if we want to make changes to orafce extension interface then we should also consider bumping the extension version.
2. pull request is implicitly adding oracle schema to search_path without checking if the oracle schema is already in the search_path. That can lead to multiple entries of oracle schema in the search_path list when the user has explicitly added 'oracle' in search_path

lk18347265415:
>

lk18347265415:
> 

@codeforall thanks for your  feedback.for your first question,we may update the version when it is released later.For your second question,if we switch to the compatible oralce mode,we don't care whether the oralce schema exists in the search_path,we just put the oralce schema at the head of the search_path.

codeforall:
> > 
> 
> @codeforall thanks for your feedback.for your first question,we may update the version when it is released later.For your second question,if we switch to the compatible oralce mode,we don't care whether the oralce schema exists in the search_path,we just put the oralce schema at the head of the search_path.

Thanks, @lk18347265415, I think bumping the extension version should also be made part of the pull request as it will allow the reviewer to verify if the extension upgrade funcrtionality.

For the second point, IMHO if the oracle schema is already present in search_path (explicitly added by user), we should leave the search_path as it is without adding the second entry for oracle schema. First, because it doesn't feel right to have two entries of the same schema in search_patch, Secondly the current implementation will silently override the user's intended precedence for search path. PostgreSQL also follows the same approach for pg_catalog schema and only adds it to the top of search_patch if it's not explicitly specified by the user.

BingAo-Office:
> 

> For the second point, IMHO if the oracle schema is already present in search_path (explicitly added by user), we should leave 

> the search_path as it is without adding the second entry for oracle schema. First, because it doesn't feel right to have two 

> entries of the same schema in search_patch, Secondly the current implementation will silently override the user's intended 

> precedence for search path. PostgreSQL also follows the same approach for pg_catalog schema and only adds it to the top of 

> search_patch if it's not explicitly specified by the user.

but ， user want to use oracle date, like this "create table test( c date )" not like "create table test( c oracle.date ); ".   we just have this solution so far.  could you give our a better solution? @codeforall

codeforall:
> > 
> 
> > For the second point, IMHO if the oracle schema is already present in search_path (explicitly added by user), we should leave
> 
> > the search_path as it is without adding the second entry for oracle schema. First, because it doesn't feel right to have two
> 
> > entries of the same schema in search_patch, Secondly the current implementation will silently override the user's intended
> 
> > precedence for search path. PostgreSQL also follows the same approach for pg_catalog schema and only adds it to the top of
> 
> > search_patch if it's not explicitly specified by the user.
> 
> but ， user want to use oracle date, like this "create table test( c date )" not like "create table test( c oracle.date ); ". we just have this solution so far. could you give our a better solution? @codeforall

I am not questioning the solution. I think the approach you followed is correct. I am just suggesting a small change in your patch to use list_member_oid instead of linitial_oid before adding oracle schema oid to the search path.

in  src/backend/catalog/namespace.c (line-no 4123)
```c
4116     /*
4117      *oracle namespace reference,for compatible
4118      */
4119     if (compatible_db == COMPATIBLE_ORA)
4120     {
4121         compatibleNS = get_namespace_oid("oracle", true);
4122         if (OidIsValid(compatibleNS))
4123             if (linitial_oid(oidlist) != compatibleNS)
4124                 oidlist = lcons_oid(compatibleNS, oidlist);
4125     }
```

```c
4116     /*
4117      *oracle namespace reference,for compatible
4118      */
4119     if (compatible_db == COMPATIBLE_ORA)
4120     {
4121         compatibleNS = get_namespace_oid("oracle", true);
4122         if (OidIsValid(compatibleNS))
4123             if (!list_member_oid(oidlist, compatibleNS))
4124                 oidlist = lcons_oid(compatibleNS, oidlist);
4125     }

```
Do you think this suggestion will cause some unwanted behavior?

BingAo-Office:
> Do you think this suggestion will cause some unwanted behavior?


Indeed, there may be, but as long as there is no exception override for objects under PG_CATALOG, the remaining problem is the same as pg_CATALOG as the first schema.

If objects in the user's own schema are overridden. So, he should not work in oracle mode, he can find another compatible mode。

In addition, if you are a normal user, you cannot create any objects in oracle schema, because only GRANT USEAGE TO PUBLIC. IF you are superuser,That sorry.  But superuser is the dba. There was no problem with security,  If someone wants to inject through a sql, they can't

KingZhaoTPG:
> `I am not questioning the solution. I think the approach you followed is correct. I am just suggesting a small change in your patch to use list_member_oid instead of linitial_oid before adding oracle schema oid to the search path.

in src/backend/catalog/namespace.c (line-no 4123)

4116     /*
4117      *oracle namespace reference,for compatible
4118      */
4119     if (compatible_db == COMPATIBLE_ORA)
4120     {
4121         compatibleNS = get_namespace_oid("oracle", true);
4122         if (OidIsValid(compatibleNS))
4123             if (linitial_oid(oidlist) != compatibleNS)
4124                 oidlist = lcons_oid(compatibleNS, oidlist);
4125     }
4116     /*
4117      *oracle namespace reference,for compatible
4118      */
4119     if (compatible_db == COMPATIBLE_ORA)
4120     {
4121         compatibleNS = get_namespace_oid("oracle", true);
4122         if (OidIsValid(compatibleNS))
4123             if (!list_member_oid(oidlist, compatibleNS))
4124                 oidlist = lcons_oid(compatibleNS, oidlist);
4125     }`

Hi @codeforall , thanks for your comments.  our solution is to ensure that the oracle mode is added to the top of search_path. so, when you set compatible_mode to oracle; you can directly use Oracle-compatible functions and types without oracle.xxx to qualify. if you want to use PG's original functions. you just need to change the compatible_mode to postgres. :)

PR #35, opened 2022-01-04
support oracle number(p,s) type

BingAo-Office:


rasifr (changes requested):
I think there are some issue still. After applying the patch, I have observed the following.

1. see below query, It should be able to execute without issues.

> postgres=# select 1/2::numeric;
ERROR:  operator is only a shell: integer / numeric
LINE 1: select 1/2::numeric;
                ^
postgres=# 

2. even after setting the compatibility mode, the error is still there. 

> postgres=# set compatible_mode TO oracle ;
SET
postgres=# select 1/2::numeric;
ERROR:  operator is only a shell: integer / numeric
LINE 1: select 1/2::numeric;
                ^
postgres=# 

3. and replacing numeric with number results in ambiguous operator selection.

> postgres=# select 1/2::number;
ERROR:  operator is not unique: integer / "number"
LINE 1: select 1/2::number;
                ^
HINT:  Could not choose a best candidate operator. You might need to add explicit type casts.
postgres=#

BingAo-Office:
Thanks for your review ,  I found many errors and redundant operators through the inspection, which led to the above patch errors, which have been repaired now

contrib/orafce/orafce--3.18.sql
```diff
@@ -7638,284 +7333,3 @@ RETURNS numeric
...
-minvfunc = oracle.number_accum_inv,
-MFINALFUNC = oracle.number_stddev_samp
-);
-
-/* to_char */
-CREATE OR REPLACE FUNCTION oracle.to_char(oracle.number, text)
-RETURNS text
-AS $$SELECT pg_catalog.to_char($1::numeric, $2)$$
-LANGUAGE sql STABLE PARALLEL SAFE STRICT;
-/* oracle.number type support */
```
BingAo-Office: All of these can be implicitly converted to numeric implementations

contrib/orafce/orafce--3.18.sql
```diff
@@ -7254,255 +7188,14 @@ LANGUAGE internal
...
-CREATE OPERATOR oracle.^ (
-  LEFTARG   = oracle.number,
-  RIGHTARG  = oracle.number,
-  PROCEDURE = oracle.number_power,
-  COMMUTATOR = operator(oracle.^)
-);
-
---end
------------------------------------------------------------------
--- end
```
BingAo-Office: All of these can be implicitly converted to numeric implementations

contrib/orafce/orafce--3.18.sql
```diff
@@ -5736,27 +5736,6 @@ ALTER FUNCTION dbms_assert.enquote_name ( character varying ) STRICT;
...
-RETURNS numeric
-AS $$SELECT pg_catalog.round($1::numeric, $2)$$
-LANGUAGE sql IMMUTABLE STRICT;
-
-CREATE OR REPLACE FUNCTION oracle.trunc(float4, int)
-RETURNS numeric
-AS $$ SELECT pg_catalog.trunc($1::numeric, $2); $$
-LANGUAGE sql IMMUTABLE STRICT;
-
-
```
BingAo-Office, line 5759: Adjust the position

rasifr:
@BingAo-Office Thanks for the update but it needs to be rebased again.

rasifr (approved):
everything seems to be working

Issue #54, opened 2022-01-20
When Drop package body, the opened cursor defined in the package body should be closed at the same time.


KingZhaoTPG:
### Current Behavior

> hges_test=# create table test(x int, y varchar(100));
> CREATE TABLE
> hges_test=# insert into test values(1, 'One');
> INSERT 0 1
> hges_test=# insert into test values(2, 'Two');
> INSERT 0 1
> hges_test=# insert into test values(3, 'Three');
> INSERT 0 1
> hges_test=#
> hges_test=# create or replace package pkg is
> hges_test$# function c_open return int;
> hges_test$# function c_fetch return int;
> hges_test$# end;
> hges_test$# /
> CREATE PACKAGE
> hges_test=#
> hges_test=# create or replace package body pkg
> hges_test-# is
> hges_test$#     CURSOR c1 IS SELECT x,y FROM test;
> hges_test$#
> hges_test$# function c_open return int as
> hges_test$# begin
> hges_test$#
> 
> hges_test$# OPEN c1;
> hges_test$#
> 
> hges_test$# return 0;
> hges_test$# end;
> hges_test$#
> hges_test$# function c_fetch return int as
> hges_test$#
> 
> hges_test$# v_x  test.x%TYPE;
> hges_test$#
> 
> hges_test$# v_y  test.y%TYPE;
> hges_test$# begin
> hges_test$#
> 
> hges_test$# LOOP
> hges_test$#             FETCH c1 INTO v_x, v_y;
> hges_test$#
> 
> hges_test$#
> 
> hges_test$# EXIT WHEN NOT FOUND;
> hges_test$#
> 
> hges_test$#
> 
> hges_test$# raise info '%, %', v_x, v_y;
> hges_test$#
> 
> hges_test$# END LOOP;
> hges_test$#
> hges_test$#
> 
> hges_test$# CLOSE c1;
> hges_test$#
> 
> hges_test$# return 0;
> hges_test$# end;
> hges_test$#
> hges_test$# begin
> hges_test$# raise info 'initializer';
> hges_test$# end;
> hges_test$# /
> CREATE PACKAGE BODY
> hges_test=#
> hges_test=# select * from pg_cursors ;
>  name | statement | is_holdable | is_binary | is_scrollable | creation_time
> ------+-----------+-------------+-----------+---------------+---------------
> (0 rows)
> 
> hges_test=#
> hges_test=# select pkg.c_open();
> INFO:  initializer
>  c_open
> --------
[...]

PR #55, opened 2022-01-21
remove open cursor when drop package body

lk18347265415:
#54


rasifr (commented):
Hi @lk18347265415, I think the test case for it should be part of the patch.

rasifr (approved):

Issue #50, opened 2022-01-17
package function should be able to return the defined type.


rasifr:
## Bug Report
a package function should be able to return the defined type.
 
CREATE OR REPLACE PACKAGE example AS
   TYPE extyp IS RECORD (a INT, b VARCHAR(100));
   FUNCTION myfunc(x INT) RETURN extyp;
end;
/

### IvorySQL Version
1.0 and current master

### Expected behavior/code
Package should be created.

### Current Behaviour
```
ivorysql=# CREATE OR REPLACE PACKAGE example AS
ivorysql$#    TYPE extyp IS RECORD (a INT, b VARCHAR);
ivorysql$#    FUNCTION myfunc(x INT) RETURN extyp; -- error here
ivorysql$# end;
ivorysql$# /
ERROR:  type "extyp" does not exist
```

PR #59, opened 2022-02-14
Fix the issue of package function can not return the defined type

KingZhaoTPG:
#50

### DIFF
--- a/src/backend/commands/functioncmds.c
+++ b/src/backend/commands/functioncmds.c
@@ -117,6 +117,10 @@ compute_return_type(TypeName *returnType, Oid languageOid,
 		rettype = typeTypeId(typtup);
 		ReleaseSysCache(typtup);
 	}
+	else if (returnType->t_pkgoid)
+	{
+		rettype = GetTypeOidByTypeName(returnType);
+	}
 	else
 	{
 		char	   *typnam = TypeNameToString(returnType);
@@ -1227,6 +1231,11 @@ CreateFunction(ParseState *pstate, CreateFunctionStmt *stmt)
 	}
 	else if (stmt->returnType)
 	{
+		/* check if it's a package's functions */
+		if (stmt->proaccess == PACKAGE_MEMBER_PUBLIC ||
+			stmt->proaccess == PACKAGE_MEMBER_PRIVATE)
+			stmt->returnType->t_pkgoid = pstate->p_pkgoid;
+
 		/* explicit RETURNS clause */
 		compute_return_type(stmt->returnType, languageOid,
 							&prorettype, &returnsSet);
--- a/src/backend/commands/functioncmds.c
+++ b/src/backend/commands/functioncmds.c
@@ -117,10 +117,6 @@ compute_return_type(TypeName *returnType, Oid languageOid,
 		rettype = typeTypeId(typtup);
 		ReleaseSysCache(typtup);
 	}
-	else if (returnType->t_pkgoid)
-	{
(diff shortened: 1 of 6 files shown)


rasifr (changes requested):
Hi @KingZhaoTPG 

> This change triggers another bug. please see the example.

```
ivorysql=# CREATE OR REPLACE PACKAGE example
ivorysql-# AS
ivorysql$# TYPE extyp IS RECORD (a INT, b VARCHAR(100));
ivorysql$# FUNCTION myfunc(x INT) RETURN extyp;
ivorysql$# end;
ivorysql$# /
CREATE PACKAGE
ivorysql=# CREATE OR REPLACE PACKAGE example
AS
TYPE extyp IS RECORD (a INT, b VARCHAR(100));
FUNCTION myfunc(x INT) RETURN extyp;
end;
/
ERROR:  cannot drop type extyp because other objects depend on it
DETAIL:  function myfunc(integer) depends on type extyp
HINT:  Use DROP ... CASCADE to drop the dependent objects too.
```

> I also think that may be we should add the packaged oid added to search path implicitly when we are creating package elements. This way the objects can be locked up as usual without making another changes, like:

```
+	OverrideSearchPath *overridePath;
+
+	overridePath = GetOverrideSearchPath(CurrentMemoryContext);
+	overridePath->schemas = lcons_oid(pkgoid, overridePath->schemas);
+	PushOverrideSearchPath(overridePath);
..... i.e. CreateFunction etc.
....
+	PopOverrideSearchPath();
```

KingZhaoTPG:
Hi @rasifr ，Many thanks for your comments. I will update the PR asap.

japinli:
After some analyze, I find the `RemovePackageFunctions()`, `RemovePackageVariables()` and `RemoveVariabletype()` might update the catalog, however, it do not increase the command counter, which leads the `cannot drop type extyp because other objects depend on it`.  Here is a patch to fix it (based on KingZhaoTPG:IVORY_REL_1_STABLE), please review it.

```diff
index 0533d1402b..5c130aa4ce 100644
--- a/src/backend/commands/packagecmds.c
+++ b/src/backend/commands/packagecmds.c
@@ -127,12 +127,6 @@ CreatePackage(ParseState *pstate, CreatePackageStmt *stmt)
 		RemoveVariabletype(pkgoid, PACKAGE_MEMBER_PUBLIC);
 	}
 
-	/*
-	 * We must bump the command counter to make the newly-created package
-	 * tuple visible for package elements.
-	 */
-	CommandCounterIncrement();
-
 	CreatePackageElems(pstate, get_namespace_name(namespaceId),
 					   packagename, stmt->elems, stmt->replace, stmt->isbody);
 
@@ -520,6 +514,8 @@ RemovePackageFunctions(Oid packageOid, char proaccess, bool updatesrc)
 
 	table_endscan(scan);
 	table_close(rel, AccessShareLock);
+
+	CommandCounterIncrement();
 }
 
 
@@ -670,6 +666,8 @@ RemovePackageVariables(Oid packageOid, char varaccess)
 
 	table_endscan(scan);
 	table_close(rel, AccessShareLock);
+
+	CommandCounterIncrement();
 }
 
 /*
@@ -715,6 +713,8 @@ RemoveVariabletype(Oid packageOid, char typaccess)
 
 	table_endscan(scan);
 	table_close(rel, AccessShareLock);
+
+	CommandCounterIncrement();
 }
```

rasifr (approved):
These changes looks good to me.

Issue #58, opened 2022-02-03
Support for HINTS


luss:
## Enhancement

I think supporting the pg_hintplan extension similar to how you have incorporated orafce by default will be important

BingAo-Office:
Sorry, do you mean that the current IvorySQL does not support, or do you want to integrate into IovrySQL like orafce?

luss:
My suggestion is for IvorySQL to integrate in the pg_hint_plan project (in
a similar way to how you include the orafce functionality).

BingAo-Office:
Oh, thanks! The community will consider this!

PR #70, opened 2022-02-21
Implement mechanism to control creation of extensions and other objec…

codeforall:
…ts at initdb

The commit includes the contrib directory to default buid target IvorySQL src.
Along with that it  adds a new file ivy_module.config to contrib directory,
that can be used to list extensions and SQL scripts required to be
executed by initdb.

ivy_module.config gets installed at '$INSTALL_DIR/share/postgresql/' and is read
by initdb to install all extensions and scripts listed in that file.

### DIFF
--- a/GNUmakefile.in
+++ b/GNUmakefile.in
@@ -8,7 +8,7 @@ subdir =
 top_builddir = .
 include $(top_builddir)/src/Makefile.global
 
-$(call recurse,all install,src config)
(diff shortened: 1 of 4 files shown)


BingAo-Office:
Mybe we can deal the #58  in this PR too!

BingAo-Office:
You might also want to consider how Windows compiles, right

Issue #46, opened 2022-01-10
NCHAR and nls_length_semantics=byte


harukat:
## Bug Report

NCHAR is affected setting nls_length_semantics=byte as well as CHAR  and VARCHAR2.
I think it is undesirable.
It may be caused by the native PostgreSQL behavior that accepts type NCHAR as CHAR.

### IvorySQL Version
1.0 and current master

### OS Version / Configuration options 
V1.0 on RHEL 8.x  (install from rpm package)
master on  CentOS 6.x  (source build with configure --enable-debug)

### Current Behavior

```
ivdb1=# CREATE EXTENSION orafce ;
ivdb1=# SET search_path TO "$user", oracle, pg_catalog, public;
ivdb1=# SET nls_length_semantics TO byte;
ivdb1=# SELECT NCHAR(10) 'あいうえお';  -- 5 chars, UTF8 3byte char x 5 = 15 bytes
ERROR:  value too long for type character(10 byte)
```

BingAo-Office:
See PR #37 , is that resolved your problem? welcome to review our code!

harukat:
PR #37 is not a solution for this issue.
I tested v1.1 and I found that NCHAR have been affected  by nls_length_semantics=byte.
I think NCHAR must always behave as nls_length_semantics=char.

PR #72, opened 2022-03-01
support oracle nchar and nchar varying type and fix the bug #46

luotao24:
support oracle nchar, nchar varying type and fix the bug #46.


luotao24:
some errors have not solved, and some left issues.

PR #73, opened 2022-03-02
support oracle nchar and nchar varying type, and fix the bug #46

luotao24:


luotao24:
Removed the effect of nls_length_semantics on varchar types.

luotao24:
Removed the effect of nls_length_semantics on varchar types.

Issue #82, opened 2022-03-07
Create package using VARCHAR2 type prompt does not exist


sun-hchao:
The VARCHAR2 type cannot be used to create a package using the official documentation sample

For example:
oracle=# select version();
                                                        version
------------------------------------------------------------------------------------------------------------------------
 PostgreSQL 14.2 (IvorySQL 1.2) on x86_64-pc-linux-gnu, compiled by gcc (GCC) 4.8.5 20150623 (Red Hat 4.8.5-44), 64-bit
(1 行记录)

oracle=# show compatible_mode ;
 compatible_mode
-----------------
 oracle
(1 行记录)

oracle=# CREATE TABLE test(x INT, y VARCHAR2(100));
CREATE TABLE
oracle=# INSERT INTO test VALUES (1, 'One');
INSERT 0 1
oracle=# INSERT INTO test VALUES (2, 'Two');
INSERT 0 1
oracle=# INSERT INTO test VALUES (3, 'Three');
INSERT 0 1
oracle=# CREATE OR REPLACE PACKAGE example AUTHID DEFINER AS
oracle$#    -- Declare public type, cursor, and exception:
oracle$#    TYPE rectype IS RECORD (a INT, b VARCHAR2(100));
oracle$#    CURSOR curtype RETURN rectype%rowtype;
oracle$#
oracle$#    rec rectype;
oracle$#
oracle$#    -- Declare public subprograms:
oracle$#    FUNCTION somefunc (
oracle$#          last_name VARCHAR2,
oracle$#          first_name VARCHAR2,
oracle$#          email VARCHAR2
oracle$#    ) RETURN NUMBER;
oracle$#
oracle$#    -- Overload preceding public subprogram:
oracle$#    PROCEDURE xfunc (emp_id NUMBER);
oracle$#    PROCEDURE xfunc (emp_email VARCHAR2);
oracle$# END example;
oracle$# /
错误:  类型 "varchar2" 不存在

rasifr:
This issue has been fixed and verified. So closing it.

PR #91, opened 2022-03-10
issue82,support extended data type in the packege

lk18347265415:
#82

### DIFF
--- a/contrib/orafce/sql/orafce.sql
+++ b/contrib/orafce/sql/orafce.sql
@@ -984,3 +984,26 @@ SELECT oracle.unistr('wrong: \+2FFFFF');
 SELECT oracle.unistr('wrong: \udb99\u0061');
 SELECT oracle.unistr('wrong: \U0000db99\U00000061');
 SELECT oracle.unistr('wrong: \U002FFFFF');
+
+--
+-- test extended type in package
+--
+CREATE OR REPLACE PACKAGE ora_type AS
+FUNCTION tfunc(name VARCHAR2) RETURN NUMBER;
+END ora_type;
+/
+
+CREATE OR REPLACE PACKAGE BODY ora_type AS
+FUNCTION tfunc(name VARCHAR2) RETURN NUMBER
+AS
+Declare
+	sal NUMBER;
+BEGIN
+	sal := 1000.5;
+	raise info 'name is:%',name;
+	raise info 'salary is:%',sal;
+	return sal;
+END;
+END;
+/
+select ora_type.tfunc('SMITH');
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -3620,7 +3620,6 @@ GetOverrideSearchPath(MemoryContext context)
 	OverrideSearchPath *result;
 	List	   *schemas;
 	MemoryContext oldcxt;
-	Oid	compatibleNS;
 
 	recomputeNamespacePath();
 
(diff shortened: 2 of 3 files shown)


BingAo-Office (approved):

BingAo-Office (changes requested):
It is recommended to add the test case from the issue to this commit

KingZhaoTPG:
Hi @lk18347265415 , I think you should append a full test case as well.

contrib/orafce/sql/orafce.sql
```diff
@@ -984,3 +984,26 @@ SELECT oracle.unistr('wrong: \+2FFFFF');
...
 SELECT oracle.unistr('wrong: \U002FFFFF');
+
+--
+-- test extended type in package
+--
+CREATE OR REPLACE PACKAGE ora_type AS
+FUNCTION tfunc(name VARCHAR2) RETURN NUMBER;
+END ora_type;
+/
+
```
rasifr, line 995: I think the test case should be part of the src/test/regress/sql/packages.sql with search_path set to oracle.

src/backend/catalog/namespace.c
```diff
@@ -3629,15 +3629,10 @@ GetOverrideSearchPath(MemoryContext context)
...
-	while (schemas && linitial_oid(schemas) != activeCreationNamespace)
+	while (schemas && (linitial_oid(schemas) != activeCreationNamespace && linitial_oid(schemas) != compatibleNS))
 	{
 		if (linitial_oid(schemas) == myTempNamespace)
 			result->addTemp = true;
-		else if(OidIsValid(compatibleNS) && linitial_oid(schemas) == compatibleNS)
-		{
-			schemas = list_delete_first(schemas);
-			continue;
-		}
```
rasifr, line 3640: I think this is breaking the existing functionality of this function. GetOverrideSearchPath() function removes the implicitly added namespaces. But with this check, we are preventing it from doing that. 

GetOverrideSearchPath() should not have these checks here, this logic should go to the PushOverrideSearchPath() instead. like
```
if (compatible_db == COMPATIBLE_ORA)
 {
 		compatibleNS = get_namespace_oid("oracle", true);
 		if (OidIsValid(compatibleNS) &&
 			!list_member_oid(oidlist, compatibleNS))
 			oidlist = lcons_oid(compatibleNS, oidlist);
}
```

rasifr (changes requested):

contrib/orafce/sql/orafce.sql
lk18347265415: I think this test case just needs to be added in orafce because this is mainly to test the data types created in orafce.And if  orafce extension is no installed,add this test case in package.sql will throws error.

src/backend/catalog/namespace.c
lk18347265415: I think GetOverrideSearchPath

> I think this is breaking the existing functionality of this function. GetOverrideSearchPath() function removes the implicitly added namespaces. But with this check, we are preventing it from doing that.
> 
> GetOverrideSearchPath() should not have these checks here, this logic should go to the PushOverrideSearchPath() instead. like
> 
> ```
> if (compatible_db == COMPATIBLE_ORA)
>  {
>  		compatibleNS = get_namespace_oid("oracle", true);
>  		if (OidIsValid(compatibleNS) &&
>  			!list_member_oid(oidlist, compatibleNS))
>  			oidlist = lcons_oid(compatibleNS, oidlist);
> }
> ```

the GetOverrideSearchPath() function only prevent to delete the oralce schema,and it will not affect others.

rasifr: is it really needed to add 'addora' variable? if "compatible_db == COMPATIBLE_ORA" is true that means we have already added oracle schema implicitly before calling GetOverrideSearchPath().

So maybe we should just add it again in PushOverrideSearchPath() when "compatible_db == COMPATIBLE_ORA" is true.

PR #97, opened 2022-03-16
fix qsyntax regress test error

Huajianjin:


rasifr:
Hi @Huajianjin I'm afraid the test still fails on CentOS 7. I test this patch on my fork and ran the regression. Please see the output at [files.](https://github.com/rasifr/IvorySQL/actions/runs/1998492014)

rasifr (changes requested):
This fix fails on the CentOS platform. So I have disabled this test case in the regression for now. We should enable it along with this patch, when this issue is fixed.

Issue #66, opened 2022-02-17
Stand alone PL/iSQL procedures does not allow calling another procedure from within without call keyword


codeforall:
## Bug Report

As Per PL/iSQL the procedures should allow calling other procedures from within without using the call keyword.
But standalone PL/iSQL procedures are not allowing that.

```
postgres=# CREATE OR REPLACE PROCEDURE pro1 IS
postgres$# BEGIN
postgres$# DBMS_OUTPUT.PUT_LINE('some text');
postgres$# END;
postgres$# /
ERROR:  syntax error at or near "DBMS_OUTPUT"
LINE 3: DBMS_OUTPUT.PUT_LINE('some text');
        ^
QUERY:  
BEGIN
DBMS_OUTPUT.PUT_LINE('some text');
END;
```

However, calling the procedure from within the package procedure works without CALL keyword

```
postgres=# CREATE OR REPLACE PACKAGE pkg1 AS
PROCEDURE pro1;
end;
/
CREATE PACKAGE

postgres=# CREATE OR REPLACE PACKAGE BODY pgk1 AS
PROCEDURE pro1 AS 
BEGIN
DBMS_OUTPUT.PUT_LINE('some text');
END;
END;
 /
CREATE PACKAGE BODY

```

luss:
+10    :-)

PR #98, opened 2022-03-17
Support stand alone PL/iSQL procedures call another procedure without…

KingZhaoTPG:
#66

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -17,10 +17,10 @@ jobs:
       run: |
         docker build -t ivorysql-${{ matrix.container-os }}:latest -<<EOF
         FROM    centos:7
-        RUN     yum update -y
         RUN     rpm --import http://download.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-7
         RUN     yum install -y epel-release centos-release-scl
         RUN     yum groups mark install "Development Tools"
+        RUN     yum groupinstall -y "Development Tools"
         RUN     yum install -y \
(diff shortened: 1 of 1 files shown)


src/pl/plisql/src/pl_gram.y
```diff
@@ -2236,11 +2316,6 @@ stmt_execsql	: K_IMPORT
 												make_execsql_stmt(T_WORD, @1);
 							new->expr = execstmt->sqlstmt;
```
rasifr, line 2317: In stmt_execsql rule, can't we just call read_sql_stmt instead of make_execsql_stmt? since we have already determined that its a procedure call...

src/test/regress/sql/select_hierar.sql
```diff
@@ -182,7 +182,7 @@ SELECT
   parent_entity, child_entity
 FROM entities
 CONNECT BY PRIOR child_entity = parent_entity AND PRIOR val = 50
-ORDER BY 1;
+ORDER BY val;
```
rasifr, line 186: I don't think it's needed anymore. The output order issue has been fixed.

src/pl/plisql/src/pl_gram.y
KingZhaoTPG: Hi @rasifr , thank you. I've tried this way, but read_sql_stmt doesn't seem to fit here.

KingZhaoTPG: Hi @rasifr , thank you. I've tried this way, but read_sql_stmt doesn't seem to fit here.

Issue #88, opened 2022-03-08
Calling package functions requires schema qualification for packages created in pg_catalog


codeforall:
If the package is created in pg_catalog namespace (implicitly part of search_path), calling the package object without schema qualification throws function/procedure not found error.

```SQL
CREATE OR REPLACE PACKAGE pg_catalog.pkg IS
FUNCTION f RETURN INT;
END;
/
CREATE PACKAGE
postgres=# CREATE OR REPLACE PACKAGE BODY pg_catalog.pkg IS
FUNCTION f RETURN INT IS
BEGIN
return 10;
END;
END;
postgres$# /
CREATE PACKAGE BODY
```
Try selecting pkg.f()
```SQL
postgres=# select pkg.f();
ERROR:  schema "pkg" does not exist
LINE 1: select pkg.f();
```
It workd with schema qualification
```SQL
postgres=# select pg_catalog.pkg.f();
 f  
----
 10
(1 row)
```

PR #99, opened 2022-03-17
Support pkg created in pg_catalog to call pkg functions without pg_ca…

KingZhaoTPG:
#88

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -4002,6 +4002,58 @@ FindDefaultConversionProc(int32 for_encoding, int32 to_encoding)
 }
 
 /*
+ * get_package_oid - given a package name, look up the OID
+ *
+ * If missing_ok is false, throw an error if name not found.  If true, just
+ * return InvalidOid.
+ */
+Oid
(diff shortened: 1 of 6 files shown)


rasifr (commented):
- With GetOverrideSearchPath/PushOverrideSearchPath we don't need to add the pg_catalog to schema list explicitly, because it get's added in PushOverrideSearchPath.
- perhaps we should move get_package_oid() to namespace.c file, so we don't have to fetch search path again after GetOverrideSearchPath/PushOverrideSearchPath. And we can loop over activeSearchPath directly.

KingZhaoTPG:
> With GetOverrideSearchPath/PushOverrideSearchPath we don't need to add the pg_catalog to schema list explicitly, because it get's added in PushOverrideSearchPath.
perhaps we should move get_package_oid() to namespace.c file, so we don't have to fetch search path again after GetOverrideSearchPath/PushOverrideSearchPath. And we can loop over activeSearchPath directly.

Hi @rasifr, thank you and your comments. I will update my pr with this approach.

rasifr (approved):
looks good to me

PR #101, opened 2022-03-18
Package tab complete

codeforall:


rasifr (commented):
1- \dPKG and \dpkg display completely different results. Although \dpkg is an invalid meta command, but since we are adding this support for packages, perhaps we should support both \dPKG and \dpkg.

`ivorysql=# \dpkg
                            Access privileges
 Schema | Name | Type | Access privileges | Column privileges | Policies 
--------+------+------+-------------------+-------------------+----------
(0 rows)`

2- S modifier is usually used to display more information. maybe we should take this opportunity to have this modifier differentiate between user-created and system installed packages?

Issue #107, opened 2022-03-24
Drop package body not working


codeforall:
DROP PACKAGE BODY does not drop the package body yet the command reports as a success.
```SQL
postgres=# drop package body example ;
DROP PACKAGE BODY
postgres=# drop package body example ;
DROP PACKAGE BODY
postgres=# 
CREATE PACKAGE BODY example AS                                                                                                                                                        nelems NUMBER; -- private variable, visible only in this package
...
END;
/
ERROR:  package "example" body already exists
```

PR #123, opened 2022-04-19
Fix the drop package body not working issue

KingZhaoTPG:
#107


rasifr (commented):
Shouldn't Anum_pg_package_pkgbody be null when package body was dropped in the first place?

I think the real issue is that update_package_src() is updating the Anum_pg_package_pkgbody with an empty string. whereas It should mark this column as null.

PR #139, opened 2022-06-06
Fixed package privileges. We were missing the check to test if a package

rasifr:
member is a private function or procedure.

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -170,6 +170,7 @@ typedef struct
 	List	   *searchPath;		/* the desired search path */
 	Oid			creationNamespace;	/* the desired creation namespace */
 	int			nestLevel;		/* subtransaction nesting level */
+	Oid			pkgoid;			/* Oid of the package pushed to override search path */
 } OverrideStackEntry;
 
 static List *overrideStack = NIL;
@@ -3621,10 +3622,14 @@ GetOverrideSearchPath(MemoryContext context)
 	OverrideSearchPath *result;
 	List	   *schemas;
 	MemoryContext oldcxt;
+	OverrideStackEntry *entry = NULL;
 
 	recomputeNamespacePath();
 
 	oldcxt = MemoryContextSwitchTo(context);
+
+	if (overrideStack != NIL)
+		entry = (OverrideStackEntry *) linitial(overrideStack);
 
 	result = (OverrideSearchPath *) palloc0(sizeof(OverrideSearchPath));
 	schemas = list_copy(activeSearchPath);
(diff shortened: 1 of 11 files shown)


KingZhaoTPG (commented):
Hi @rasifr , I think there are still some issues, please check again. Let's look at my TC as follows:

`CREATE OR REPLACE PACKAGE exp 
AS
  FUNCTION func(x VARCHAR(100)) RETURN INT;
  
END;
/

CREATE OR REPLACE PACKAGE BODY exp AS
  priv_var  INT := 1;

  FUNCTION func(x VARCHAR(100)) RETURN INT AS
  BEGIN
    RETURN 0;
  END;
    
  FUNCTION func(y int) RETURN int
  AS
  BEGIN
    RETURN 1000;
  END;
  
  PROCEDURE priv_proc(x VARCHAR(100)) AS
  BEGIN
    NULL;
  END;
  FUNCTION priv_func(x VARCHAR(100)) RETURN INT AS
  BEGIN
    RETURN 0;
  END;
END;
/



create or replace package pkg
as
    function tfunc(a int) return int;
end;
/

create or replace package body pkg
as
    p_n int;
    
    function tfunc(a int) return int
    as
    begin
        return p_n;
    end;
begin
    p_n := exp.func(1);
end;
/`



hges_test=# select pkg.tfunc(2);
WARNING:  cache reference leak: cache pg_package (85), tuple 0/3 has count 1
WARNING:  cache reference leak: cache pg_proc (44), list 0x7f4336c79b68 has count 1
WARNING:  leaked override search path
 tfunc
-------

(1 row)

As you can see, here I get such a WARNING prompt, I call the private function of exp in pkg package, and it does not prompt me that package private function"xxxx" is not accessible.

KingZhaoTPG (commented):
Hi @rasifr ,  This update nicely resolves the 'WARNING: cache reference" issue. but there is still a problem with the private function call of the package. you can see my TC as follows:

```
create or replace package pkg
is
	function f return int;
end;
/

create or replace package body pkg
is
    pv_i int := 1;
    
    function func(a int) return int
    is
    begin
        return pv_i + a;
    end;
    
	function f return int
	is
        m int;
	begin
		m := func(100);
        return m;
	end;
end;
/

hges_test=#
hges_test=# select pkg.f();
ERROR:  package private function ("func") is not accessible
LINE 1: m := func(100)
             ^
QUERY:  m := func(100)
CONTEXT:  PL/iSQL function f() line 4 at assignment
hges_test=#
```

Functions inside a package on Oracle can access the package's private functions. for my TC, I run the 'select pkg.f()' and I should get a result of 101. please check it again.

Thanks!
/shengbin

PR #169, opened 2022-11-02
Fix when creating SEQUENCE, it is mistaken for creating a package

xiepeng38616:
Repair When creating a SEQUENCE, if the SEQUENCE name contains' package ', it is mistaken for creating a package statement under psql。
if the SEQUENCE name contains' function ' or 'procedure' is the same.
e.g.
CREATE SEQUENCE public.ecologypackageinfo_id_seq
AS integer
START WITH 1
INCREMENT BY 1
NO MINVALUE
NO MAXVALUE
CACHE 1;

### DIFF
--- a/src/fe_utils/psqlscan.l
+++ b/src/fe_utils/psqlscan.l
@@ -628,9 +628,45 @@ other			.
 
 {create_fun}	{
 					char *str = pg_strdup(output_buf->data);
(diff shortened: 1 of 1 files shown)


tanyang-star:
thanks the commit .  but IvorySQL 3.0 will use new framework. so Temporarily unable to merge into the main branch

PR #173, opened 2022-12-08
bugfix: When the PRIOR clause is followed by an illegal column, the server terminated abnormally

HuangSongZ:
=# CREATE TABLE t_tab (id int, manager_id int);
=# SELECT id, manager_id from t_tab CONNECT BY prior 50 =  manager_id ;
        This probably means the server terminated abnormally
        before or while processing the request.


rasifr (commented):
The same changes needed for the ConnectByRoot as well. It looks like there is another PR 174 that takes care of it. So This PR is not needed anymore.

PR #174, opened 2022-12-08
bugfix: When CONNECT_BY_ROOT parameter exception, the server terminated abnormally

HuangSongZ:
```sql
create table example(manager_id int, employee_id int, employee VARCHAR);

SELECT employee_id, CONNECT_BY_ROOT 'employee_id' as "Manager"
FROM example
START WITH manager_id is not null
CONNECT BY PRIOR employee_id = manager_id;

        This probably means the server terminated abnormally
        before or while processing the request.
```


rasifr (approved):
looks good to me.

PR #175, opened 2022-12-08
Hierarchical query supports column expression in targetList

HuangSongZ:
Currently，Hierarchical query do not support column expression in targetList.

```sql
SELECT employee_id + 1
FROM example
START WITH manager_id is not null
CONNECT BY PRIOR employee_id = manager_id;
ERROR:  column reference "employee_id" is ambiguous
LINE 1: SELECT employee_id + 1
               ^

```


rasifr (changes requested):
most of the code in the PR is same one from the PR#174. That patch is solving some other problem and should not be part of current PR. please remove the previous patch and submit code for the current one only.

grantzhou:
Please remove the accepted code (PR 174) from this patch and re-create the PR.
Thanks!

PR #213, opened 2023-05-04
IvorySQL:Add initdb -m,support postgres mode or oracle mode

tanyang-star:
Add initdb -m,support postgres mode or oracle mode


tanyang-star:
#212

PR #226, opened 2023-05-09
IvorySQL:case conversion

wangjie-star:
1. Supports case conversion for quoted identifiers.
2. add guc parameter "enable_case_switch" to control whether to enable or disable "case conversion" feature.


tanyang-star:
#221

PR #227, opened 2023-05-10
IvorySQL:Add dual cmd:psql parser to support different database prototol

tanyang-star:


tanyang-star:
Add psql parser ,when use cmd:psql can support oracle gramer  #225

PR #231, opened 2023-05-15
IvorySQL:Modified the format(=) of including ivorysql.conf in the pos…

tanyang-star:
…tgres.conf file


tanyang-star:
fix:Modified the format(=) of including ivorysql.conf in the postgres.conf file #230

tanyang-star:
#230

PR #240, opened 2023-05-30
IvorySQL: Compatible with Oracle's built-in data types and built-in functions.

wangjie-star:
Compatible with Oracle's built-in data types and built-in functions.


wangjie-star:
#239

PR #244, opened 2023-05-31
IvorySQL:Long and raw are a domain with typmod that triggers assert().

wangjie-star:
long and raw are a domain with typmod that triggers assert().


tanyang-star:
#243

PR #245, opened 2023-06-01
IvorySQL: fix bug - unable to switch pure uppercase mode by using SQL statements to modify 'search_path'

DutMsn:


tanyang-star:
#247  fix:seaarch_path

PR #249, opened 2023-06-01
IvorySQL:fix oracle-check-world

tanyang-star:
support cmd:"make oracle_check_world", but not support "pg_upgrade" ,"027_stream_regress.pl"


tanyang-star:
#246

Issue #342, opened 2023-08-01
When the sequence value in the inserted table is greater than the maximum value of the field type, the insert cannot be inserted


zhanghongyuan111:
## Bug Report
When the sequence value in the inserted table is greater than the maximum value of the field type, the insert cannot be inserted
### IvorySQL Version
IvorySQL 3.0
### OS Version (uname -a)
CentOS 7

### Current Behavior
![image](https://github.com/IvorySQL/IvorySQL/assets/132979275/363d9433-8ea4-45fe-8401-2658b5d73108)

### Expected behavior/code
![image](https://github.com/IvorySQL/IvorySQL/assets/132979275/90ff3287-03a0-4875-be0a-c41914e6b0ad)

### Step to reproduce
create table other_max2(userid number(8) generated always as identity  start with 99999999 increment by 2,uname varchar2(20));
insert into other_max2 (uname) values ('xiaodong');
insert into other_max2 (uname) values ('xiaohong');

PR #394, opened 2023-08-04
IvorySQL:When the sequence value in the inserted table is greater than the maximum value of the field type, the insert cannot be inserted

zhanghongyuan111:


tanyang-star:
#342

Issue #400, opened 2023-08-07
After changing the NLS time format, SYSTIMESTAMP does not display the time according to the prescribed format


zhanghongyuan111:
## Bug Report
After changing the NLS time format, SYSTIMESTAMP does not display the time according to the prescribed format
### IvorySQL Version
IvorySQL master
### OS Version (uname -a)
CentOS 7

### Current Behavior
![image](https://github.com/IvorySQL/IvorySQL/assets/132979275/4ef20d7e-ee70-4212-ab04-ff7e297d6468)

### Expected behavior/code
返回符合NLS格式的时间
### Step to reproduce
alter session set NLS_TIMESTAMP_FORMAT='DD-MON-RR HH24:MI:SS tzh:tzm';
select systimestamp() from dual;

PR #424, opened 2023-08-10
IvorySQL:After changing the NLS time format, SYSTIMESTAMP does not display thetime according to the prescribed format

zhanghongyuan111:


wangjie-star:
#400

Issue #627, opened 2024-01-30
匿名块不能正确处理outer label


Leo-XM-Zeng:
## Enhancement
```
<<outer>>  -- label
DECLARE                           
  birthdate DATE := TO_DATE('09-AUG-70', 'DD-MON-YY');
BEGIN                                 
  DECLARE
    birthdate DATE := TO_DATE('29-SEP-70', 'DD-MON-YY');
  BEGIN
    IF birthdate = outer.birthdate THEN 
      RAISE NOTICE '%', 'Same Birthday';
    ELSE                    
      RAISE NOTICE '%', 'Different Birthday';
    END IF;
  END;
END;
/
```
![image](https://github.com/IvorySQL/IvorySQL/assets/69310610/78ca2f9d-520b-497f-99a0-7ceca47c2a8c)

PR #628, opened 2024-01-30
支持匿名块正确处理outer label #627

Leo-XM-Zeng:
#627
image

## Summary by CodeRabbit

- **New Features**
	- Introduced support for the 'outer' keyword in PL/SQL syntax.


bigplaice:
Hi, Leo, thanks for support for IvorySQL! 
We spent some time to understand your change and our conclusion is this PR may not be necessary. 

The error is actually complaining that one reserved keyword is used as lable name and this is not allowed. 
If you use other reserved keyword to replace "outer", same error would happen.  You can try folliwng keywords:

AUTHORIZATION
BINARY
COLLATION
CONCURRENTLY
... 

more keywords can be found by searching "reserved (can be function or type)" at http://www.postgres.cn/docs/14/sql-keywords-appendix.html


These keywords have been reserved. Turning them into unreserved keyword may bring some unexpected results. 

While Oracle SQL keywords are not reserved and looks like you can use them, but Oracle don't encourage you to use keyword as object name. 
Please refer to this page:
https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Oracle-SQL-Keywords.html#GUID-82EA000B-5661-41EB-AAF7-6BDDB4AB58EE
Oracle SQL keywords are not reserved. However, Oracle uses them internally in specific ways. Therefore, if you use these words as names for objects and object parts, then your SQL statements may be more difficult to read and may lead to unpredictable results.

Leo-XM-Zeng:
> Hi, Leo, thanks for support for IvorySQL! We spent some time to understand your change and our conclusion is this PR may not be necessary.
> 
> The error is actually complaining that one reserved keyword is used as lable name and this is not allowed. If you use other reserved keyword to replace "outer", same error would happen. You can try folliwng keywords:
> 
> AUTHORIZATION BINARY COLLATION CONCURRENTLY ...
> 
> more keywords can be found by searching "reserved (can be function or type)" at https://www.postgresql.org/docs/16/sql-keywords-appendix.html
> 
> These keywords have been reserved. Turning them into unreserved keyword may bring some unexpected results.
> 
> While Oracle SQL keywords are not reserved and looks like you can use them, but Oracle don't encourage you to use keyword as object name. Please refer to this page: https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Oracle-SQL-Keywords.html#GUID-82EA000B-5661-41EB-AAF7-6BDDB4AB58EE Oracle SQL keywords are not reserved. However, Oracle uses them internally in specific ways. Therefore, if you use these words as names for objects and object parts, then your SQL statements may be more difficult to read and may lead to unpredictable results.

我表示理解 虽然这个是oracle提供的测试案例
image

Issue #624, opened 2024-01-23
IvorySQL 3.0 创建 FUNCTION  报specified for unlabeled block错误


shlei6067:
## Bug Report
创建function 报 specified for unlabeled block

### IvorySQL Version
IvorySQL 3.0

### OS Version (uname -a)
Centos 7.9

=====具体代码如下
[ivorysql@ivorysqldb ~]$ psql -p 1521 -d postgres
psql (16.0)
Type "help" for help.

postgres=# CREATE TABLE LOGON_LOG_EN
postgres-# (
postgres(#   REC_ID           NUMBER(20)                   NOT NULL,
postgres(#   COM_ID           NUMBER(10)                   NOT NULL,
postgres(#   OPERATOR_NO      VARCHAR2(10 CHAR)            NOT NULL,
postgres(#   LOG_USER_NAME    VARCHAR2(255 CHAR)           NOT NULL,
postgres(#   LAST_LOGON_TIME  DATE                         NOT NULL,
postgres(#   LOGON_PAGE       VARCHAR2(1500 CHAR),
postgres(#   LOGON_IP         VARCHAR2(50 CHAR),
postgres(#   REP_TIME         TIMESTAMP(6) WITH TIME ZONE  DEFAULT TO_DATE('1900-01-01 00:00:00','YYYY-MM-DD HH24:MI:SS') NOT NULL,
postgres(#   SESSION_ID       VARCHAR2(100 CHAR),
postgres(#   COOKIE_ID        VARCHAR2(100 CHAR),
postgres(#   LOGON_TYPE       VARCHAR2(2 CHAR),
postgres(#   PRE_LOGON_PAGE   VARCHAR2(1500 CHAR),
postgres(#   LOGON_IN_CODE    VARCHAR2(3 CHAR),
postgres(#   COOKIE_PID       VARCHAR2(100 CHAR)
postgres(# );
CREATE TABLE
postgres=# 
postgres=# CREATE OR REPLACE FUNCTION "F_GET_LAST_LOGON_IP" (
postgres(#    P_COMID         NUMBER,
postgres(#    P_OPEARTORNO    VARCHAR2
postgres(# )
postgres-#    RETURN VARCHAR2
postgres-# IS
postgres-#    output_value   VARCHAR2 (1000);
postgres-# BEGIN
postgres-#    output_value := '';
postgres-#    BEGIN
postgres-#       SELECT   LOGON_IP
postgres-#         INTO   output_value
[...]

PR #629, opened 2024-02-02
Match END label with the function or procedure name

bigplaice:
Match END label with the function or procedure name.
Issue url: https://github.com/IvorySQL/IvorySQL/issues/624


gaoxueyu:
approved

PR #639, opened 2024-03-07
Support xml functions with bug fixes

bigplaice:
Support xml functions with bug fixes


bigplaice:
close as need rework.

PR #647, opened 2024-03-26
Clear warnings in build for IvorySQL 3.2

bigplaice:


contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
```diff
@@ -1953,14 +1953,28 @@ tm_trunc(struct pg_tm *tm, text *fmt)
 		break;
 	CASE_fmt_YYYY
 		tm->tm_mon = 1;
+		tm->tm_mday = 1;
```
bigplaice, line 1956: the code except "break" is necessary?

bigplaice: I think I know the reason.
Before the "break;" lines are inserted, all CASE_fmt_xxx would fall through until the to the last one. As a result: 
    tm->tm_mday = 1;
    tm->tm_hour = 0;
    tm->tm_min = 0;
are executed. 

So, if we want to add "break ;" for each case, we have to ensure the equivalence. This is why we add other code besides "break;"

gaoxueyu (approved):

Issue #609, opened 2023-12-28
The TO_MULTI_BYTE(), TO_SINGLE_BYTE(), ASCIISTR(), COMPOSE(), DECOMPOSE() functions are not implemented


zhanghongyuan111:

gaoxueyu:
related to PR #671 and PR #677

PR #671, opened 2024-06-29
Add support for TO_MULTI_BYTE(), TO_SINGLE_BYTE(), ASCIISTR(), COMPOSE(), DECOMPOSE()

imranzaheer612:
Add support for oracle ASCIISTR() function.
Add test cases for verification.

https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/ASCIISTR.html

This PR is related to issue https://github.com/IvorySQL/IvorySQL/issues/609

## Summary by CodeRabbit

- **New Features**
	- Introduced SQL functions for character datatype conversions:
		- `sys.asciistr(text)`
		- `sys.decompose(text, text DEFAULT 'canonical')`
		- `sys.compose(text)`
		- `sys.to_multi_byte(text)`
		- `sys.to_single_byte(text)`
- **Tests**
	- Added tests for `asciistr`, `compose`, and `decompose` functions with various input strings, including ASCII, non-ASCII, and mixed characters.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
+++ b/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
@@ -4,6 +4,15 @@
  *
  ***************************************************************/
 /* length/lengthb for CHAR(n char/byte) */
+
+CREATE FUNCTION sys.asciistr(text)
+RETURNS text
+AS 'MODULE_PATHNAME','ora_asciistr'
+LANGUAGE C
+STRICT
+PARALLEL SAFE
+IMMUTABLE;
+
 CREATE FUNCTION sys.length(text)
 RETURNS integer
 AS 'MODULE_PATHNAME','oracharlen'
--- a/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
+++ b/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
@@ -12,6 +12,34 @@ STRICT
 PARALLEL SAFE
 IMMUTABLE;
 
+
+CREATE FUNCTION sys.decompose(text, text DEFAULT 'canonical')
+RETURNS text
+AS $$
+BEGIN
+    IF LOWER($2) = 'canonical' THEN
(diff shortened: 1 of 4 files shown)


contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
```diff
@@ -4,6 +4,15 @@
  *
  ***************************************************************/
 /* length/lengthb for CHAR(n char/byte) */
```
balinorLiang, line 6: It's better to add this comment in the header of length function

contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
```diff
@@ -1690,4 +1691,84 @@ text_position_prev(int start_pos, TextPositionState *state)
...
+            str++;
+        } else if ((c & 0xE0) == 0xC0) {
+            /* Two-byte UTF-8 sequence */
+            codePoint = ((c & 0x1F) << 6) | (str[1] & 0x3F);
+            appendUTF16Escape(&output, (uint16_t) codePoint);
+            str += 2;
+        } else if ((c & 0xF0) == 0xE0) {
+            /* Three-byte UTF-8 sequence */
+            codePoint = ((c & 0x0F) << 12) | ((str[1] & 0x3F) << 6) | (str[2] & 0x3F);
+            appendUTF16Escape(&output, (uint16_t) codePoint);
```
bigplaice, line 1752: since a three-byte unicode sequence is being added, uint32_t should be used here. 
As a result, the second parameter of appendUTF16Escape() should be uint32_t too.

imranzaheer612: * The three-byte UTF-8 sequence (0xE0) covers code points from U+0800 to U+FFFF which falls under the BMP and can be represented by a single uint16_t.

https://exploringjs.com/js/book/ch_unicode.html#utf-16
https://learn.microsoft.com/en-us/globalization/encoding/unicode-standard#surrogate-pairs

Issue #731, opened 2025-01-20
incomplete sql when using invisible columns and pg_dump


jiaoshuntian:
When exporting a table with invisible columns using pgdump, the output SQL statement is incomplete
![Image](https://github.com/user-attachments/assets/c845262c-1566-411c-8af4-61b9829c2931)

gaoxueyu:
Issue has been fixed via PR 743.
And regression test cases need to be added later.

PR #743, opened 2025-02-14
Fix pg_dump for invisible columns

imranzaheer612:
* There was a bug while dumping the invisible columns of a table.

related issue: https://github.com/IvorySQL/IvorySQL/issues/731

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected an issue with how column visibility is handled during schema export to ensure accurate results.
  - Removed redundant checks for column visibility, leading to a more efficient and reliable schema dump process.


gaoxueyu:
Thanks Imran for your contribution. The code for this PR is fine. 
I think we can merge it first, and the regression code can be submitted in a separate PR.

PR #837, opened 2025-07-30
development for OUT parameter

bigplaice:
Modify the create and call behavior of the function, if the function has the out parameter.

If the function has the out parameter, the returned datatype of the function can be any datatype, and the function returned datatype has no relationship with out pararmeter datatype.

If the function returned datatype is not VOID, must provide the RETURN statement in the function body.

When calling a function in plisql, must specify a variable to carry the out parameter value.

If turn off the guc ivorysql.allow_out_parameter_const, calling a function in SQL will report an error if the constant or NULL is passed for the out parameter.
If turn on the guc ivorysql.allow_out_parameter_const, calling a function a SQL, passing constant or NULL for the out parameter is allowed.

Add a new grammer for anonymous block:
DO [LANGUAGE lang_name] code [USING IN|OUT|INOUT|IN OUT, ...]

Libpq now supports binding parameters by position or name when calling a function. Some interface functions similar to OCI are implemented.


src/backend/access/common/printtup.c
```diff
@@ -513,3 +513,35 @@ debugtup(TupleTableSlot *slot, DestReceiver *self)
 
 	return true;
 }
+
+/*
+ * SetDestSendDescription:
+ * During PBE executing, we should set send_description to be true for 
+ * a anonymous block which has out parameter.
```
c2main, line 520: `for an anonymous block`

bigplaice: thanks for catching this typo, will fix it in another patch.

PR #850, opened 2025-08-12
fix oracle regression test case status unstable: remove it

OreoYang:
such as PR#844

### DIFF
--- a/src/oracle_test/regress/expected/stats.out
+++ b/src/oracle_test/regress/expected/stats.out
@@ -1481,23 +1481,21 @@ DECLARE
 BEGIN
   -- we don't want to wait forever; loop will exit after 30 seconds
   FOR i IN 1 .. 300 LOOP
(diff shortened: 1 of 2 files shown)


src/oracle_test/regress/expected/stats.out
```diff
@@ -1507,19 +1505,14 @@ UPDATE brin_hot SET val = -3 WHERE id = 42;
...
-SELECT pg_stat_get_tuples_hot_updated('brin_hot'::regclass::oid);
- pg_stat_get_tuples_hot_updated 
---------------------------------
-                              1
+ wait_for_hot_stats 
+--------------------
+ 
 (1 row)
 
+-- SELECT pg_stat_get_tuples_hot_updated('brin_hot'::regclass::oid);
```
bigplaice, line 1513: Since the function wait_for_hot_stats() can work well now, you don't have to delete this line.

OreoYang: submit new PR https://github.com/IvorySQL/IvorySQL/pull/851

PR #911, opened 2025-10-09
fix Issue#887 ivorysql-wasm meson build error

OreoYang:
1. encountered: uuid library ossp missing required function uuid_export
2. FAILED: src/bin/initdb/po/zh_CN/LC_MESSAGES/initdb-18.mo

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -160,7 +160,7 @@ option('systemd', type: 'feature', value: 'auto',
   description: 'systemd support')
 
 option('uuid', type: 'combo', choices: ['none', 'bsd', 'e2fs', 'ossp'],
-  value: 'ossp',
+  value: 'e2fs',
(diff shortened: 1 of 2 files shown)


meson_options.txt
```diff
@@ -160,7 +160,7 @@ option('systemd', type: 'feature', value: 'auto',
   description: 'systemd support')
 
 option('uuid', type: 'combo', choices: ['none', 'bsd', 'e2fs', 'ossp'],
-  value: 'ossp',
+  value: 'e2fs',
   description: 'Use LIB for contrib/uuid-ossp support')
```
bigplaice, line 164: should also modify the description.

OreoYang: ‘contrib/uuid-ossp’ Here 'uuid-ossp' is the extension name, let's keep it.

Issue #985, opened 2025-11-27
PL/iSQL does not support COMMIT/ROLLBACK in procedures (Oracle compatibility gap)


rophy:
# Issue: COMMIT/ROLLBACK Not Supported in PL/iSQL Procedures

## Problem

PL/iSQL procedures do not support COMMIT/ROLLBACK, while PL/pgSQL procedures do. This is a critical Oracle compatibility gap.

### 1. Test in IvorySQL Port 5432 (PostgreSQL Mode) - ✅ WORKS

**Connect:**
```bash
psql -h localhost -p 5432 -d postgres
```

**Run this code:**
```sql
CREATE TABLE test_tx (id INT, msg TEXT);

CREATE PROCEDURE test_rollback(p_id INT) AS $$
BEGIN
    INSERT INTO test_tx VALUES (p_id, 'should rollback');
    ROLLBACK;
    INSERT INTO test_tx VALUES (p_id + 1, 'after rollback');
    COMMIT;
END;
$$;

CALL test_rollback(10);

SELECT * FROM test_tx ORDER BY id;
```

**Expected Result: ✅ WORKS**
```
 id |      msg
----+----------------
 11 | after rollback
```
- First insert (id=10) was rolled back
- Second insert (id=11) was committed
- No errors

### 2. Test in IvorySQL Port 1521 (Oracle Mode) - ❌ FAILS

**Connect:**
```bash
psql -h localhost -p 1521 -d postgres
```

**Run this code:**
```sql
CREATE TABLE test_tx (id INT, msg TEXT);

CREATE PROCEDURE test_rollback(p_id INT) AS $$
BEGIN
    INSERT INTO test_tx VALUES (p_id, 'should rollback');
    ROLLBACK;
    INSERT INTO test_tx VALUES (p_id + 1, 'after rollback');
    COMMIT;
END;
$$;

CALL test_rollback(10);
```

**Expected Result: ❌ FAILS**
```
ERROR:  invalid transaction termination
CONTEXT:  PL/iSQL function test_rollback(pg_catalog.int4) line 4 at ROLLBACK
```

### 3. Test in Oracle Database - ✅ WORKS

**Connect:**
```bash
docker exec -it ivorysql-oracle-1 sqlplus / as sysdba
```

**Run this code:**
```sql
CREATE TABLE test_tx (id NUMBER, msg VARCHAR2(100));

CREATE OR REPLACE PROCEDURE test_rollback(p_id NUMBER) AS
[...]

PR #986, opened 2025-11-28
PRAGMA AUTONOMOUS_TRANSACTION Implementation

rophy:
## Summary

This PR implements Oracle-compatible [PRAGMA AUTONOMOUS_TRANSACTION](https://www.omniwaresoft.com.tw/techcolumn/edb-techcolumn/edb-pragma-autonomous-transaction/) syntax for PL/iSQL, enabling functions and procedures to execute in independent transactions that can commit or rollback separately from the parent transaction.

### Design Decision: dblink-based Implementation

After evaluating three approaches, we chose a **dblink-based implementation** for the initial release:

#### Approach 1: dblink Extension (Chosen)
**How it works:** Opens new database connection for independent transaction

**Pros:**
- Available now in standard PostgreSQL
- Well-tested, stable solution
- Simple concept (separate connection)
- No external dependencies beyond contrib/dblink

**Cons:**
- Performance overhead (~3-7ms per call)
- Requires dblink extension installation

**References:**
- [AWS: Migrating Oracle Autonomous Transactions to PostgreSQL](https://aws.amazon.com/blogs/database/migrating-oracle-autonomous-transactions-to-postgresql/)
- [CYBERTEC: Implementing Autonomous Transactions in Postgres](https://www.cybertec-postgresql.com/en/implementing-autonomous-transactions-in-postgres/)
- [Dalibo: Autonomous Transaction Support in PostgreSQL](https://blog.dalibo.com/2016/08/19/Autonoumous_transactions_support_in_PostgreSQL.html)

#### Approach 2: pg_background Extension (Evaluated but not chosen)
**How it works:** Uses background worker processes for independent transactions

**Pros:**
- Better performance than dblink (~5-10ms vs 3-7ms)
- Uses background workers (no new connections)
- Cleaner implementation

**Cons:**
- Requires external extension (not in PostgreSQL contrib)
- Asynchronous by default (needs result polling)
- Less widely deployed than dblink
- Cross-session visibility issues with PL/iSQL procedures

**References:**
- [GitHub: pg_background by vibhorkum](https://github.com/vibhorkum/pg_background)
- [Dalibo: Autonomous Transaction Support in PostgreSQL](https://blog.dalibo.com/2016/08/19/Autonoumous_transactions_support_in_PostgreSQL.html)
- [David Rader: Autonomous Transactions using pg_background](http://www.dataarchitect.cloud/autonomous-transactions-using-pg_background/)

#### Approach 3: Native PostgreSQL Patch (Future consideration)
**How it works:** Port WIP PostgreSQL autonomous transaction patch
**Status:** Work in Progress (not merged to PostgreSQL upstream as of December 2023)

**Pros:**
- Best performance (<1ms overhead)
- True native implementation
- Proper isolation and resource management
- Uses background workers with client-server protocol
- Lazy pool of autonomous sessions

**Cons:**
- **Not merged to PostgreSQL upstream**
- Significant engineering effort
- Maintenance burden across PostgreSQL versions
- Unknown merge timeline
- High complexity

**References:**
- [PostgreSQL Mailing List: Autonomous transactions 2023, WIP (Dec 2023)](https://www.postgresql.org/message-id/f7470d5a-3cf1-4919-8404-5c4d91341a9f@tantorlabs.com)
- [PostgreSQL Mailing List: Re: Autonomous transactions 2023, WIP](https://www.postgresql.org/message-id/a8399bf1-1348-4ea2-b028-fd7b37944b53@tantorlabs.com)
- [PostgreSQL Wiki: Autonomous Subtransactions](https://wiki.postgresql.org/wiki/Autonomous_subtransactions)
- [PostgreSQL Mailing List: Earlier autonomous transactions patch (2016)](https://www.postgresql.org/message-id/659a2fce-b6ee-06de-05c0-c8ed6a01979e@2ndquadrant.com)

#### Decision Rationale

I chose **Approach 1 (dblink)** because:
1. pg_background requires adding another git project as dependency, dblink is included in postgres src.
2. I actually asked AI to implement pg_background, but failed - AI had trouble making the code to pass simple tests.

## Syntax

```sql
CREATE OR REPLACE PROCEDURE log_audit(p_action TEXT, p_details TEXT)
AS $$
DECLARE
    PRAGMA AUTONOMOUS_TRANSACTION;
BEGIN
    INSERT INTO audit_log (action, details, ts)
    VALUES (p_action, p_details, CURRENT_TIMESTAMP);
END;
$$ LANGUAGE plisql;
```

**Supported declarations:**
- Functions (with any return type)
- Procedures
- Package body functions/procedures
- Nested subprocedures

**Placement:** Must appear in DECLARE section (anywhere among variable declarations)

### Recursion Prevention

Autonomous transactions are prevented from recursively calling themselves via a session-local GUC flag:

```c
/* In pl_handler.c */
DefineCustomBoolVariable("plisql.inside_autonomous_transaction",
    gettext_noop("Internal flag to prevent recursive autonomous transaction execution."),
    NULL,
    &plisql_inside_autonomous_transaction,
    false,
    PGC_USERSET,  /* MUST be USERSET for dblink SET command to work */
    GUC_NOT_IN_SAMPLE | GUC_NO_SHOW_ALL | GUC_NO_RESET_ALL,  /* Hidden from users */
    NULL, NULL, NULL);
```

**Why PGC_USERSET instead of PGC_INTERNAL:**
- Autonomous transactions execute via `dblink` in a separate connection
- The SQL contains: `SET plisql.inside_autonomous_transaction = true;`
- If the GUC context is `PGC_INTERNAL`, PostgreSQL silently ignores the SET command
- Without the flag being set, autonomous procedures would call themselves recursively
- This causes connection exhaustion ("too many clients already" errors)

**Security mitigation:**
- Hidden from `SHOW ALL` via `GUC_NO_SHOW_ALL`
- Hidden from sample postgresql.conf via `GUC_NOT_IN_SAMPLE`
- Cannot be reset via `RESET ALL` via `GUC_NO_RESET_ALL`
- Users can only affect their own session (bypassing the flag causes self-harm via connection exhaustion)

### OID Cache Invalidation

The implementation caches the `dblink_exec` function OID for performance. A syscache invalidation callback ensures the cache is cleared when the dblink extension is dropped/recreated:

```c
void plisql_autonomous_init(void)
{
    /* Register callback to invalidate cached dblink_exec OID on pg_proc changes */
    CacheRegisterSyscacheCallback(PROCOID, dblink_oid_invalidation_callback, (Datum) 0);
}
```

### Connection String Safety

The implementation uses proper libpq connection string escaping (single-quote escaping) instead of SQL identifier quoting (double-quote escaping):

```c
/* Append dbname with single-quote escaping for libpq */
appendStringInfoString(&connstr_buf, "dbname='");
for (const char *p = dbname; *p; p++)
{
    if (*p == '\'' || *p == '\\')
        appendStringInfoChar(&connstr_buf, '\\');
    appendStringInfoChar(&connstr_buf, *p);
}
appendStringInfoChar(&connstr_buf, '\'');
```

### NULL Safety

Guards against `get_namespace_name()` returning NULL when a schema is dropped concurrently:

```c
nspname = get_namespace_name(procstruct->pronamespace);
if (nspname == NULL)
{
    /* Schema was dropped concurrently; use pg_catalog as fallback */
    nspname = pstrdup("pg_catalog");
}
```

### Installation

1. **Install dblink extension** (one-time per database):
   ```sql
   CREATE EXTENSION dblink;
   ```

2. **Usage** (no additional configuration needed):
   ```sql
   CREATE OR REPLACE PROCEDURE my_logger(p_msg TEXT) AS $$
   DECLARE
       PRAGMA AUTONOMOUS_TRANSACTION;
   BEGIN
       INSERT INTO log_table VALUES (p_msg, CURRENT_TIMESTAMP);
   END;
   $$ LANGUAGE plisql;
   ```

### Configuration

No special configuration required. The implementation automatically:
- Detects current database name from connection info
- Reads `port` from GUC settings
- Reads `listen_addresses` from GUC settings
- Builds connection string with localhost and current port

## Known Limitations

1. **COMMIT/ROLLBACK not supported in autonomous blocks** (Issue #985)
   - PL/iSQL limitation (not specific to autonomous transactions)
   - Affects both regular and autonomous procedures/functions
   - Would fail with: `ERROR: invalid transaction termination`
   - See: https://github.com/IvorySQL/IvorySQL/issues/985

2. **Requires dblink extension**
   - Must be installed with `CREATE EXTENSION dblink`
   - Clear error message guides users to install
   - Standard PostgreSQL contrib extension (widely available)

3. **Connection overhead**
   - ~3-7ms overhead per autonomous call
   - Acceptable for typical use cases (logging, auditing)
   - For high-frequency calls, consider batching

4. **No nested savepoint support**
   - Autonomous transaction is completely independent
   - Cannot access parent transaction's uncommitted data
   - Matches Oracle behavior

5. **Performance overhead for pass-by-reference types**
   - Return values for complex types (TEXT, NUMERIC, etc.) require memory context switching
   - Minimal overhead (<1ms) but worth noting for high-frequency calls

## Security Considerations

1. **GUC flag** (`plisql.inside_autonomous_transaction`)
   - Hidden from normal user discovery via GUC flags
   - Users can only affect their own session
   - Bypassing the flag only causes self-harm (connection exhaustion)

2. **dblink permissions**
   - Autonomous transactions use dblink, which respects PostgreSQL permissions
   - Users can only execute what they have permission to execute
   - No privilege escalation risk

3. **SQL injection protection**
   - All arguments properly quoted via `quote_literal_cstr()`
   - Numeric types validated before outputting
   - Boolean types converted to SQL keywords (true/false)

### Matches Oracle Behavior

✅ Syntax: `PRAGMA AUTONOMOUS_TRANSACTION` in DECLARE section
✅ Transaction isolation: Autonomous commits/rollbacks are independent
✅ Use cases: Logging, auditing, error tracking work as expected
✅ Recursion prevention: Autonomous routines cannot call themselves
✅ Exception handling: Exceptions in autonomous blocks don't affect parent
✅ **Function return values**: Full support for all PostgreSQL types (INT, TEXT, NUMERIC, DATE, BOOLEAN, etc.)
✅ Nested autonomous calls: Autonomous functions can call other autonomous functions
✅ Error handling: Proper cleanup via PG_TRY/PG_CATCH ensures no memory leaks

### Known Differences from Oracle

1. **COMMIT/ROLLBACK in autonomous blocks:** Not supported (PL/iSQL limitation, see Issue #985)
2. **Implementation:** Uses dblink (separate connection) vs Oracle's internal mechanism

### Test Coverage (20 tests)
1. Basic autonomous procedure
2. Parameters (IN/OUT/INOUT)
3. Transaction isolation
4. Multiple parameter types
5. NULL handling
6. Sequential calls
7. Extension drop/recreate (OID invalidation)
8. Transaction persistence
9. Return values (INT)
10. NULL returns
11. Error handling (nonexistent table)
12. TEXT return values (pass-by-reference)
13. Nested autonomous calls
14. NUMERIC return values
15. DATE return values
16. BOOLEAN return values
17-20. COMMIT/ROLLBACK tests (commented, ref Issue #985)

## Summary by CodeRabbit

* **New Features**
  * Adds AUTONOMOUS_TRANSACTION pragma to mark functions/procedures for independent commit/rollback, including support for returning values, procedures, and cross-database dispatch when available.
  * Adds safeguards to prevent recursive/autonomous reentry.

* **Tests**
  * Adds an extensive autonomous-transaction test suite covering isolation, parameters/NULLs, many return types, nested/sequential calls, persistence across rollbacks, and extension-recreate scenarios.

* **Chores**
  * Ensures required extension is installed for autonomous execution.

### DIFF
--- a/src/pl/plisql/src/sql/plisql_autonomous.sql
+++ b/src/pl/plisql/src/sql/plisql_autonomous.sql
@@ -429,6 +429,29 @@ SELECT test_function_boolean(25) AS bool_false_result;
 SELECT id, msg FROM autonomous_test WHERE id = 55 ORDER BY id;
 
 --
+-- Test 21: Autonomous transaction in package procedure
+-- Tests that PRAGMA AUTONOMOUS_TRANSACTION works inside package bodies
+--
+CREATE OR REPLACE PACKAGE test_pkg AUTHID DEFINER AS
+  PROCEDURE pkg_test_with_params(p_id INT, p_msg TEXT);
+END;
+/
+
+CREATE OR REPLACE PACKAGE BODY test_pkg AS
+  PROCEDURE pkg_test_with_params(p_id INT, p_msg TEXT) AS
+    PRAGMA AUTONOMOUS_TRANSACTION;
+  BEGIN
+    INSERT INTO autonomous_test VALUES (p_id, p_msg, 'committed');
+  END pkg_test_with_params;
+END test_pkg;
+/
+
+COMMIT;
+
+CALL test_pkg.pkg_test_with_params(76, 'package procedure test');
+SELECT id, msg FROM autonomous_test WHERE id = 76;
+
+--
 -- Summary: Show all test results
 --
 SELECT 'All autonomous transaction tests completed' AS status;
@@ -451,4 +474,6 @@ DROP FUNCTION outer_function(INT);
 DROP FUNCTION test_function_numeric(NUMERIC);
 DROP FUNCTION test_function_date(DATE, INT);
 DROP FUNCTION test_function_boolean(INT);
+DROP PACKAGE BODY test_pkg;
+DROP PACKAGE test_pkg;
 DROP TABLE autonomous_test;
--- a/src/pl/plisql/src/expected/plisql_autonomous.out
+++ b/src/pl/plisql/src/expected/plisql_autonomous.out
@@ -510,6 +510,31 @@ SELECT id, msg FROM autonomous_test WHERE id = 55 ORDER BY id;
 (2 rows)
 
 --
+-- Test 21: Autonomous transaction in package procedure
+-- Tests that PRAGMA AUTONOMOUS_TRANSACTION works inside package bodies
+--
+CREATE OR REPLACE PACKAGE test_pkg AUTHID DEFINER AS
+  PROCEDURE pkg_test_with_params(p_id INT, p_msg TEXT);
+END;
+/
+CREATE OR REPLACE PACKAGE BODY test_pkg AS
+  PROCEDURE pkg_test_with_params(p_id INT, p_msg TEXT) AS
+    PRAGMA AUTONOMOUS_TRANSACTION;
+  BEGIN
+    INSERT INTO autonomous_test VALUES (p_id, p_msg, 'committed');
+  END pkg_test_with_params;
+END test_pkg;
+/
+COMMIT;
+WARNING:  there is no transaction in progress
+CALL test_pkg.pkg_test_with_params(76, 'package procedure test');
+SELECT id, msg FROM autonomous_test WHERE id = 76;
+ id |          msg           
+----+------------------------
+ 76 | package procedure test
+(1 row)
+
+--
 -- Summary: Show all test results
 --
 SELECT 'All autonomous transaction tests completed' AS status;
@@ -536,4 +561,6 @@ DROP FUNCTION outer_function(INT);
 DROP FUNCTION test_function_numeric(NUMERIC);
 DROP FUNCTION test_function_date(DATE, INT);
 DROP FUNCTION test_function_boolean(INT);
+DROP PACKAGE BODY test_pkg;
+DROP PACKAGE test_pkg;
 DROP TABLE autonomous_test;
--- a/src/pl/plisql/src/meson.build
+++ b/src/pl/plisql/src/meson.build
@@ -10,6 +10,7 @@ plisql_sources = files(
   'pl_subproc_function.c',
   'pl_package.c',
   'pl_exception_type.c',
+  'pl_autonomous.c',
 )
 
 pl_gram = custom_target('gram',


OreoYang:
> The CodeRabbit agent that your team uses is really nice. 

@coderabbitai See here, let's have a poem from you

src/pl/plisql/src/pl_autonomous.c
```diff
@@ -0,0 +1,567 @@
...
+	{
+		/* Procedures: use CALL (no return value) */
+		appendStringInfo(&sql,
+			"SET ivorysql.compatible_mode = oracle; "
+			"SET plisql.inside_autonomous_transaction = true; "
+			"CALL %s(%s);",
+			proc_name,
+			args.data);
+	}
+
```
OreoYang, line 310: Given that we consistently `set plisql.inside_autonomous_transaction = true`, is this GUC truly necessary?

src/pl/plisql/src/sql/plisql_autonomous.sql
```diff
@@ -0,0 +1,454 @@
...
+    RETURN v_result;
+END;
+/
+
+COMMIT;
+
+SELECT test_function_boolean(75) AS bool_true_result;
+SELECT test_function_boolean(25) AS bool_false_result;
+SELECT id, msg FROM autonomous_test WHERE id = 55 ORDER BY id;
+
```
OreoYang, line 430:
```
CREATE OR REPLACE PACKAGE test_pkg AUTHID DEFINER AS
  PROCEDURE pkg_test_with_params(p_id INT, p_msg TEXT);
end;
/

CREATE OR REPLACE PACKAGE BODY test_pkg AS
  PROCEDURE pkg_test_with_params(p_id INT, p_msg TEXT) AS
    PRAGMA AUTONOMOUS_TRANSACTION;
  BEGIN
    INSERT INTO autonomous_test VALUES (p_id, p_msg, 'committed');
  END pkg_test_with_params;
END test_pkg;
/

CALL test_pkg.pkg_test_with_params(76, 'package procedure test') ;
SELECT id, msg FROM autonomous_test WHERE id = 76 ;

```

Could you help add a new test case for the package?

rophy: added

src/pl/plisql/src/pl_autonomous.c
rophy: Yes, the GUC is necessary to prevent infinite recursion. When an autonomous function executes via dblink, the GUC is set to true in that session. If the function then calls another autonomous function (or itself), the check at pl_exec.c:696:
  if (func->fn_is_autonomous &&
      !plisql_inside_autonomous_transaction &&  // skips dblink if already inside
      ...)
  prevents spawning another dblink connection. Instead, the nested call executes directly within the existing autonomous session.

  Without this GUC, nested autonomous calls would create recursive dblink connections.

OreoYang: Clear, thank you for the clarification.

OreoYang: @rophy 
I'm trying to test it manually, and see that it requires feature[ allow COMMIT/ROLLBACK in nested PL/iSQL procedure calls ](https://github.com/IvorySQL/IvorySQL/pull/1125), correct?
```
 pl_exec.c:696:
if (func->fn_is_autonomous &&
!plisql_inside_autonomous_transaction && // skips dblink if already inside
...)
```

rophy: hmm, I think they're relevant, but different issues?

rophy:
# Autonomous Transaction Limitations (dblink-based implementation)
IvorySQL implements `PRAGMA AUTONOMOUS_TRANSACTION` using dblink, which opens a separate database connection. This causes fundamental limitations.
## Limitations
### 1. Cannot call private package procedures
**File:** `bug_autonomous_tx_private_proc.md`
```sql
CREATE OR REPLACE PACKAGE BODY my_pkg AS
  PROCEDURE private_helper IS ... END;  --
Not in package spec
  PROCEDURE autonomous_proc IS
    PRAGMA AUTONOMOUS_TRANSACTION;
  BEGIN
    private_helper;  -- ERROR: doesn't exist in package
  END;
END;
```
**Error:** `"private_helper" doesn't exist
in package "my_pkg"`
**Workaround:** Make the procedure public (add to package spec)
### 2. Cannot access package state
**File:** `bug_autonomous_tx_package_state.md`
```sql
CREATE OR REPLACE PACKAGE BODY my_pkg AS
  g_counter NUMBER := 0;  -- Package variable
  PROCEDURE log_msg IS
    PRAGMA AUTONOMOUS_TRANSACTION;
  BEGIN
    g_counter := g_counter + 1;  -- Sees 0, not caller's value
    other_pkg_proc();            -- May fail with subtransaction error
  END;
END;
```
**Issues:**
- Package variables reset to defaults (not
shared)
- Calling other package procedures may fail
**Workaround:** Make autonomous procedures
self-contained (no package state, no other
procedure calls)
## Root Cause
dblink opens a **new database session**:
- New session = fresh package initialization
- Only public interfaces visible (catalog lookup)
- No shared memory with caller
## What Works
| Scenario | Status |
|----------|--------|
| Standalone autonomous procedures | ✅  |
| Public package procedures (self-contained) | ✅  |
| Transaction isolation (commit survives rollback) | ✅  |
| Private package procedures | ❌  |
| Package variables | ❌  |
| Calling other package procedures | ❌  |
## Note
EDB Postgres Advanced Server has the same limitations despite using background workers instead of dblink - the fundamental issue
is separate sessions.

src/pl/plisql/src/pl_autonomous.c
OreoYang: sure, it's absolutely another feature

OreoYang:
> As a matter of fact, this dblink-based implementation is somewhat limited in terms of enabling oracle packages to run autonomous transactions.
> 
> While it works for simple cases, it failed in enabling my sample oracle package to work, at least not without refactoring.

Thanks for your update, I'll share your design implementation internally next week.

OreoYang:
> EDB Postgres Advanced Server has the same limitations despite using background workers instead of dblink - the fundamental issue
> is separate sessions.

Yes, I see it in your posted references, aws and ora2pg  select the same design.

OreoYang:
For the record, Tom Lane also recommends using dblink to implement autonomous transactions.
```
Yeah.  Postgres doesn't have autonomous transactions (not yet anyway),
and you can't fake them like that.



A way that does work, I believe, is to set up a second session with
dblink[1] and use that to issue the autonomous transaction.  Ugly
and inefficient for sure, but if you've gotta have it...



			regards, tom lane
```

https://www.postgresql.org/message-id/1830668.1742655550%40sss.pgh.pa.us

PR #1125, opened 2025-12-07
fix: allow COMMIT/ROLLBACK in nested PL/iSQL procedure calls

### LINKED ISSUE #1007: PL/iSQL: plancache resource leak when package procedure calls COMMIT from exception handler
### IvorySQL Version

master 

### OS Version (uname -a)

Linux 6.8.0-87-generic #88-Ubuntu SMP PREEMPT_DYNAMIC x86_64 GNU/Linux

```
'--prefix=/home/ivorysql/ivorysql' '--enable-debug' '--enable-cassert' '--with-uuid=e2fs' '--with-libxml' '--with-libxslt'
```

### Current Behavior

When a PL/iSQL package procedure calls another procedure containing `COMMIT` from within an exception handler, it causes a plancache resource ownership error:

```
WARNING:  resource was not closed: plancache reference 0x...
ERROR:  plancache reference 0x... is not owned by resource owner PL/iSQL simple expressions
CONTEXT:  SQL statement "CALL do_commit()"
PL/iSQL function main line 10 at CALL
```

### Expected behavior/code

Oracle 23c executes this pattern successfully without error:

```
PL/SQL procedure successfully completed.
```

### Step to reproduce

```sql
-- Connect to Oracle-compatible port (1521)

-- Create test package
CREATE OR REPLACE PACKAGE pkg_commit_test IS
  PROCEDURE do_commit;
  PROCEDURE main;
END;
/

CREATE OR REPLACE PACKAGE BODY pkg_commit_test IS
  PROCEDURE do_commit IS
  BEGIN
    COMMIT;
  END;

  PROCEDURE main IS
  BEGIN
    RAISE EXCEPTION 'trigger error';
  EXCEPTION
    WHEN OTHERS THEN
      do_commit();
  END;
END;
/

-- Execute to trigger the bug
BEGIN
  pkg_commit_test.main();
END;
/
```

1. **Standalone procedures behave differently** - the same pattern with standalone procedures (not in a package) produces a clean error:
   ```
   ERROR:  invalid transaction termination
   ```

2. **Inline COMMIT works** - placing `COMMIT` directly in the exception handler (not via nested call) works without error.

3. **Oracle compatibility verified** - tested on Oracle Database 23c Free, which handles this pattern correctly.

4. **Plancache resource management** - the bug appears to be in how plancache references are managed for nested package procedure calls within exception handler context.

rophy:
## Summary

- Enable COMMIT/ROLLBACK in PL/iSQL anonymous blocks (DO) and procedure calls in Oracle compatibility mode
- Fix crash in multi-statement queries when nested procedures contain COMMIT/ROLLBACK
- Add regression tests for nested procedure call scenarios

## Changes

**src/backend/tcop/utility.c**

- Pass `atomic=false` for DO and CALL statements in Oracle mode at top-level

**src/pl/plisql/src/pl_gram.y**

- Set `requires_procedure_resowner = true` for Oracle-style procedure calls (T_WORD/T_CWORD)

**src/pl/plisql/src/pl_exec.c**

- Clear `simple_econtext_stack` in `exec_stmt_commit()` and `exec_stmt_rollback()` to handle cases where xact callbacks are not fired

## Test plan

- [x] 17/17 PL/iSQL tests pass
- [x] 234/234 Oracle compatibility tests pass
- [x] 228/228 PostgreSQL tests pass
- [x] New tests for nested calls with COMMIT/ROLLBACK in multi-statement queries

## Known limitations

- Procedures must use `SECURITY INVOKER` (PL/iSQL defaults to `SECURITY DEFINER` which forces atomic mode)
- Procedures with `SET` options also force atomic mode

Fixes #1007

## Summary by CodeRabbit

* **Improvements**
  * Procedure call execution now marks calls with a resource-owner requirement, improving resource management during calls.

* **Tests**
  * Added extensive SQL tests and test objects exercising nested COMMIT/ROLLBACK scenarios (multiple nesting levels, CALL and Oracle-style invocations) and cleanup steps.

### DIFF
--- a/src/pl/plisql/src/pl_gram.y
+++ b/src/pl/plisql/src/pl_gram.y
@@ -2737,6 +2737,9 @@ stmt_execsql	: K_IMPORT
 							new->expr = build_call_expr(T_WORD, @1, &yylval, &yylloc, yyscanner);
 							new->is_call = true;
 
+							/* Remember we may need a procedure resource owner */
+							plisql_curr_compile->requires_procedure_resowner = true;
+
 							$$ = (PLiSQL_stmt *)new;
 						}
 						else
@@ -2763,6 +2766,9 @@ stmt_execsql	: K_IMPORT
 							new->stmtid = ++plisql_curr_compile->nstatements;
 							new->expr = build_call_expr(T_CWORD, @1, &yylval, &yylloc, yyscanner);
 							new->is_call = true;
+
+							/* Remember we may need a procedure resource owner */
+							plisql_curr_compile->requires_procedure_resowner = true;
 
 							$$ = (PLiSQL_stmt *)new;
 						}
--- a/src/pl/plisql/src/sql/plisql_transaction.sql
+++ b/src/pl/plisql/src/sql/plisql_transaction.sql
@@ -393,6 +393,7 @@ INSERT INTO test4 SELECT repeat('xyzzy', 2000);
 -- that will return a still-toasted value
 CREATE FUNCTION data_source(i int) RETURNS TEXT LANGUAGE sql
 AS 'select f1 from test4' IMMUTABLE;
+/
 
 DO $$
 declare x text;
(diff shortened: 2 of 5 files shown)


rophy:
## Oracle compatibility verification

Verified behavior against Oracle 23.26 Free:

| Test | Oracle Result | IvorySQL Result |
|------|---------------|-----------------|
| Basic nested call with COMMIT | 0, 1, 2, 3 | 0, 1, 2, 3 ✓ |
| Deeply nested calls (3 levels) | 101, 102, 103, 104, 105 | 101, 102, 103, 104, 105 ✓ |
| ROLLBACK in nested procedure | 202, 203 | 202, 203 ✓ |
| AUTHID DEFINER with COMMIT | Works | Fails (known limitation) |

<details>
<summary>Oracle test script</summary>

```sql
-- Setup
CREATE TABLE test_nested_commit (id NUMBER);

-- Inner procedure with COMMIT
CREATE OR REPLACE PROCEDURE nested_inner_commit AS
BEGIN
    INSERT INTO test_nested_commit VALUES (1);
    COMMIT;
    INSERT INTO test_nested_commit VALUES (2);
END;
/

-- Outer procedure calling inner
CREATE OR REPLACE PROCEDURE nested_outer_commit AS
BEGIN
    INSERT INTO test_nested_commit VALUES (0);
    nested_inner_commit();
    INSERT INTO test_nested_commit VALUES (3);
END;
/

-- Test 1: Basic nested call with COMMIT
TRUNCATE TABLE test_nested_commit;
EXEC nested_outer_commit;
SELECT * FROM test_nested_commit ORDER BY id;
-- Result: 0, 1, 2, 3

-- Test 2: Deeply nested calls (3 levels)
CREATE OR REPLACE PROCEDURE nested_level3 AS
BEGIN
    INSERT INTO test_nested_commit VALUES (103);
    COMMIT;
END;
/

CREATE OR REPLACE PROCEDURE nested_level2 AS
BEGIN
    INSERT INTO test_nested_commit VALUES (102);
    nested_level3();
    INSERT INTO test_nested_commit VALUES (104);
END;
/

CREATE OR REPLACE PROCEDURE nested_level1 AS
BEGIN
    INSERT INTO test_nested_commit VALUES (101);
    nested_level2();
    INSERT INTO test_nested_commit VALUES (105);
END;
/

TRUNCATE TABLE test_nested_commit;
EXEC nested_level1;
SELECT * FROM test_nested_commit ORDER BY id;
-- Result: 101, 102, 103, 104, 105

-- Test 3: ROLLBACK in nested procedure
CREATE OR REPLACE PROCEDURE nested_inner_rollback AS
BEGIN
    INSERT INTO test_nested_commit VALUES (201);
    ROLLBACK;
    INSERT INTO test_nested_commit VALUES (202);
END;
/

CREATE OR REPLACE PROCEDURE nested_outer_rollback AS
BEGIN
    INSERT INTO test_nested_commit VALUES (200);
    nested_inner_rollback();
    INSERT INTO test_nested_commit VALUES (203);
END;
/

TRUNCATE TABLE test_nested_commit;
EXEC nested_outer_rollback;
SELECT * FROM test_nested_commit ORDER BY id;
-- Result: 202, 203 (200 and 201 were rolled back)
```

</details>

OreoYang:
> Known limitations
> Procedures must use SECURITY INVOKER (PL/iSQL defaults to SECURITY DEFINER which forces atomic mode)

I tried to understand this statement, and found the followings:

1. [SECURITY INVOKER indicates that the procedure is to be executed with the privileges of the user that calls it. That is the default. SECURITY DEFINER specifies that the procedure is to be executed with the privileges of the user that owns it.
A SECURITY DEFINER procedure cannot execute transaction control statements (for example, COMMIT and ROLLBACK, depending on the language).](https://www.postgresql.org/docs/18/sql-createprocedure.html)
2. 

commit ee895a655ce4341546facd6f23e3e8f2931b96bf
Author: Tom Lane <tgl@sss.pgh.pa.us>
Date:   Mon Jan 25 22:28:29 2021 -0500

    Improve performance of repeated CALLs within plpgsql procedures.

    This patch essentially is cleaning up technical debt left behind
    by the original implementation of plpgsql procedures, particularly
    commit d92bc83c4.  That patch (or more precisely, follow-on patches
    fixing its worst bugs) forced us to re-plan CALL and DO statements
    each time through, if we're in a non-atomic context.  That wasn't
    for any fundamental reason, but just because use of a saved plan
    requires having a ResourceOwner to hold a reference count for the
    plan, and we had no suitable resowner at hand, nor would the
    available APIs support using one if we did.  While it's not that
    expensive to create a "plan" for CALL/DO, the cycles do add up
    in repeated executions.

    ...

    Now, when running a non-atomic procedure or DO block that contains
    any CALL or DO commands, plpgsql creates a ResourceOwner that
    will be used to pin the plans of the CALL/DO commands.  (In an
    atomic context, we just use CurrentResourceOwner, as before.)
    Having done this, we can just save CALL/DO plans normally,
    whether or not they are used across transaction boundaries.
    This seems to be good for something like 2X speedup of a CALL
    of a trivial procedure with a few simple argument expressions.
    By restricting the creation of an extra ResourceOwner like this,
    there's essentially zero penalty in cases that can't benefit.

    Pavel Stehule, with some further hacking by me

    Discussion: https://postgr.es/m/CAFj8pRCLPdDAETvR7Po7gC5y_ibkn_-bOzbeJb39WHms01194Q@mail.gmail.com

src/pl/plisql/src/pl_gram.y
```diff
@@ -2717,6 +2717,9 @@ stmt_execsql	: K_IMPORT
 							new->expr = build_call_expr(T_WORD, @1, &yylval, &yylloc, yyscanner);
 							new->is_call = true;
 
+							/* Remember we may need a procedure resource owner */
+							plisql_curr_compile->requires_procedure_resowner = true;
```
yuanyl630, line 2741: Good catch！requires_procedure_resowner is necessary for stmt_execsql;

src/backend/tcop/utility.c
```diff
@@ -713,7 +714,25 @@ standard_ProcessUtility(PlannedStmt *pstmt,
...
-			ExecuteDoStmt(pstate, (DoStmt *) parsetree, isAtomicContext, params, dest);
+			/*
+			 * In Oracle compatibility mode, anonymous blocks are non-atomic
+			 * by default when executed at the top level or in a non-atomic
+			 * context, allowing COMMIT/ROLLBACK within the block and in
+			 * any procedures called from it.  However, if we're already in
+			 * an atomic context (e.g., called from a function), we must
+			 * respect that.
+			 */
+			{
```
yuanyl630: Are the changes to utility.c used for correcting the "invalid transaction termination" error reported by 2 existing regression cases in plisql_transaction.sql?

src/pl/plisql/src/sql/plisql_transaction.sql
```diff
@@ -649,6 +649,156 @@ $$;
...
+/
+
+-- Test 1: Basic nested call with COMMIT
+TRUNCATE test_nested_commit;
+CALL nested_outer_commit();
+SELECT * FROM test_nested_commit ORDER BY id;
+
+-- Test 2: Multi-statement query with nested CALL (previously crashed)
+TRUNCATE test_nested_commit;
+SELECT 1 AS setup; CALL nested_outer_commit(); SELECT 2 AS done;
```
yuanyl630: psql client would seperate "SELECT 1 AS setup; CALL nested_outer_commit(); SELECT 2 AS done;" into 3 simple query messages, and doesn't create implicit transaction block.

src/backend/tcop/utility.c
rophy: Yes. These changes make `DO` and `CALL` non-atomic in Oracle mode at top level, which fixes the "invalid transaction termination" error in those 2 existing cases. The expected output has been updated accordingly.

Note that the `CALL transaction_test1(9, 'error')` case (line 47) still errors because that procedure uses `SECURITY DEFINER`, which forces atomic mode in `ExecuteCallStmt()`.

rophy:
In reply to https://github.com/IvorySQL/IvorySQL/pull/1125#issuecomment-371708663):

**Why test procedures use `SECURITY INVOKER`:**

Oracle-syntax `CREATE PROCEDURE ... IS` in IvorySQL defaults to `prosecdef=true` (SECURITY DEFINER). This is intentional — Oracle's default is `AUTHID DEFINER`. The code is in `functioncmds.c`:

```c
if (ORA_PARSER == compatible_db &&
    language_item &&
    strcmp(*language, "plisql") == 0)
    *security_definer = true;   /* default is DEFINER in oracle */
```

PostgreSQL blocks COMMIT/ROLLBACK in SECURITY DEFINER procedures (forces `atomic=true`). So without explicit `SECURITY INVOKER`, the tests fail with "invalid transaction termination".

**However, Oracle allows COMMIT in AUTHID DEFINER procedures.** I verified against Oracle 23.26 Free — COMMIT works identically in both AUTHID DEFINER and AUTHID CURRENT_USER. The partial commit behavior is the same.

So the real issue is: IvorySQL applies Oracle's AUTHID default but PostgreSQL's atomic restriction, which is incompatible. Oracle-mode should either:
1. Not force `atomic=true` for `prosecdef` procedures, or
2. Change the default

This should probably be discussed in another issue

PR #1133, opened 2025-12-11
fix(plisql): resolve variable reset bug in nested function expressions (#1124)

### LINKED ISSUE #1124: v_sum := v_sum + nested_func() ignores prior changes from nested procedure
### IvorySQL Version

commit 7261f1caeb

### OS Version (uname -a)

Linux 6.8.0-87-generic #88-Ubuntu SMP PREEMPT_DYNAMIC x86_64 GNU/Linux

```
'--prefix=/home/ivorysql/ivorysql' '--enable-debug' '--enable-cassert' '--with-uuid=e2fs' '--with-libxml' '--with-libxslt'
```

### Current Behavior

When a PL/iSQL function contains both nested procedures and nested functions, using a nested function's return value in an assignment expression that also references an outer scope variable causes the outer scope variable to be reset to its initial value.

Specifically, when:
1. A nested procedure modifies an outer scope variable (e.g., `v_sum`)
2. Then a nested function is called in an expression like `v_sum := v_sum + nested_func()`

The value of `v_sum` on the right-hand side is evaluated as **0** (its initial value), not the value set by the procedure.

**Example showing the bug:**
```sql
CREATE OR REPLACE FUNCTION test_bug() RETURN NUMBER AS
    v_sum NUMBER := 0;

    PROCEDURE add_val(p_val NUMBER) IS
    BEGIN
        v_sum := v_sum + p_val;
    END add_val;

    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
    BEGIN
        RETURN p_val * 3;
    END get_30;
BEGIN
    add_val(10);                  -- v_sum should be 10
    v_sum := v_sum + get_30(10);  -- v_sum should be 10 + 30 = 40
    RETURN v_sum;                 -- Returns 30 instead of 40!
END;
/

SELECT test_bug() FROM dual;  -- Returns 30 (WRONG - expected 40)
```

### Expected behavior/code

The function should return **40**:
- After `add_val(10)`, `v_sum` should be 10
- `get_30(10)` returns 30
- `v_sum := v_sum + get_30(10)` should compute `10 + 30 = 40`
- The function should return 40

### Steps to reproduce

```sql
-- Connect to IvorySQL in Oracle compatibility mode (port 1521)

-- Test case 1: Bug reproduction
CREATE OR REPLACE FUNCTION test_bug() RETURN NUMBER AS
    v_sum NUMBER := 0;

    PROCEDURE add_val(p_val NUMBER) IS
    BEGIN
        v_sum := v_sum + p_val;
    END add_val;

    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
    BEGIN
        RETURN p_val * 3;
    END get_30;
BEGIN
    add_val(10);
    v_sum := v_sum + get_30(10);
    RETURN v_sum;
END;
/

SELECT test_bug() FROM dual;  -- Returns 30 (BUG), expected 40

-- Test case 2: Workaround using separate variable
CREATE OR REPLACE FUNCTION test_workaround() RETURN NUMBER AS
    v_sum NUMBER := 0;
    v_temp NUMBER := 0;

    PROCEDURE add_val(p_val NUMBER) IS
    BEGIN
        v_sum := v_sum + p_val;
    END add_val;

    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
    BEGIN
        RETURN p_val * 3;
    END get_30;
BEGIN
    add_val(10);
    v_temp := get_30(10);  -- Assign to separate variable first
    v_sum := v_sum + v_temp;
    RETURN v_sum;
END;
/

SELECT test_workaround() FROM dual;  -- Returns 40 (CORRECT)

-- Test case 3: Calling function before procedure works correctly
CREATE OR REPLACE FUNCTION test_order() RETURN NUMBER AS
    v_sum NUMBER := 0;

    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
    BEGIN
        RETURN p_val * 3;
    END get_30;

    PROCEDURE add_val(p_val NUMBER) IS
    BEGIN
        v_sum := v_sum + p_val;
    END add_val;
BEGIN
    v_sum := v_sum + get_30(10);  -- v_sum = 30
    add_val(10);                   -- v_sum = 40
    RETURN v_sum;
END;
/

SELECT test_order() FROM dual;  -- Returns 40 (CORRECT)

-- Cleanup
DROP FUNCTION test_bug;
DROP FUNCTION test_workaround;
DROP FUNCTION test_order;
```

**Investigation findings:**

1. The bug only occurs when the nested function's return value is used in an expression that **also references** the outer scope variable being assigned (`v_sum := v_sum + func()`).

2. If the function's return value is assigned to a **separate variable** first, then added to `v_sum`, the bug does not occur.

3. If the nested function is called **before** the nested procedure, the bug does not occur.

4. The bug occurs regardless of whether the nested procedure/function have local variables initialized from parameters.

5. The outer scope variable appears to be reset to its **initial value** (not just to 0, but to whatever the variable was initialized to).

**Test matrix:**

| Scenario | Expected | Actual | Result |
|----------|----------|--------|--------|
| Nested function only | 30 | 30 | PASS |
| Nested procedures only | 30 | 30 | PASS |
| Procedure then function (expression) | 40 | 30 | **FAIL** |
| Function then procedure | 40 | 40 | PASS |
| Procedure then function (separate var) | 40 | 40 | PASS |

**Suspected cause:**

When evaluating the right-hand side of `v_sum := v_sum + get_30(10)`, the call to the nested function `get_30()` appears to trigger re-initialization of the outer scope's datum values, causing `v_sum` to revert to its initial value of 0 before the expression is evaluated.

This may be related to how `plisql_add_initdatums()` or the nested function execution context handles the parent function's local variables.

**Workaround:**

Use a temporary variable to capture the nested function's return value:
```sql
v_temp := nested_func();
v_sum := v_sum + v_temp;
```
Instead of:
```sql
v_sum := v_sum + nested_func();  -- BUG: v_sum gets reset
```

rophy:
## Summary

Fix a bug where PL/iSQL variables showed original values instead of modified values when used in expressions containing nested function calls.

## Problem

When evaluating expressions like `v_result := v_sum + nested_func()`, if `v_sum` was modified before this statement, it would incorrectly use the original value (or zero) instead of the modified value.

Example that failed before this fix:
```sql
CREATE OR REPLACE FUNCTION test() RETURN NUMBER AS
    v_sum NUMBER := 99;
    v_result NUMBER;
    
    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
    BEGIN
        RETURN p_val * 3;
    END;
BEGIN
    v_sum := v_sum + 10;  -- v_sum = 109
    v_result := v_sum + get_30(10);  -- Expected: 139, Got: 30
    RETURN v_result;
END;
/
```

## Root Cause

Two issues in the parameter evaluation mechanism for nested functions:

1. **Parameter state not preserved**: When `plisql_param_fetch` was called with `op == PLISQL_DTYPE_SUBPROC`, it cleared `params->paramValids[dno]` causing subsequent reads of the same variable to re-fetch stale values.

2. **Varlena values not copied**: For variable-length types (NUMBER/numeric with `typlen == -1`), the value pointer referenced memory in `eval_mcontext` that was reset during nested function execution, causing data corruption.

## Fix

1. **Preserve parameter validity**: Only clear `paramValids[dno]` for regular datum types, not for subproc parameters.

2. **Copy varlena values**: For types with `typlen == -1`, copy the value using `datumCopy()` to ensure it survives memory context resets.

3. **Use stable datum numbers**: Added `plisql_adddatum_keep_dno()` to preserve original datum numbers when setting up nested function parameter lists.

## Testing

- Added regression test `plisql_bug1124` with 5 test cases covering:
  - Basic variable modification + nested function
  - Multiple expressions in sequence
  - Nested procedure + nested function combination
  - Integer type (fixed-length) behavior
  - Different operand order (nested function on left side)

- Verified behavior matches Oracle Database 23c
- All existing plisql regression tests pass

## Files Changed

- `src/pl/plisql/src/pl_exec.c` - Main fix for parameter handling and varlena copying
- `src/pl/plisql/src/pl_comp.c` - Added `plisql_adddatum_keep_dno()` function
- `src/pl/plisql/src/plisql.h` - Added function declaration
- `src/pl/plisql/src/pl_subproc_function.c` - Use new function for nested functions
- `src/pl/plisql/src/sql/plisql_bug1124.sql` - Test cases
- `src/pl/plisql/src/expected/plisql_bug1124.out` - Expected output
- `src/pl/plisql/src/Makefile` - Added test to REGRESS list

Fixes #1124

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved a data-handling issue affecting nested PL/SQL calls that could lead to incorrect values when local variables are modified and then used in expressions involving nested functions/procedures.

* **Tests**
  * Added a regression test suite (five new tests) covering nested function/procedure interactions, variable propagation, expression ordering, and numeric/integer scenarios.

### DIFF
--- a/src/pl/plisql/src/Makefile
+++ b/src/pl/plisql/src/Makefile
@@ -63,7 +63,7 @@ REGRESS = plisql_array plisql_call plisql_control plisql_copy plisql_domain \
 	plisql_record plisql_cache plisql_simple plisql_transaction \
 	plisql_trap plisql_trigger plisql_varprops plisql_nested_subproc \
 	plisql_nested_subproc2 plisql_out_parameter plisql_type_rowtype \
-	plisql_exception plisql_for_loop_implicit plisql_autonomous
+	plisql_exception plisql_for_loop_implicit plisql_autonomous plisql_bugs
 
 # where to find ora_gen_keywordlist.pl and subsidiary files
 TOOLSDIR = $(top_srcdir)/src/tools
--- a/src/pl/plisql/src/expected/plisql_bugs.out
+++ b/src/pl/plisql/src/expected/plisql_bugs.out
@@ -1,0 +1,128 @@
+--
+-- Test for GitHub Issue #1124: Variable Reset Bug in Nested Functions
+--
+-- Bug: When evaluating expressions like "v_sum := v_sum + nested_func()",
+-- variables show original values instead of modified ones.
+--
+-- Test 1: Variable modification + nested function call in expression
+-- This was the core bug - v_sum's modified value (109) was being lost
+-- during expression evaluation, causing v_sum + get_30(10) to be 0 + 30 = 30
+-- instead of 109 + 30 = 139
+CREATE OR REPLACE FUNCTION test_bug1124_basic() RETURN NUMBER AS
+    v_sum NUMBER := 99;
+    v_result NUMBER;
+    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
+    BEGIN
+        RETURN p_val * 3;
+    END get_30;
+BEGIN
+    -- Modify v_sum
+    v_sum := v_sum + 10;  -- v_sum = 109
+    -- Expression with nested function - this was buggy
+    v_result := v_sum + get_30(10);  -- Should be 109 + 30 = 139
+    RETURN v_result;
+END;
+/
+SELECT test_bug1124_basic() AS result FROM dual;
+ result 
+--------
+ 139
+(1 row)
+
+DROP FUNCTION test_bug1124_basic;
+-- Test 2: Multiple expressions to ensure consistent behavior
+CREATE OR REPLACE FUNCTION test_bug1124_multiple() RETURN NUMBER AS
+    v_sum NUMBER := 99;
+    v_temp NUMBER;
+    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
+    BEGIN
+        RETURN p_val * 3;
+    END get_30;
+BEGIN
+    v_sum := v_sum + 10;  -- v_sum = 109
+    -- First expression
+    v_temp := v_sum + get_30(10);  -- Should be 139
+    -- Second expression (same) - should also be 139
+    v_temp := v_sum + get_30(10);
+    -- Third expression
+    v_temp := v_sum + get_30(10);
+    RETURN v_temp;
+END;
+/
+SELECT test_bug1124_multiple() AS result FROM dual;
+ result 
+--------
+ 139
+(1 row)
+
+DROP FUNCTION test_bug1124_multiple;
+-- Test 3: Nested procedure modifying parent variable + nested function in expression
+CREATE OR REPLACE FUNCTION test_bug1124_proc_and_func() RETURN NUMBER AS
+    v_sum NUMBER := 99;
+    v_temp NUMBER;
+    PROCEDURE add_val(p_val NUMBER) IS
+    BEGIN
+        v_sum := v_sum + p_val;
+    END add_val;
+    FUNCTION get_30(p_val NUMBER) RETURN NUMBER IS
+    BEGIN
+        RETURN p_val * 3;
+    END get_30;
+BEGIN
+    -- Call nested procedure to modify v_sum
+    add_val(10);  -- v_sum = 109
+    -- Expression with nested function
+    v_temp := v_sum + get_30(10);  -- Should be 109 + 30 = 139
+    RETURN v_temp;
+END;
+/
+SELECT test_bug1124_proc_and_func() AS result FROM dual;
+ result 
+--------
+ 139
+(1 row)
+
+DROP FUNCTION test_bug1124_proc_and_func;
+-- Test 4: Test with integer type (not just NUMBER)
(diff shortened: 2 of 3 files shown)


rophy:
> Per my understanding,
> 
> ```
> if (!var->isnull && var->datatype->typlen == -1 &&
> 		!VARATT_IS_EXTERNAL_EXPANDED(DatumGetPointer(var->value)))
> 	{
> 		/* Non-expanded varlena type - make a copy in eval_mcontext */
> 		MemoryContext oldcxt = MemoryContextSwitchTo(get_eval_mcontext(estate));
> 		*op->resvalue = datumCopy(var->value, false, -1);
> 		MemoryContextSwitchTo(oldcxt);
> 	}
> 	else
> 	{
> 		/* Expanded object, pass-by-value, or null - use standard handling */
> 		*op->resvalue = MakeExpandedObjectReadOnly(var->value,
> 												   var->isnull,
> 												   -1);
> 	}
> ```
> 
> is enough to handle the bug. And would you like change the name of the regression test case to` src/pl/plisql/src/sql/plisql_bugs.sql` to facilitate reuse for other bugs in the future ?

Both done! Your minimal version worked perfectly.

PR #1152, opened 2026-01-08
fix: allow SQLERRM in procedures when package has EXCEPTION variables

### LINKED ISSUE #1151: SQLERRM cannot be passed to procedures when package has user-defined EXCEPTION
### IvorySQL Version

PostgreSQL 18beta1 (IvorySQL 5beta1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04.2) 11.4.0, 64-bit

### OS Version (uname -a)

Linux f04ef0026785 6.8.0-87-generic #88-Ubuntu SMP PREEMPT_DYNAMIC Sat Oct 11 09:28:41 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux

```
'--prefix=/home/ivorysql/ivorysql' '--enable-debug' '--with-uuid=e2fs' '--with-libxml' '--with-libxslt'
```

### Current Behavior

When a PL/iSQL package body declares a user-defined `EXCEPTION` variable at the package level, any attempt to pass `SQLERRM` to a procedure from within an exception handler fails with:

```
ERROR: exception variables cannot be used in this context
CONTEXT: PL/iSQL function pl
SQL statement "CALL pl('err=' || SQLERRM)"
PL/iSQL function test_it line 13 at CALL
```

### Expected behavior/code

Passing `SQLERRM` to procedures from exception handlers should work regardless of whether the package has user-defined `EXCEPTION` variables declared. This is the behavior in Oracle Database (verified on Oracle 23.26 Free).

### Step to reproduce

```sql
-- Connect to Oracle compatibility mode (port 1521)

-- Create package spec
CREATE OR REPLACE PACKAGE bug_test IS
  PROCEDURE test_it;
END;
/

-- TEST 1: WITHOUT user-defined exception - WORKS
CREATE OR REPLACE PACKAGE BODY bug_test IS
  gs NUMBER := 0;

  PROCEDURE pl(s VARCHAR2) IS
  BEGIN
    DBMS_OUTPUT.PUT_LINE(s);
  END;

  PROCEDURE test_it IS
  BEGIN
    RAISE EXCEPTION 'test error';
  EXCEPTION
    WHEN OTHERS THEN
      pl('err=' || SQLERRM);  -- This WORKS
  END;

BEGIN
  gs := 0;
END;
/

CALL bug_test.test_it();  -- Success: outputs "err=test error"

-- TEST 2: WITH user-defined exception - FAILS
CREATE OR REPLACE PACKAGE BODY bug_test IS
  gs NUMBER := 0;
  my_exception EXCEPTION;  -- Adding this causes the bug!

  PROCEDURE pl(s VARCHAR2) IS
  BEGIN
    DBMS_OUTPUT.PUT_LINE(s);
  END;

  PROCEDURE test_it IS
  BEGIN
    RAISE EXCEPTION 'test error';
  EXCEPTION
    WHEN OTHERS THEN
      pl('err=' || SQLERRM);  -- ERROR: exception variables cannot be used in this context
  END;

BEGIN
  gs := 0;
END;
/

CALL bug_test.test_it();  -- FAILS with error above
```

- The bug appears to be in how PL/iSQL handles exception variable scoping
- The presence of ANY user-defined `EXCEPTION` variable at package level seems to incorrectly trigger exception variable restrictions on `SQLERRM` usage within the entire package
- The user-defined exception (`my_exception`) doesn't even need to be used - merely declaring it triggers the bug
- Verified against Oracle Database 23.26 Free: both test cases work correctly in Oracle
- Workaround: Avoid declaring user-defined `EXCEPTION` variables at package level, or assign `SQLERRM` to a local variable before using it (though this workaround may not work in all cases)

rophy:
## Summary
- Fix bug where passing `SQLERRM` to procedures from exception handlers failed when the package declared user-defined `EXCEPTION` variables at package scope
- Add regression test for this fix

## Problem
When a package declares user-defined `EXCEPTION` variables (e.g., `my_exception EXCEPTION;`), calling a procedure with `SQLERRM` as an argument from an exception handler would fail with:

```
ERROR: exception variables cannot be used in this context
```

This contradicts Oracle Database behavior where both scenarios work correctly.

## Root Cause
The `is_const_datum()` function in `pl_package.c` threw an error when encountering `PLISQL_DTYPE_EXCEPTION` datums. This function is called when iterating over datums during subprocedure calls to determine which variables need to be copied between stack frames.

## Fix
Changed `is_const_datum()` to return `true` for EXCEPTION datums instead of throwing an error. This is correct because:
- Exception variables are constant identifiers (they cannot be assigned to)
- They don't need to be copied between function stack frames
- Returning `true` causes callers to skip them appropriately

## Test Plan
- [x] Verified fix with test case from issue #1151
- [x] Added regression test (`plisql_exception` test 12)
- [x] All 20 PL/iSQL oracle-check tests pass
- [x] All 234 Oracle compatibility tests pass (`make oracle-check`)

Fixes #1151

## Summary by CodeRabbit

* **Tests**
  * Added test scenario to verify SQLERRM handling with package-level exception variables and exception handlers

* **Bug Fixes**
  * Improved exception handling to properly support exception variables within package scopes

### DIFF
--- a/src/pl/plisql/src/sql/plisql_exception.sql
+++ b/src/pl/plisql/src/sql/plisql_exception.sql
@@ -315,6 +315,45 @@ END test_pragma_zero;
 
 SELECT 'PRAGMA EXCEPTION_INIT negative error code tests completed' AS result;
 
+--
+-- Test 12: SQLERRM in procedures when package has EXCEPTION variables (issue #1151)
+-- When a package declares user-defined EXCEPTION variables at package scope,
+-- passing SQLERRM to procedures from exception handlers should work.
+--
+CREATE OR REPLACE PACKAGE test_sqlerrm_with_exc IS
+  my_exception EXCEPTION;  -- User-defined exception at package scope
+  PROCEDURE log_error(p_msg VARCHAR2);
+  PROCEDURE test_sqlerrm;
+END test_sqlerrm_with_exc;
+/
+
+CREATE OR REPLACE PACKAGE BODY test_sqlerrm_with_exc IS
+  PROCEDURE log_error(p_msg VARCHAR2) IS
+  BEGIN
+    RAISE INFO 'Logged: %', p_msg;
+  END log_error;
+
+  PROCEDURE test_sqlerrm IS
+  BEGIN
+    RAISE my_exception;
+  EXCEPTION
+    WHEN my_exception THEN
+      -- This should work: passing SQLERRM to procedure
+      log_error(SQLERRM);
+  END test_sqlerrm;
+END test_sqlerrm_with_exc;
+/
+
+-- Test execution - this used to fail with "exception variables cannot be used in this context"
+BEGIN
+  test_sqlerrm_with_exc.test_sqlerrm();
+END;
+/
+
+SELECT 'SQLERRM with package EXCEPTION test passed (issue #1151)' AS result;
+
+DROP PACKAGE test_sqlerrm_with_exc;
+
 -- Cleanup
 DROP PACKAGE test_pragma_init;
 DROP PACKAGE test_pragma_raise;
--- a/src/pl/plisql/src/expected/plisql_exception.out
+++ b/src/pl/plisql/src/expected/plisql_exception.out
@@ -340,6 +340,45 @@ SELECT 'PRAGMA EXCEPTION_INIT negative error code tests completed' AS result;
  PRAGMA EXCEPTION_INIT negative error code tests completed
 (1 row)
 
+--
+-- Test 12: SQLERRM in procedures when package has EXCEPTION variables (issue #1151)
+-- When a package declares user-defined EXCEPTION variables at package scope,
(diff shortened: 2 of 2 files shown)


src/pl/plisql/src/sql/plisql_exception.sql
```diff
@@ -315,6 +315,45 @@ END test_pragma_zero;
...
+END test_sqlerrm_with_exc;
+/
+
+-- Test execution - this used to fail with "exception variables cannot be used in this context"
+BEGIN
+  test_sqlerrm_with_exc.test_sqlerrm();
+END;
+/
+
+SELECT 'SQLERRM with package EXCEPTION test passed (issue #1151)' AS result;
```
bigplaice, line 353: Comment: 
What does this line mean? Seems unnecessary.

PR #1202, opened 2026-03-04
Fix crash when nested sub-functions access NEW after DECLARE uses NEW

NotHimmel:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed incorrect handling of certain record-field datums during subprocedure compilation, preventing misinitialization and wrong OUT assignments.

* **Tests**
  * Added a comprehensive trigger-based test scenario that exercises nested subprocedure behavior with sample inserts and informational logs.


c2main (approved):
I confirm it fixed bug on our testing and the added regression tests correctly ensure the case is correctly managed.

PR #1206, opened 2026-03-10
Add multibyte search_path case conversion coverage

c2main:
Extend case_conversion.sql to cover multibyte quoted schema names    through the search_path code path. The scenario now makes the active    identifier_case_switch setting explicit, checks both interchange and    normal behavior, and includes a failure case where a quoted schema name    stops resolving once the case-conversion mode changes.
    
Also removed the alternative file output, I suppose it was added by mistake    because their diff was very small.

## Summary by CodeRabbit

* **Tests**
  * Significantly expanded test coverage for SQL identifier case conversion with comprehensive new test sequences.
  * Tests now cover schema lifecycle management, search path configuration, and switching between different case conversion modes.
  * Updated expected test outputs and consolidated test fixtures for better maintainability.
  * Improved test environment isolation with enhanced parameter reset procedures.


OreoYang:
@yuanyl630

yuanyl630 (approved):

PR #1207, opened 2026-03-10
Initialize fallback command tag for CompileFunctionStmt

c2main:
`CreateCommandTag()` can leave `tag` uninitialized for `CompileFunctionStmt` when objtype is neither FUNCTION nor PROCEDURE. Callers normally pass one of those two values, but Clang still reports the missing fallback path.

Warning excerpt:

```
utility.c:2757:13: warning: variable 'tag' is used uninitialized whenever 'if' condition is false
[-Wsometimes-uninitialized]
```

Set `tag = CMDTAG_UNKNOWN` in the remaining branch so the fallback behavior is explicit.

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensures command processing assigns a defined fallback tag for unrecognized object types, preventing ambiguous or unspecified command state and improving stability when handling non-procedure/function objects.


OreoYang:
@hs-liuxh

PR #1210, opened 2026-03-10
Keep scanner quote types self-consistent

c2main:
The Oracle scanner path mixed PsqlScanQuoteType values into helpers that are typed for Ora_psqlScanQuoteType. Clang reports those implicit enum-to-enum conversions, and the mismatch also weakens separation between PostgreSQL and Oracle scanner code paths.

Warning excerpt:

```
ora_psqlscanslash.l:254:20: warning:
implicit conversion from enumeration type PsqlScanQuoteType to different enumeration type Ora_psqlScanQuoteType
[-Wimplicit-enum-enum-cast]
```

Use Ora_psqlScanQuoteType consistently in declarations and call sites across both Oracle scanner files to keep the enum domain coherent.

## Refactor

* Updated quote type handling constants across Oracle-specific psql scanner modules for improved consistency in variable substitution and SQL identifier escaping.

### DIFF
--- a/src/bin/psql/ora_psqlscanslash.l
+++ b/src/bin/psql/ora_psqlscanslash.l
@@ -250,8 +250,8 @@ other			.
 															 yytext + 1,
 															 yyleng - 1);
 						value = cur_state->oracallbacks->get_variable(varname,
-																   PQUOTE_PLAIN,
-																   cur_state->cb_passthrough);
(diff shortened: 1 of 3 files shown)


OreoYang:
I was able to reproduce the issue now by switching the compiler from gcc to clang.

PR #1211, opened 2026-03-10
Use FLOAT8_FITS_IN_INT64 in dsinterval_mul

### LINKED ISSUE #1366: Use FLOAT8_FITS_IN_INT64 in dsinterval_mul
code improvement.

c2main:
dsinterval_mul() compared a rounded double directly to PG_INT64_MAX, which triggers Clang because that integer constant is not exactly representable as double near 2^63.

Warning excerpt:

```
src/datatype/dsinterval.c:2336:22: warning:
implicit conversion from long to double changes value from 9223372036854775807 to 9223372036854775808
[-Wimplicit-const-int-float-conversion]
```

Follow the same integer-fit pattern used in timestamp.c by checking isnan() and FLOAT8_FITS_IN_INT64() after rint(). This avoids custom bound constants and keeps overflow checks consistent with core code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved datetime/interval multiplication and division checks to detect NaN and out-of-range intermediate results, returning clearer “interval out of range” errors instead of producing invalid interval values.

* **Tests**
  * Expanded interval arithmetic test coverage with additional overflow and near-zero divisor cases, including NaN inputs, for both year-to-month and day-to-second intervals.

* **Style / Refactor**
  * Applied minor interval-handling code cleanup and parsing/formatting adjustments without changing public behavior.

close  #1366

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_datetime.out
+++ b/contrib/ivorysql_ora/expected/ora_datetime.out
@@ -1172,6 +1172,10 @@ SELECT interval'365 11:11:11' day(3) to second * 12.34;
  +000004509 20:26:24.140000000
 (1 row)
 
+-- interval day to second * huge number (overflow)
+SELECT interval'365 11:11:11' day(3) to second * 100000000000000000000;
+ERROR:  interval out of range
+
 -- interval year to month / number
 SELECT interval'10-11' year to month / 12.34;
    ?column?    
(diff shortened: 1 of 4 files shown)


c2main:
> in dsinterval_div(): result->month = (int32) (span->month / factor); result->day = (int32) (span->day / factor); Should the same checks be implemented?

Yes, addressed in the latest commit. Sorry for the delay.

PR #1214, opened 2026-03-10
Move GB18030 skip after buffer setup in test_escape

c2main:
IvorySQL keeps a GB18030-specific skip path in test_one_vector_escape(), but placing that branch before local declarations and buffer setup triggered Clang diagnostics in both test_escape modules.

Warning excerpt:

```
test_escape.c:640:14: warning:
mixing declarations and code is incompatible with standards before C99
[-Wdeclaration-after-statement]
test_escape.c:636:6: warning:
variable 'escape_err' is used uninitialized whenever 'if' condition is true
[-Wsometimes-uninitialized]
```

Allocate PQExpBuffer objects first, then apply the existing GB18030 memchr-based skip check. This preserves IvorySQL behavior while making the early-exit cleanup path safe in both modules.

## Summary by CodeRabbit

* **Tests**
  * Adjusted test execution sequence for escape functionality validation to modify allocation timing during test operations.

### DIFF
--- a/src/oracle_test/modules/test_escape/test_escape.c
+++ b/src/oracle_test/modules/test_escape/test_escape.c
@@ -633,10 +633,6 @@ test_psql_parse(pe_test_config *tc, PQExpBuffer testname,
 static void
 test_one_vector_escape(pe_test_config *tc, const pe_test_vector *tv, const pe_test_escape_func *ef)
 {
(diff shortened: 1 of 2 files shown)


NotHimmel (approved):
Clean fix for Clang warnings , no functional change. Thanks!

Issue #1292, opened 2026-04-14
Should all backend compatibility behaviors use compatible_db?


anapplebeee:
## Bug Report
Is database_mode only used to start the Oracle port, while some functions still use database_mode to determine compatibility behavior?
<img width="930" height="709" alt="Image" src="https://github.com/user-attachments/assets/66e64cdc-38ea-4fb4-81d7-255b923bb627" />

<img width="1639" height="769" alt="Image" src="https://github.com/user-attachments/assets/232eb0c6-257f-4603-8e13-cbc2c62c884a" />

### IvorySQL Version
IvorySQL_5.3

hs-liuxh:
This is a very valuable observation. The problem indeed exists. We will work on fixing these issues shortly. 
Thanks for your contribution!

hs-liuxh:
I have carefully reviewed the two files you mentioned where database_mode and compatible_db are used interchangeably.
In the parse_expr.c file, the database_mode variable indeed should be replaced with compatible_db because they are used for syntax parsing during database runtime.

However, several functions(like varstr_sortsupport) in the varlena.c file are applied to string sorting. This underlying behavior should be database-related and not change with the compatibility mode, so we have decided not to make changes here after discussion.

Additionally, I have thoroughly checked the usage of both database_mode and compatible_db variables throughout the project and supplemented other related changes. Please refer to the #1299  for details.

Thanks again for your contribution.

PR #1299, opened 2026-04-20
Fix mixing of database_mode and compatible_db variables

hs-liuxh:
#1292 

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Oracle compatibility mode detection for table function execution to properly adjust return types.
  * Enhanced identifier case transformation logic in parser and lexer for consistent Oracle mode behavior.
  * Improved DECODE/CASE expression handling under Oracle compatibility mode.
  * Refined startup parameter processing for database and user identifiers in Oracle mode.


jiaoshuntian (approved):

bigplaice (approved):
no comments.

PR #1349, opened 2026-06-04
Support pg_rewind used without wal_log_hints and data_checksums

movead123:
**引入问题**
PostgreSQL中使用pg_rewind进行脑裂后旧主跟随新主的操作，但是需要开启wal_log_hints或者data_checksums才能使用此功能，这大大降低了pg_rewind的使用范围。

**目前依赖wal_log_hints或者data_checksums的原因**
在pg_rewind功能引入时，由于需要处理分岔点之前修改数据，分岔点之后修改tuple hint的场景。因此需要开启wal_log_hints或者data_checksums来规避一些场景。如下是如果不开启这两个参数而使用pg_rewind，可能引起的数据不一致问题。

```
1.我们有一个主库 A 和一个备库 B。
2.有一个事务 xact1 正在执行，并且 xact1 修改了一些页面。
3.在 A 上执行一个检查点。
4.备库 B 被提升为主库。
5.事务 xact1 提交，并且在 A 上查询所有被 xact1 修改的数据。
6.在 A 上执行 pg_rewind。
```
这时如果执行rewind那么, 步骤5的修改不会被rewind察觉，因此不会重新拷贝这些page。这将导致rewind后的A和B的数据不一致。好了，这是社区版本要求wal_log_hints或者data_checksums的理由。现在我们有不开启这两个参数而规避此问题的方法。

**新的方案**
我们需要记录在分叉点之后提交的所有事务 ID，并在分叉点之前进一步分析 WAL 日志，找到一个起点 LSN（称为 lsn_s），该 LSN 比所有这些事务都更早。然后，我们应该从 lsn_s 读取到分叉点的 LSN，以收集这些事务所影响的页面。这样，我们就可以在 rewind 阶段将这些页面复制过来，以此达到主备一致的目的。

**使用介绍**
在pg_rewind的使用参数中添加了--deep-dig, 如果不加这个参数将走原生pg_rewind的逻辑，添加这个参数之后，将会走新的rewind逻辑。

## Summary by CodeRabbit

* **New Features**
  * Added --deep-dig (-d) option to pg_rewind to enable a deeper scan for modified pages when basic page-consistency info is missing.
  * Reworked page-map discovery so pg_rewind can perform forward then optional backward analysis when requested.

* **Bug Fixes / Reliability**
  * Improved safety checks and conditional behavior to avoid unnecessary failures; use --deep-dig to proceed with extended verification.


bigplaice:
Thanks for bringing new feature to IvorySQL!
I will review the code change and provide feedback. 

Steven

bigplaice:
Thanks for working on this — allowing `pg_rewind` to operate without `wal_log_hints` or `data_checksums` addresses a real production need, and the two-phase WAL scanning approach is conceptually interesting. However, there are several issues that need to be addressed before this can be merged.

---

**Correctness Issues (must fix)**

**1. ABORT records corrupt `min_commited_xid` (critical)**

In `track_rewind_snapshot()`, for `XLOG_XACT_ABORT` / `XLOG_XACT_ABORT_PREPARED` records, neither the `XLOG_XACT_COMMIT` nor `XLOG_XACT_COMMIT_PREPARED` branch is taken, leaving `current_xid` as `InvalidTransactionId (0)`. There is no early return before the min/max update block, so `RewindTransactionIdPrecedes(0, any_normal_xid)` returns `true` and overwrites `min_commited_xid` with 0. This causes `check_rewind_safety()` to return `true` (hitting the `InvalidTransactionId` branch), allowing rewind to proceed silently when it should not — risking data corruption. This will trigger on any WAL that contains rolled-back transactions, which is virtually always the case.

**2. Backward scan has no guaranteed termination**

The loop in `extractPageMapBackward()` terminates only when `base_xid <= min_commited_xid`. Two failure modes exist: (a) if no `XLOG_RUNNING_XACTS` record is found, `base_xid` stays `InvalidTransactionId (0)`, and `0 <= any_normal_xid` is always true, so the loop exits immediately after the first record without scanning anything useful; (b) if the condition is never satisfied, the loop walks all the way back via `xl_prev` until it hits `xl_prev == 0`, at which point `XLogBeginRead(xlogreader, 0)` fails with a fatal error. A secondary termination condition based on checkpoint records (checking `CheckPoint.nextXid <= min_commited_xid`) and a hard stop for `xl_prev == InvalidXLogRecPtr` are both needed.

**3. XID comparisons are not wraparound-safe**

`if (base_xid <= min_commited_xid)` in both `extractPageMapBackward()` and `check_rewind_safety()` uses direct `<=` on 32-bit `TransactionId` values. This gives wrong results near XID wraparound. All such comparisons should use `RewindTransactionIdPrecedes()`, which is already defined in this patch.

**4. `check_rewind_safety()` lacks documentation for its correctness argument**

The function's comments just say "it's safe" without explaining why. The actual argument depends on PostgreSQL's full-page write (FPW) mechanism: if all post-fork committed transactions started *after* the last checkpoint, their page modifications all carry FPW records already captured by the forward scan. This reasoning should be documented clearly so reviewers can verify it. Additionally, the function needs a guard for `base_xid == InvalidTransactionId` to avoid a premature "safe" verdict when no `XLOG_RUNNING_XACTS` record has been seen yet.

---

**Design Issues**

**5. Removing the sanity check unconditionally changes default behavior**

The existing check in `sanityChecks()` was removed entirely rather than gated behind `--deep-dig`. This means users without checksums/wal_log_hints will now run a full forward scan before getting a fatal error — wasting I/O. Worse, if `check_rewind_safety()` incorrectly returns `true` (due to bug #1), the rewind proceeds silently with potentially missing page copies. The original check should be preserved and bypassed only when `--deep-dig` is explicitly specified, with a prominent warning.

**6. XID range filter `[min, max]` is too coarse**

`check_transaction_need_rewind()` copies pages for all transactions whose XID falls in `[min_commited_xid, max_commited_xid]`, but this range includes aborted transactions whose page modifications should not be copied. The code comment acknowledges this but offers no path forward. A bitmap or list of actually-committed XIDs would be more accurate; `palloc` is available in frontend code.

**7. Global XID trackers are never reset**

`min_commited_xid`, `max_commited_xid`, and `base_xid` are static globals with no explicit reset at function entry. This is harmless for a single-use tool today, but will cause state pollution if `pg_rewind` is ever embedded in a library or run multiple times in a test harness. They should be initialised at the start of `extractPageMapForward()`.

**8. New backend dependencies increase maintenance coupling**

The patch pulls in `xactdesc.c` and `standbydesc.c` from the backend via symlinks. These files carry heavy backend infrastructure (`StringInfo`, `appendStringInfo`, etc.) even though only `ParseCommitRecord` and a few struct definitions are actually needed. If upstream PostgreSQL changes the WAL record format or rmgrdesc interfaces, `pg_rewind` will need to track those changes. Consider extracting only the needed declarations instead of linking the full descriptor files.

---

**Test Coverage**

The patch adds no regression tests. Given that this bypasses a long-standing safety restriction, tests are essential before merging. At minimum, please add:

- TAP end-to-end tests for rewind without checksums/`wal_log_hints`
- TAP tests covering the `--deep-dig` path
- Tests for `check_rewind_safety()` returning false (user-facing error path)
- Tests with mixed commit/abort WAL to verify XID tracking correctness
- XID wraparound boundary tests

movead123:
感谢指出代码中的诸多问题，已经根据review意见做出了修改，但是对第6点保持了原来的设计。
当我们在设计上达成一致之后，再去写用例。

evanchaoli:
感谢提交。我看了这个提交，并且把你之前在pg社区提交的讨论也仔细看了一遍，在进入到代码层面之前，有几个问题我想先理解一下：

1. 感觉上，这个功能主要的目的是为了少写WAL日志。请解释一下实际的使用场景以及这是否是一个在生产环境中遇到过的痛点问题？如果是，可否给一个真实案例。
2. 这个提交的理论基础是通过回溯足够的WAL日志以准确判断两边的block是否一致。请具体阐述和证明一下这个理论的正确性。
3. 请问如果在极端情况下找不到足够老的WAL的情况下，怎么处理？此时因为那两个开关都没打开，老的方式也无法使用了。
4. 在PG19里，有几个新功能可能影响hint bits，比如b46e1e54d，可能在query的时候修改hint bits，请问是否有评估过对这个提交有没有影响？

Issue #1310, opened 2026-04-21
Ecosystem Integration: IvorySQL + pg_profile


hs-liuxh:
## Feature Request
Ecosystem Integration: IvorySQL + pg_profile
**Describe the feature you'd like to propose:**
Implement pg_profile to enable performance profiling and statistical monitoring for bottleneck identification.

PR #1364, opened 2026-06-22
Add a PG-dialect allow list

hs-liuxh:
Add a pg_dialect configuration to the PG extension's .control file to specify whether the PostgreSQL parser should be used for extension installation and function invocation under Oracle compatibility mode. 

Many PG extensions rely on native PostgreSQL syntax, particularly for time and interval data types, which cannot be parsed by the Oracle parser. This commit provides a unified solution to this issue.

#1310

## Summary by CodeRabbit

* **Bug Fixes**
  * Extension scripts that require PostgreSQL compatibility now run with the correct SQL dialect automatically.
  * Functions created by those extension scripts are pinned to the same compatibility mode to avoid unexpected behavior.
  * If a function definition already sets the compatibility mode, that setting is preserved and not overwritten.


jiaoshuntian (approved):

src/backend/commands/extension.c
```diff
@@ -1367,6 +1420,14 @@ execute_extension_script(Oid extensionOid, ExtensionControlFile *control,
 	 */
 	creating_extension = true;
 	CurrentExtensionObject = extensionOid;
+	/*
+	 * IvorySQL: flag that the PG dialect is being forced for this script so
+	 * that CreateFunction can pin functions created by the script to the PG
+	 * dialect (see functioncmds.c).
+	 */
+	if (DB_ORACLE == database_mode &&
```
bigplaice, line 1428: can be combined with line 1385.

bigplaice (approved):

PR #1372, opened 2026-06-30
fix: ALTER ... SET ... TO DEFAULT 可绕过新限制

enkilee:
fix issue #1374 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for role- and database-level setting changes so unsupported options can’t be saved with `ALTER ... SET` or `RESET`.
  * Updated error messaging to clearly reference both `SET` and `RESET`.
* **Tests**
  * Added regression coverage ensuring `ivorysql.compatible_mode` cannot be persisted or reset via role-level `ALTER ... SET/RESET`.


jiaoshuntian:
@bigplaice will follow up.

bigplaice:
Thanks for the patch! Basically looks good to me. 
Please solve the regression test failure. @enkilee

enkilee:
> Thanks for the patch! Basically looks good to me. Please solve the regression test failure. @enkilee

@bigplaice  already done.

bigplaice (approved):
no more comments.

PR #1381, opened 2026-07-07
Fix package datum misresolution in plisql_param_compile

### LINKED ISSUE #1382: Package CONSTANT/variable referenced more than once in an expression returns wrong value
## Bug Report
A PL/iSQL package-qualified CONSTANT (or variable) can silently return the wrong value — either empty, or the leftover value of a completely unrelated local variable — when it is referenced more than once while a "simple expression" is being evaluated, particularly when the reference is on the right-hand side of an assignment to a variable that was already assigned earlier in the same block.

This is a correctness bug in the core PL/iSQL engine (src/pl/plisql/src/pl_exec.c), not specific to any one package — it reproduces with any freshly-created package that exposes a VARCHAR2 constant.

Reported-by: Yasir Hussain Shah <yasir.hussain.shah@data-bene.io>

### IvorySQL Version
IvorySQL master branch

### Current Behavior
NOTICE:  Line 1
Line 2
NOTICE:  Line 3Line 1
Line 2Line 4

### Expected behavior/code
NOTICE:  Line 1
Line 2
NOTICE:  Line 3
Line 4

### Step to reproduce

CREATE OR REPLACE PACKAGE my_test_pkg AS
    my_const CONSTANT VARCHAR2(2) := chr(13) || chr(10);
END my_test_pkg;
/

DO $$
DECLARE
    l_message VARCHAR2(100);
    v_crlf    VARCHAR2(2) := my_test_pkg.my_const;
BEGIN
   l_message := 'Line 1' || v_crlf || 'Line 2';
   RAISE NOTICE '%', l_message;
   l_message := 'Line 3' || my_test_pkg.my_const || 'Line 4';
   RAISE NOTICE '%', l_message;
END;
$$;

db-nikhil:
Fixes #1382

plisql_param_compile rebinds dno to the resolved datum's own dno, which for a package-qualified reference lives in the package's datum namespace rather than the calling function's. That dno was then compared against expr->target_param (a dno in the calling function's namespace to decide whether to install the plisql_param_eval_var_check fast path. Unlike its siblings plisql_param_eval_var/_ro, that path indexes estate->datums[dno] directly with no pkgoid awareness, so a coincidental numeric match between the two namespaces (common for small packages/blocks, e.g. both being dno 0) made it read and return an unrelated local variable's value instead of the package datum.

Symptom: referencing a package CONSTANT/variable more than once while evaluating a simple expression could silently return the wrong value
- either empty (uninitialized target) or the stale value of whatever variable happened to share that dno in the calling block.

Skip that fast path for package-qualified datums so they fall through to plisql_param_eval_var_ro, which already resolves package datums correctly via get_package_datum_bydno(). Adds a regression case to ora_package.sql reproducing the exact failure pattern.

Reported-by: Yasir Hussain Shah <yasir.hussain.shah@data-bene.io>

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved inconsistent behavior when repeatedly evaluating package constants in expressions, especially for CRLF/newline-containing strings.
  * Improved evaluation for simple expressions involving package-qualified references to ensure correct variable handling.

* **Tests**
  * Added a new regression test that exercises repeated expression building using both a package constant and a mutable package variable, including validation of CRLF/newline results and expected pass/fail checks.

### DIFF
--- a/src/oracle_test/regress/expected/ora_package.out
+++ b/src/oracle_test/regress/expected/ora_package.out
@@ -6321,6 +6321,61 @@ SELECT * FROM test_nested_commit ORDER BY id;
 
 DROP PACKAGE pkg_commit_test;
 DROP TABLE test_nested_commit;
+-- Test that a package CONSTANT referenced more than once while evaluating a
+-- simple expression always yields its own value, even when the reference is
+-- the RHS of an assignment to a variable that was already assigned earlier
+-- in the same block.  (Previously, plisql_param_compile rebound dno to the
+-- resolved datum's own dno within the package's namespace, and
+-- plisql_param_eval_var_check compared that against expr->target_param -- a
+-- dno in the *calling* function's namespace -- without accounting for
+-- pkgoid.  A coincidental numeric match caused it to read an unrelated local
+-- variable's value instead of the package constant.)
+CREATE OR REPLACE PACKAGE pkg_const_repeat_test IS
+  my_const CONSTANT VARCHAR2(2) := chr(13) || chr(10);
+  my_var VARCHAR2(20) := 'pkgvar-init';
+END pkg_const_repeat_test;
+/
+DO $$
+DECLARE
+  l_message VARCHAR2(100);
+  v_local   VARCHAR2(2) := pkg_const_repeat_test.my_const;
+BEGIN
+  IF v_local != chr(13) || chr(10) THEN
+    RAISE EXCEPTION 'unexpected v_local value';
+  END IF;
+
+  l_message := 'Line 1' || v_local || 'Line 2';
+  IF l_message != 'Line 1' || chr(13) || chr(10) || 'Line 2' THEN
+    RAISE EXCEPTION 'unexpected l_message value (first assignment): %', l_message;
+  END IF;
+
+  l_message := 'Line 3' || pkg_const_repeat_test.my_const || 'Line 4';
+  IF l_message != 'Line 3' || chr(13) || chr(10) || 'Line 4' THEN
+    RAISE EXCEPTION 'unexpected l_message value (second assignment): %', l_message;
+  END IF;
+
+  -- Same failure pattern, but for a mutable (non-CONSTANT) package variable,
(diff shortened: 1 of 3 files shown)


bigplaice:
Thanks for your patch! We'll review it soon.

db-nikhil:
> 🧹 Nitpick comments (2)
> > src/pl/plisql/src/pl_exec.c (1)> `7101-7118`: _📐 Maintainability & Code Quality_ | _🔵 Trivial_ | _💤 Low value_
> > > **Fix looks correct.**
> > > 🧹 Optional cleanup
> > > ```diff
> > > -		if (isvarlena && dno == expr->target_param && expr->expr_simple_expr &&
> > > -			!OidIsValid(((PLiSQL_var *) datum)->pkgoid))
> > > +		if (isvarlena && dno == expr->target_param && expr->expr_simple_expr &&
> > > +			!OidIsValid(datum->pkgoid))
> > > ```
> > > 
> > > In `@src/pl/plisql/src/pl_exec.c` around lines 7101 - 7118, The fast-path guard in
> > > the PL/pgSQL parameter setup is correct, but the new `pkgoid` check is using an
> > > unnecessary cast to `PLiSQL_var` even though `pkgoid` is available on
> > > `PLiSQL_datum`. Update the condition in this `plisql_param_eval_var_check`
> > > selection logic to read `pkgoid` directly from `datum`, keeping the existing
> > > `dno`, `expr->target_param`, and `expr->expr_simple_expr` checks unchanged.
> > > ```
> > src/oracle_test/regress/sql/ora_package.sql (1)> `6133-6171`: _🎯 Functional Correctness_ | _🔵 Trivial_ | _⚡ Quick win_
> > > **Consider covering the non-constant variable case too.**
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against current code. Fix only still-valid issues, skip the
> rest with a brief reason, keep changes minimal, and validate.
> 

Nitpicks have been addressed in the latest commit

bigplaice (approved):
This is a nice fix. Suggest to merge it.

PR #1392, opened 2026-07-12
Fix json gb18030 encoding

zhaohai666:
This patch fixes misparsing issues of JSON/JSONB text under GB18030 and similar multi-byte encodings.
The root cause is that ASCII special bytes like backslash can exist as trailing bytes of multi-byte characters, while the original lexer scans byte by byte without encoding awareness.
We adjust json_lex_string and incremental lexing logic to consume complete multibyte characters via libpg encoding utilities for non-UTF8 encodings. Optimized fast paths for UTF-8 are untouched. Comprehensive regression tests are added in gb18030_2022 extension.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved JSON and JSONB parsing for non-UTF-8 multibyte encodings, including GB18030.
  - Prevented valid multibyte characters from being misinterpreted as quotation marks or backslashes.
  - Improved handling of incremental JSON parsing and escaped characters.

- **Tests**
  - Added comprehensive regression coverage for GB18030-2022 JSON and JSONB behavior, including nested structures, Unicode characters, escapes, and JSON functions.

### DIFF
--- a/contrib/gb18030_2022/Makefile
+++ b/contrib/gb18030_2022/Makefile
@@ -10,7 +10,8 @@ EXTENSION = gb18030_2022
 PGFILEDESC = "gb18030_2022,support gb18030 2022 with extension"
 DATA = gb18030_2022--1.0.sql
 REGRESS = gb18030_2022_and_utf8 \
-		copy
+		copy \
+		json_gb18030
 
 ifdef USE_PGXS
 PG_CONFIG = pg_config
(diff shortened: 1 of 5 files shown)


NotHimmel:
Hi @zhaohai666 , thank you for this PR and for digging into the GB18030 parsing issue — the problem analysis and the broad regression coverage you added were both really nice to see.

There's a parallel PR (#1393) for the same issue. That contributor had explicitly claimed the issue , which is the workflow we encourage to avoid duplicated effort. 

On top of that, after reviewing both approaches side by side, we felt theirs was the better fit: it scopes the change tightly to the encodings actually affected (GB18030), so UTF-8 and the other server encodings keep their existing fast path untouched. For those reasons we're going to move forward with #1393, and I'll be closing this one.

One tip for next time: dropping a quick comment on an issue to claim it before you start helps everyone avoid landing on the same problem in parallel.

Thanks again!

PR #1393, opened 2026-07-12
Fix JSON/JSONB parsing for GB18030

### LINKED ISSUE #1340: Convert UTF8 failed
IvorySQL Version

  IvorySQL 5.3 (PostgreSQL 18.3)

  OS Version (uname -a)

  Linux yansbook 6.12.90+deb13-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.12.90-1 (2026-05-22) x86_64 GNU/Linux

  Configuration options

  - server_encoding = GB18030
  - client_encoding = UTF8
  - Database initialized with initdb using GB18030 locale/encoding

  Current Behavior

  When server_encoding = GB18030 and client_encoding = UTF8, inserting JSON data containing non-ASCII characters (Chinese, Japanese, Korean, etc.) into a JSONB column causes the PostgreSQL JSON parser to report an
  error:

  ERROR:  invalid input syntax for type json
  LINE 3: ('1', '{"zh": "加拿大,美国", "en": "Canada,United States of ...
  DETAIL:  Escape sequence "\国" is invalid.
  CONTEXT:  JSON data, line 1: ...States of America", "ja": "カナダ、アメリカ合衆国...

  There are no actual backslash characters in the JSON string. However, the JSON parser incorrectly interprets certain bytes within UTF-8 multi-byte sequences as \ (0x5C), treating subsequent bytes as an invalid escape
  sequence.

  Reproduction SQL:

  SET client_encoding = 'UTF8';

  -- Fails: JSON with non-ASCII characters
  SELECT '{"ja": "カナダ、アメリカ合衆国"}'::jsonb;

  -- Succeeds: equivalent JSON with \uXXXX escapes
  SELECT '{"ja": "\u30ab\u30ca\u30c0\u3001\u30a2\u30e1\u30ea\u30ab\u5408\u8846\u56fd"}'::jsonb;

  Expected behavior

  When client_encoding = UTF8, the JSON parser should correctly handle UTF-8 encoded JSON text. It should not misinterpret UTF-8 multi-byte sequences as escape characters. PostgreSQL should either correctly perform the
  client_encoding → server_encoding conversion before JSON parsing, or the JSON parser should interpret the byte stream based on client_encoding.

  Step to reproduce

  1. Initialize the database with GB18030 encoding: initdb --encoding=GB18030
  2. Set client_encoding = 'UTF8'
  3. Execute: SELECT '{"ja": "カナダ、アメリカ合衆国"}'::jsonb;
  4. Observe the error: invalid input syntax for type json

  Additional context

  - Impact: All scenarios involving JSON/JSONB operations via a UTF-8 client under GB18030 server_encoding.
  - Workaround: Pre-escaping non-ASCII characters in JSON as \uXXXX sequences (i.e., ensure_ascii=True) avoids this issue.
  - Suspected root cause: The JSON parser interprets the byte stream using server_encoding (GB18030) rather than client_encoding (UTF8) when processing JSON text. In GB18030, certain multi-byte sequences may contain the
   0x5C byte, which is misidentified as the JSON escape initiator \.
  - Vanilla PostgreSQL behavior: Standard PostgreSQL defaults to UTF-8 database initialization, so this issue is typically not encountered. IvorySQL 5.3 may default to GB18030 in Chinese environments, thereby exposing
  this encoding conversion defect.

btlqql:
## Summary

- resolve #1340 by scanning complete GB18030 characters while lexing JSON strings
- prevent a GB18030 trailing byte in the ASCII range from being interpreted as JSON syntax
- reject truncated GB18030 input without reading past the supplied buffer
- keep the behavior change scoped to the server-supported GB18030 encoding

## Verification

- rebased onto the current `master`
- built `src/test/modules/test_escape/test_escape` on `WZU_Server`
- ran all 1,029 `test_escape` cases against a SQL_ASCII database: 0 failures

Resolves #1340.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON string parsing for GB18030-encoded text containing multibyte characters.
  * Prevented valid continuation bytes from being misinterpreted as JSON syntax.
  * Improved handling and error reporting for truncated multibyte characters.

* **Tests**
  * Added regression coverage for GB18030 escape handling and incomplete input.
  * Added validation to ensure reported error positions are accurate.

### DIFF
--- a/src/common/jsonapi.c
+++ b/src/common/jsonapi.c
@@ -2214,26 +2214,67 @@ json_lex_string(JsonLexContext *lex)
 			/*
 			 * Skip to the first byte that requires special handling, so we
 			 * can batch calls to jsonapi_appendBinaryStringInfo.
+			 *
+			 * A GB18030 trailing byte can fall in the 7-bit ASCII range.  A
+			 * byte that continues a multibyte character must not be mistaken
+			 * for a backslash, quote, or control character, so advance by
+			 * whole characters for that encoding.
 			 */
-			while (p < end - sizeof(Vector8) &&
-				   !pg_lfind8('\\', (const uint8 *) p, sizeof(Vector8)) &&
-				   !pg_lfind8('"', (const uint8 *) p, sizeof(Vector8)) &&
-				   !pg_lfind8_le(31, (const uint8 *) p, sizeof(Vector8)))
-				p += sizeof(Vector8);
-
-			for (; p < end; p++)
+			if (lex->input_encoding == PG_GB18030)
 			{
-				if (*p == '\\' || *p == '"')
-					break;
-				else if ((unsigned char) *p <= 31)
+				while (p < end)
 				{
-					/* Per RFC4627, these characters MUST be escaped. */
-					/*
-					 * Since *p isn't printable, exclude it from the context
-					 * string
-					 */
-					lex->token_terminator = p;
-					return JSON_ESCAPING_REQUIRED;
+					if (*p == '\\' || *p == '"')
+						break;
+					else if ((unsigned char) *p <= 31)
+					{
+						/* Per RFC4627, these characters MUST be escaped. */
+						/*
+						 * Since *p isn't printable, exclude it from the
+						 * context string
+						 */
+						lex->token_terminator = p;
+						return JSON_ESCAPING_REQUIRED;
+					}
+					else if (IS_HIGHBIT_SET((unsigned char) *p))
+					{
+						ptrdiff_t	remaining = end - p;
+						int			mblen;
+
+						mblen = pg_encoding_mblen_or_incomplete(lex->input_encoding,
+																p, remaining);
+						if (mblen > remaining)
+						{
+							s = p;
+							FAIL_OR_INCOMPLETE_AT_CHAR_START(JSON_INVALID_TOKEN);
+						}
+						p += mblen - 1;
+					}
+					p++;
+				}
+			}
+			else
+			{
+				while (p < end - sizeof(Vector8) &&
+					   !pg_lfind8('\\', (const uint8 *) p, sizeof(Vector8)) &&
+					   !pg_lfind8('"', (const uint8 *) p, sizeof(Vector8)) &&
+					   !pg_lfind8_le(31, (const uint8 *) p, sizeof(Vector8)))
+					p += sizeof(Vector8);
+
+				for (; p < end; p++)
+				{
+					if (*p == '\\' || *p == '"')
+						break;
+					else if ((unsigned char) *p <= 31)
+					{
+						/* Per RFC4627, these characters MUST be escaped. */
+						/*
+						 * Since *p isn't printable, exclude it from the
+						 * context string
+						 */
+						lex->token_terminator = p;
+						return JSON_ESCAPING_REQUIRED;
+					}
 				}
 			}
 
--- a/src/test/modules/test_escape/test_escape.c
+++ b/src/test/modules/test_escape/test_escape.c
@@ -246,6 +246,101 @@ test_gb18030_json(pe_test_config *tc)
 	destroyPQExpBuffer(raw_buf);
 }
 
+/*
+ * Exercise JSON strings containing complete and truncated GB18030 multibyte
+ * characters.  A complete character whose trailing byte is 0x5C ('\') must
+ * not be mistaken for a backslash, while a truncated character must be
+ * rejected at the correct input position.
+ *
+ * The bytes below encode a JSON object whose value includes U+8846 in
+ * GB18030.  That code point is encoded as 0xD0 0x5C, so its second byte is
+ * the ASCII backslash.  Previously this raised "invalid input syntax for
+ * type json: Escape sequence ... is invalid".
+ */
+static void
+test_gb18030_json_multibyte(pe_test_config *tc)
+{
+	static const unsigned char valid_input[] = {
+		0x7b, 0x22, 0x6a, 0x61, 0x22, 0x3a, 0x22,	/* {"ja":" */
+		0xa5, 0xab, 0xa5, 0xca, 0xa5, 0xc0, 0xa1, 0xa2, /* U+30AB etc. */
+		0xa5, 0xa2, 0xa5, 0xe1, 0xa5, 0xea, 0xa5, 0xab, /* U+30A2 etc. */
+		0xba, 0xcf,				/* U+5408 */
+		0xd0, 0x5c,				/* U+8846; second byte is '\' */
+		0xb9, 0xfa,				/* U+56FD */
+		0x22, 0x7d				/* "} */
+	};
+	static const unsigned char truncated_input[] = {
+		0x7b, 0x22, 0x6a, 0x61, 0x22, 0x3a, 0x22,	/* {"ja":" */
+		0x41, 0x42, 0x43, 0x44, 0x45, 0x46, 0x47,	/* ABCDEFG */
+		0xd0,					/* truncated GB18030 lead byte */
+	};
+	size_t		input_len = sizeof(valid_input);
+	PQExpBuffer raw_buf;
+	JsonLexContext *lex;
+	JsonSemAction sem = {0};	/* no callbacks */
+	JsonParseErrorType json_error;
+	bool		ok;
+
+	/* Catch any attempt to read beyond the supplied JSON text. */
+	raw_buf = createPQExpBuffer();
+	appendBinaryPQExpBuffer(raw_buf, (const char *) valid_input, input_len);
+	appendPQExpBufferStr(raw_buf, NEVER_ACCESS_STR);
+	VALGRIND_MAKE_MEM_NOACCESS(&raw_buf->data[input_len],
+							   raw_buf->len - input_len);
+
+	/* need_escapes=true exercises the jsonb input path */
+	lex = makeJsonLexContextCstringLen(NULL, raw_buf->data, input_len,
+									   PG_GB18030, true);
+	json_error = pg_parse_json(lex, &sem);
+	report_result(tc, json_error == JSON_SUCCESS,
+				  "GB18030 multibyte backslash - pg_parse_json",
+				  "", "need_escapes=true",
+				  json_error == JSON_SUCCESS ?
+				  "accepted" : json_errdetail(json_error, lex));
+	freeJsonLexContext(lex);
+
+	/* need_escapes=false exercises the validation-only path */
+	lex = makeJsonLexContextCstringLen(NULL, raw_buf->data, input_len,
+									   PG_GB18030, false);
+	json_error = pg_parse_json(lex, &sem);
+	report_result(tc, json_error == JSON_SUCCESS,
+				  "GB18030 multibyte backslash - pg_parse_json",
+				  "", "need_escapes=false",
+				  json_error == JSON_SUCCESS ?
+				  "accepted" : json_errdetail(json_error, lex));
+	freeJsonLexContext(lex);
+	destroyPQExpBuffer(raw_buf);
+
+	/*
+	 * The truncated input advances beyond an ASCII run before ending at the
+	 * lead byte, so the reported position must not use the stale run start.
+	 */
+	input_len = sizeof(truncated_input);
+	raw_buf = createPQExpBuffer();
+	appendBinaryPQExpBuffer(raw_buf, (const char *) truncated_input, input_len);
+	appendPQExpBufferStr(raw_buf, NEVER_ACCESS_STR);
+	VALGRIND_MAKE_MEM_NOACCESS(&raw_buf->data[input_len],
+							   raw_buf->len - input_len);
+
+	lex = makeJsonLexContextCstringLen(NULL, raw_buf->data, input_len,
+									   PG_GB18030, false);
+	json_error = pg_parse_json(lex, &sem);
+
+	/*
+	 * The input must be rejected as an invalid token, and the reported token
+	 * must end at the truncated lead byte rather than at the ASCII run start.
+	 */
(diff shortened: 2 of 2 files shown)


NotHimmel:
Hi, @btlqql.  Welcome to contribute to the IvorySQL community! The regression checks has started and we will review the code soon.

btlqql:
@NotHimmel Thanks for the feedback. I've read it and am reviewing the requested changes now. I'll follow up here with the implementation and verification results.

btlqql:
@NotHimmel Thanks for the review. I addressed the point about the error context for truncated GB18030 / client-only multibyte input.

In `json_lex_string()`'s multibyte scan loop, the incomplete-character branch now advances the outer cursor to the scan position before reporting, so `lex->token_terminator` (and the message produced by `json_errdetail()`) points at the truncated character instead of the start of the preceding ASCII run. Previously a long run of plain ASCII ahead of a truncated character left the terminator pointing well before the actual truncated byte, even though parsing still rejected the input correctly.

I also added `test_gb18030_json_truncated_multibyte`, which feeds a GB18030 JSON string truncated mid-character at end of input and checks both that it is rejected (`JSON_INVALID_TOKEN`) and that the terminator lands on the truncated lead byte.

Verified locally: the existing well-formed GB18030 backslash case still parses in both `need_escapes` modes, the truncated case now reports the error at the correct position, and `pgindent --check` is clean on the changed files.

src/common/jsonapi.c
```diff
@@ -2165,26 +2165,85 @@ json_lex_string(JsonLexContext *lex)
...
+			 * so advance by whole characters instead of using the
+			 * byte-oriented fast path.
 			 */
-			while (p < end - sizeof(Vector8) &&
-				   !pg_lfind8('\\', (uint8 *) p, sizeof(Vector8)) &&
-				   !pg_lfind8('"', (uint8 *) p, sizeof(Vector8)) &&
-				   !pg_lfind8_le(31, (uint8 *) p, sizeof(Vector8)))
-				p += sizeof(Vector8);
+			if (lex->input_encoding == PG_GB18030 ||
+				PG_ENCODING_IS_CLIENT_ONLY(lex->input_encoding))
```
NotHimmel: `PG_ENCODING_IS_CLIENT_ONLY()` here is broader than #1340, in the backend `input_encoding` is never client-only. What's the intent of adding the client-only arm here?

src/test/modules/test_escape/test_escape.c
NotHimmel, line 238: example here

src/test/modules/test_escape/test_escape.c
```diff
@@ -246,6 +246,100 @@ test_gb18030_json(pe_test_config *tc)
...
+		0x22, 0x7d				/* "} */
+	};
+	size_t		input_len = sizeof(input);
+	JsonLexContext *lex;
+	JsonSemAction sem = {0};	/* no callbacks */
+	JsonParseErrorType json_error;
+
+	/* need_escapes=true exercises the jsonb input path */
+	lex = makeJsonLexContextCstringLen(NULL, (const char *) input, input_len,
+									   PG_GB18030, true);
```
NotHimmel: The existing `test_gb18030_json()` in this same file deliberately poisons memory
past the input so an over-read is caught under Valgrind.
Please follow the existing `test_gb18030_json()` in this same file as the
template — it uses `raw_buf` + `NEVER_ACCESS_STR` + `VALGRIND_MAKE_MEM_NOACCESS`

src/test/modules/test_escape/test_escape.c
```diff
@@ -246,6 +246,100 @@ test_gb18030_json(pe_test_config *tc)
...
+	report_result(tc, json_error == JSON_SUCCESS,
+				  "GB18030 multibyte backslash - pg_parse_json",
+				  "", "need_escapes=true",
+				  json_error == JSON_SUCCESS ?
+				  "accepted" : json_errdetail(json_error, lex));
+	freeJsonLexContext(lex);
+
+	/* need_escapes=false exercises the validation-only path */
+	lex = makeJsonLexContextCstringLen(NULL, (const char *) input, input_len,
+									   PG_GB18030, false);
```
NotHimmel: The existing `test_gb18030_json()` in this same file deliberately poisons memory
past the input so an over-read is caught under Valgrind.
Please follow the existing `test_gb18030_json()` in this same file as the
template — it uses `raw_buf` + `NEVER_ACCESS_STR` + `VALGRIND_MAKE_MEM_NOACCESS`

src/test/modules/test_escape/test_escape.c
```diff
@@ -246,6 +246,100 @@ test_gb18030_json(pe_test_config *tc)
...
+	};
+	const unsigned char *const truncated = input + sizeof(input) - 1;
+	size_t		input_len = sizeof(input);
+	JsonLexContext *lex;
+	JsonSemAction sem = {0};	/* no callbacks */
+	JsonParseErrorType json_error;
+	bool		ok;
+
+	lex = makeJsonLexContextCstringLen(NULL, (const char *) input, input_len,
+									   PG_GB18030, false);
```
NotHimmel: The existing `test_gb18030_json()` in this same file deliberately poisons memory
past the input so an over-read is caught under Valgrind.
Please follow the existing `test_gb18030_json()` in this same file as the
template — it uses `raw_buf` + `NEVER_ACCESS_STR` + `VALGRIND_MAKE_MEM_NOACCESS`

src/test/modules/test_escape/test_escape.c
```diff
@@ -246,6 +246,100 @@ test_gb18030_json(pe_test_config *tc)
 	destroyPQExpBuffer(raw_buf);
 }
 
+/*
+ * Confirm that a valid GB18030 JSON document is accepted even when a
+ * multibyte character's trailing byte is 0x5C ('\').  The JSON lexer must
+ * consume whole characters and not mistake such a byte for a backslash.
+ *
+ * The bytes below spell {"ja":"カナダ、アメリカ合衆国"} in GB18030.  Notably,
+ * the character 衆 (U+8846) is encoded as 0xD0 0x5C, so its second byte is
```
NotHimmel: Suggest using code points (e.g. "U+8846") in the comments and keeping the hex arrays as-is.

src/test/modules/test_escape/test_escape.c
```diff
@@ -246,6 +246,100 @@ test_gb18030_json(pe_test_config *tc)
 	destroyPQExpBuffer(raw_buf);
 }
 
+/*
+ * Confirm that a valid GB18030 JSON document is accepted even when a
+ * multibyte character's trailing byte is 0x5C ('\').  The JSON lexer must
+ * consume whole characters and not mistake such a byte for a backslash.
+ *
+ * The bytes below spell {"ja":"カナダ、アメリカ合衆国"} in GB18030.  Notably,
```
NotHimmel: It's best not to put non-ASCII (Japanese) characters into the C source.

NotHimmel (changes requested):

NotHimmel:
@btlqql Thanks for the contribution, I left a few more comments just now. One worth settling before merge: the `PG_ENCODING_IS_CLIENT_ONLY()` arm goes beyond #1340 (which is GB18030-as-server only)  — could you explain why it's needed here?

Otherwise in good shape; I verified it resolves #1340 on my end.

btlqql:
Thanks for the review. The current head now keeps this change strictly scoped to `PG_GB18030`; the broader `PG_ENCODING_IS_CLIENT_ONLY()` arm has been removed. The lexer therefore uses whole-character scanning only for the GB18030 server-encoding case from #1340, while all other encodings retain the existing vectorized path.

The truncated-character diagnostic and regression coverage are also included, and the current build, Meson, contrib, PostgreSQL, Oracle, and policy checks pass. Could you please re-review when convenient?

src/test/modules/test_escape/test_escape.c
```diff
@@ -958,6 +1070,8 @@ main(int argc, char *argv[])
 
 	test_gb18030_page_multiple(&tc);
 	test_gb18030_json(&tc);
+	test_gb18030_json_multibyte_backslash(&tc);
+	test_gb18030_json_truncated_multibyte(&tc);
```
NotHimmel: The two new functions can be folded into one

NotHimmel (changes requested):

src/test/modules/test_escape/test_escape.c
btlqql: Addressed in 91a2f91ac9. I folded the valid multibyte-backslash and truncated-character cases into a single `test_gb18030_json_multibyte()` function, while preserving both `need_escapes` paths, the truncated token-position assertion, and the Valgrind no-access guards.

btlqql:
Point-by-point status of the review on `ee49206280c489fd55f52ab83d017482ed2b85c5`, so this is not left on a `CHANGES_REQUESTED` that no longer reflects the code.

**`src/common/jsonapi.c` — "`PG_ENCODING_IS_CLIENT_ONLY()` here is broader than #1340, in the backend `input_encoding` is never client-only. What's the intent?"**

The arm is gone. The fast path is now selected by `lex->input_encoding == PG_GB18030` alone, and the original `pg_lfind8`/`pg_lfind8_le(31, ...)` loop is preserved verbatim in the `else` branch:

```c
if (lex->input_encoding == PG_GB18030)
{
    ...
    mblen = pg_encoding_mblen_or_incomplete(lex->input_encoding, p, remaining);
    if (mblen > remaining)
    {
        s = p;
        FAIL_OR_INCOMPLETE_AT_CHAR_START(JSON_INVALID_TOKEN);
    }
    p += mblen - 1;
}
else
{
    while (p < end - sizeof(Vector8) && ...)   /* unchanged */
}
```

The intent is only that a GB18030 trailing byte can sit in the 7-bit ASCII range (U+8846 is `0xD0 0x5C`, so its second byte *is* `\`), and must not be mistaken for a backslash or a control character. That is why the scan advances by whole characters for that encoding only.

**`test_escape.c` — "The two new functions can be folded into one"**

Done in `91a2f91ac9`: a single `test_gb18030_json_multibyte()` now covers the complete multibyte character (`need_escapes=true` and `need_escapes=false`) and the truncated one.

**`test_escape.c` — "Please follow the existing `test_gb18030_json()` as the template — it uses `raw_buf` + `NEVER_ACCESS_STR` + `VALGRIND_MAKE_MEM_NOACCESS`"**

The new function follows that template exactly:

```c
/* Catch any attempt to read beyond the supplied JSON text. */
raw_buf = createPQExpBuffer();
appendBinaryPQExpBuffer(raw_buf, (const char *) valid_input, input_len);
appendPQExpBufferStr(raw_buf, NEVER_ACCESS_STR);
VALGRIND_MAKE_MEM_NOACCESS(&raw_buf->data[input_len], raw_buf->len - input_len);
```

so an over-read past `input_len` is still caught, including for the truncated input where the scan has already advanced past an ASCII run.

**`test_escape.c` — "It's best not to put non-ASCII (Japanese) characters into the C source" / "Suggest using code points (e.g. `U+8846`) in the comments and keeping the hex arrays as-is"**

Both applied: the file contains no non-ASCII bytes at all, and the comments name code points while the byte arrays stay hex, e.g.

```c
0xa5, 0xab, 0xa5, 0xca, 0xa5, 0xc0, 0xa1, 0xa2, /* U+30AB etc. */
0xd0, 0x5c,                                     /* U+8846; second byte is '\' */
```

**`test_escape.c:238` — "example here"**

That line is from the two-function version; the function it pointed at was folded away. If you meant that the truncated case needs a worked example, it now asserts the reported position explicitly — the truncated input runs `{"ja":"ABCDEFG` then a lone `0xD0`, so the failure must not use the stale ASCII run start:

```c
/*
 * The truncated input advances beyond an ASCII run before ending at the
 * lead byte, so the reported position must not use the stale run start.
 */
```

and checks `lex->token_terminator == raw_buf->data + <lead byte offset>`, which also resolves the CodeRabbit note about `token_terminator` on truncated input.

Re-requesting review. The regression is `src/test/modules/test_escape` (`make check` in that directory) plus the main `json`/`jsonb` suites; I cannot run them here (no C toolchain, and the module is not built in my checkout), so CI is the execution proof.

btlqql:
All of the inline findings from the July review round are addressed at the current head `ee492062`; the replies were missing.

1. **`PG_ENCODING_IS_CLIENT_ONLY()` in `src/common/jsonapi.c`** — the arm is gone. The scan is now gated on the concrete encoding only, so the backend never takes a client-only branch:

```c
	if (lex->input_encoding == PG_GB18030)   /* jsonapi.c:2223 */
```

   `PG_ENCODING_IS_CLIENT_ONLY` no longer appears anywhere in `src/common/jsonapi.c`.

2. **Follow the `test_gb18030_json()` over-read template** — the new tests use the same `raw_buf` + `NEVER_ACCESS_STR` + `VALGRIND_MAKE_MEM_NOACCESS` pattern, so a read past `input_len` is caught under Valgrind:

```c
	raw_buf = createPQExpBuffer();
	appendBinaryPQExpBuffer(raw_buf, (const char *) valid_input, input_len);
	appendPQExpBufferStr(raw_buf, NEVER_ACCESS_STR);
	VALGRIND_MAKE_MEM_NOACCESS(&raw_buf->data[input_len],
							   raw_buf->len - input_len);
```
   (same pattern for `truncated_input`)

3. **Use code points in comments** — the byte comments now name code points, e.g. `/* U+8846; second byte is '\' */`, and the function-level comment no longer spells out the character.

4. **No non-ASCII in the C source** — `src/test/modules/test_escape/test_escape.c` contains zero bytes outside ASCII (verified by scanning the file contents).

5. **Fold the two new functions** — done in `91a2f91ac9`: the valid-multibyte and truncated-character cases are one `test_gb18030_json_multibyte()` that keeps both `need_escapes` paths and the truncated token-position assertion.

PR #1394, opened 2026-07-12
configure: require libxml2 by default and reject --without-libxml (#1353)

### LINKED ISSUE #1353: make can not find libxml2
### IvorySQL Version
5.4

### OS Version (uname -a)
openEuler 22.03 arm

Linux pg 5.10.0-182.0.0.95.oe2203sp3.aarch64 #1 SMP Sat Dec 30 13:16:24 CST 2023 aarch64 aarch64 aarch64 GNU/Linux

first step:
./configure --prefix=/opt/pg/isql --without-icu --without-zlib

second step:
make

### Current Behavior

make result error log (the last  three lines):

gcc -Wall -Wmissing-prototypes -Wpointer-arith -Wdeclaration-after-statement -Werror=vla -Wendif-labels -Wmissing-format-attribute -Wimplicit-fallthrough=3 -Wcast-function-type -Wshadow=compatible-local -Wformat-security -fno-strict-aliasing -fwrapv -fexcess-precision=standard -Wno-format-truncation -Wno-stringop-truncation -O2 -fPIC -fvisibility=hidden -shared -o ivorysql_ora.so  src/ivorysql_ora.o src/guc/guc.o src/datatype/oracharchar.o src/datatype/oracharbyte.o src/datatype/oravarcharchar.o src/datatype/oravarcharbyte.o src/datatype/oradate.o src/datatype/oratimestamp.o src/datatype/oratimestamptz.o src/datatype/oratimestampltz.o src/datatype/yminterval.o src/datatype/dsinterval.o src/datatype/compatible_oracle_precedence.o src/datatype/common_datatypes.o src/datatype/raw_long.o src/datatype/binary_float.o src/datatype/binary_double.o src/builtin_functions/character_datatype_functions.o src/builtin_functions/datetime_datatype_functions.o src/builtin_functions/numeric_datatype_functions.o src/builtin_functions/misc_functions.o src/merge/ora_merge.o src/sysview/sysview_functions.o src/xml_functions/ora_xml_functions.o -L../../src/port -L../../src/common    -Wl,--as-needed -Wl,-rpath,'/opt/pg/isql/lib',--enable-new-dtags -fvisibility=hidden -lxml2 
/usr/bin/ld: cannot find -lxml2
collect2: error: ld returned 1 exit status

### Expected behavior/code

the same step is normal in PG 18.4 on the same machine.

if IvorySQL need libxml2, why configure not report it.

if I do not want libxml2, the configure program does not have disable flag  in current.

btlqql:
## Summary

Fixes #1353.

`make` failed with `/usr/bin/ld: cannot find -lxml2` when building `ivorysql_ora` after a plain `./configure`, because `--with-libxml` was opt-in (default off), so libxml2 was never detected. The same `./configure` step works on upstream PostgreSQL.

Per the maintainer decision in #1353, libxml2 is now **mandatory** for IvorySQL because the `ivorysql_ora` plugin depends on it.

## Changes

- **`configure.ac`**: default `--with-libxml` to `yes`, and error out on `--without-libxml`:
  > `--without-libxml is not supported. The ivorysql_ora plugin depends on the libxml library.`
- **`configure`**: regenerated from `configure.ac` with autoconf 2.69 (only the libxml option handling changed).
- **`doc/src/sgml/installation.sgml`**: document that IvorySQL enables libxml2 by default and requires it.

## Behavior after this change

- `./configure` (no libxml flag): libxml2 is enabled and required. If libxml2-dev is missing, configure fails **early** with a clear message (`library 'xml2' ... is required for XML support` / `header file <libxml/parser.h> is required for XML support`) instead of failing later in `make`.
- `./configure --without-libxml`: configure fails immediately with the unsupported-option error.
- `make` / `ivorysql_ora`: builds and links libxml2 (`USE_LIBXML` defined, `-lxml2` linked).

## Verification

```
# --without-libxml is rejected
./configure --without-libxml --without-icu --without-zlib
# -> configure: error: --without-libxml is not supported. The ivorysql_ora plugin depends on the libxml library.

# default enables libxml2
./configure --without-icu --without-zlib
# -> checking whether to build with XML support... yes
# -> checking for libxml-2.0 >= 2.6.23... yes
# src/Makefile.global: with_libxml = yes
# src/include/pg_config.h: #define USE_LIBXML 1

make -C contrib/ivorysql_ora   # builds ivorysql_ora.so, links libxml2 symbols
```

The manually-applied `configure` edit is byte-identical to a fresh `autoconf 2.69` run on the edited `configure.ac` (only the libxml option section changed; no unrelated regeneration noise).

## Notes

- Leaves the conditional link in `contrib/ivorysql_ora/Makefile` (added in #1370) untouched; it is now always-true since `with_libxml` is always `yes` once configure succeeds, so it acts as harmless defense-in-depth.
- The meson path already defaults `libxml` to `auto` and links cleanly when absent, so it is unaffected; this change is scoped to the autoconf `./configure` path reported in the issue.

Assisted-by: opencode:glm-5.2

## Summary by CodeRabbit

- **New Features**
  - XML support is now enabled by default during installation.
  - The build process automatically enables the required XML functionality when a usable `libxml2` installation is available.

- **Bug Fixes**
  - Installation now reports a clear error when XML support is explicitly disabled or when `libxml2` cannot be found.

- **Documentation**
  - Updated installation guidance to explain the default XML support, dependency requirements, and unsupported `--without-libxml` option.

### DIFF
--- a/configure
+++ b/configure
@@ -1607,7 +1607,7 @@ Optional Packages:
   --with-ossp-uuid        obsolete spelling of --with-uuid=ossp
   --with-libcurl          build with libcurl support
   --with-libnuma          build with libnuma support
-  --with-libxml           build with XML support
+  --without-libxml        build with XML support
   --with-libxslt          use XSLT support when building contrib/xml2
   --with-system-tzdata=DIR
                           use system time zone data in DIR
@@ -9436,7 +9436,7 @@ $as_echo "#define USE_LIBXML 1" >>confdefs.h
 
       ;;
     no)
-      :
+      as_fn_error $? "--without-libxml is not supported. The ivorysql_ora plugin depends on the libxml library." "$LINENO" 5
       ;;
     *)
       as_fn_error $? "no argument expected for --with-libxml option" "$LINENO" 5
@@ -9444,7 +9444,9 @@ $as_echo "#define USE_LIBXML 1" >>confdefs.h
   esac
 
 else
-  with_libxml=no
+  with_libxml=yes
+
+$as_echo "#define USE_LIBXML 1" >>confdefs.h
 
 fi
 
--- a/configure.ac
+++ b/configure.ac
@@ -1214,8 +1214,9 @@ fi
 # XML
 #
 AC_MSG_CHECKING([whether to build with XML support])
-PGAC_ARG_BOOL(with, libxml, no, [build with XML support],
-              [AC_DEFINE([USE_LIBXML], 1, [Define to 1 to build with XML support. (--with-libxml)])])
+PGAC_ARG_BOOL(with, libxml, yes, [build with XML support],
+              [AC_DEFINE([USE_LIBXML], 1, [Define to 1 to build with XML support. (--with-libxml)])],
+              [AC_MSG_ERROR([--without-libxml is not supported. The ivorysql_ora plugin depends on the libxml library.])])
 AC_MSG_RESULT([$with_libxml])
 AC_SUBST(with_libxml)
 
--- a/doc/src/sgml/installation.sgml
+++ b/doc/src/sgml/installation.sgml
@@ -1201,6 +1201,14 @@ build-postgresql:
         </para>
 
         <para>
+         In <productname>IvorySQL</productname>, libxml2 is enabled by default
+         and is required, because the <filename>ivorysql_ora</filename> plugin
+         depends on it.  The <option>--without-libxml</option> option is not
+         supported; <command>configure</command> will fail if it is given, or if
+         a usable libxml2 installation cannot be found.
+        </para>
+
+        <para>
          To detect the required compiler and linker options, PostgreSQL will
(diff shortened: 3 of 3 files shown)


NotHimmel:
Hi @btlqql , thank you for the thorough investigation — the root-cause analysis in the description is excellent.

We've decided to go with a different approach for #1353, and the community has already merged #1370 to resolve it.

That solution matches PostgreSQL's general philosophy on optional dependencies: a library is only pulled in when the user actually asks for it (--with-...), rather than being made mandatory. 

Really appreciate the effort you put into this. Even though we went another way here, thank you again for contributing!

btlqql:
@NotHimmel Thanks for the clarification and for pointing me to #1370. That approach makes sense, especially keeping libxml2 optional unless explicitly requested. I’ll close this PR. Thanks for taking the time to review it.

PR #1401, opened 2026-07-15
Fix conflicting SHARD option validation

### LINKED ISSUE #1402: wrong variable is used in init_params()
## Bug Report
In src/backend/commands/sequence.c： init_param()  function, there is following code:

                else if(strcmp(defel->defname, "shard_extend") == 0 || strcmp(defel->defname, "shard_noextend") == 0
                        || strcmp(defel->defname, "noshard") == 0)
                {
                        shardcnt++;
                        if(keepcnt > 1)
                        {
                                ereport(ERROR,
                                        (errcode(ERRCODE_SYNTAX_ERROR),
                                         errmsg("dumplicate or conflicting SHARD EXTEND/NOEXTEND specifications"),
                                         parser_errposition(pstate, defel->location)));
                        }
                }

Obviously the shardcnt instead of keepcnt should be used to be compared to 1. 

### IvorySQL Version
IvorySQL master branch

### OS Version (uname -a)
N/A

messere1:
fix #1402 

## Summary

- use `shardcnt` when validating duplicate or conflicting SHARD options
- add Oracle regression coverage for `SHARD EXTEND` combined with `NOSHARD`

## Background

The Clang warning check in #1386 exposed that `shardcnt` was incremented but never read. The following validation condition mistakenly checked `keepcnt`, copied from the preceding KEEP/NOKEEP block. As a result, conflicting SHARD options were not rejected by the existing error path.

This runtime logic fix is split from #1386 so that the Clang CI PR remains CI-only.

## Testing

- `make check`
- `make oracle-check` (250 tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected validation for conflicting sequence sharding options.
  * Invalid combinations such as `SHARD EXTEND NOSHARD` are now consistently rejected with an accurate error location.

## Related issue

Discovered while implementing the Clang warning check requested in #1221. The runtime fix is kept in this separate PR so that #1386 remains CI-only.

### DIFF
--- a/src/backend/commands/sequence.c
+++ b/src/backend/commands/sequence.c
@@ -1770,7 +1770,7 @@ init_params(ParseState *pstate, List *options, bool for_identity,
 			|| strcmp(defel->defname, "noshard") == 0)
 		{
 			shardcnt++;
-			if(keepcnt > 1)
+			if(shardcnt > 1)
 			{
 				ereport(ERROR,
 					(errcode(ERRCODE_SYNTAX_ERROR),
--- a/src/oracle_test/regress/expected/sequence.out
+++ b/src/oracle_test/regress/expected/sequence.out
@@ -868,6 +868,11 @@ DROP SEQUENCE seq_shard4;
 create sequence seq_shard5 shard extend ;
 alter sequence seq_shard5 noshard ;
 DROP SEQUENCE seq_shard5;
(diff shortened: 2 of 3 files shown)


bigplaice:
Good catch! I agree this is a real bug. Thanks for fixing it.

bigplaice:
fix issue #1402

PR #1412, opened 2026-07-17
fix: fix issue 1411

### LINKED ISSUE #1411: DISCARD ALL/PACKAGES 执行失败时 DBMS_SESSION/DBMS_OUTPUT 状态未被清除
## Bug Report
原因（与 PR #1363/#1372 同一"绕过"家族）
/* Call the previous hook or standard function */
if (prev_ProcessUtility_hook)
      (*prev_ProcessUtility_hook)(pstmt, ...);     // 可能 ereport(ERROR)
  else
      standard_ProcessUtility(pstmt, ...);         // 可能 ereport(ERROR)
/* Reset ... after DISCARD ALL/PACKAGES */
  if (is_discard_reset)
 {
      ora_dbms_output_reset();
      ora_dbms_session_reset();
  }

在 PostgreSQL 中，ereport(ERROR) 会触发 PG_CATCH/longjmp，跳出当前函数。因此当standard_ProcessUtility 执行 DISCARD ALL 时抛出 ERROR（典型场景：在事务块中执行 DISCARD ALL，会被discard.c 中的 PreventInTransactionBlock 拒绝），第 192-196 行的重置代码永远不会执行。

影响
dbms_session.c 第 435-436 行明确声称：
> "Called by DISCARD ALL/PACKAGES so that a pooled connection does not leak one client's context to the next."
  
但由于这个 bug，在事务中执行 DISCARD ALL 失败后：
•DBMS_SESSION 应用上下文残留（SYS_CONTEXT 仍返回上一个客户端的值）
•DBMS_OUTPUT 缓冲区残留

在 PgBouncer 等连接池环境下，下一个客户端可能读取到上一个客户端的应用上下文，破坏行级安全（RLS）谓词所依赖的隔离性。

### IvorySQL Version
master

### OS Version (uname -a)
Linux KJ-ZX-5994-2 6.17.0-35-generic #35~24.04.1-Ubuntu SMP PREEMPT_DYNAMIC Tue May 26 19:30:42 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

enkilee:
fix #1411 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup reliability for `DISCARD ALL` and `DISCARD PACKAGES` so session/output state is reset even when the command fails.
  * Prevents context leakage by ensuring DBMS session state is cleared after failed `DISCARD ALL` attempts (including error paths).

* **Tests**
  * Added a regression test validating `DBMS_SESSION` behavior when `DISCARD ALL` is executed inside an explicit transaction.

* **Style**
  * Applied minor license/formatting adjustments with no functional changes.

### DIFF
--- a/contrib/ivorysql_ora/src/ivorysql_ora.c
+++ b/contrib/ivorysql_ora/src/ivorysql_ora.c
@@ -182,11 +182,13 @@ ivorysql_ora_ProcessUtility(PlannedStmt *pstmt,
 
 	/*
 	 * IvorySQL: call the previous hook or standard function.  Wrap it in
-	 * PG_TRY()/PG_CATCH() so that DBMS_OUTPUT and DBMS_SESSION state is reset
-	 * even when DISCARD ALL/PACKAGES itself fails (e.g. "DISCARD ALL cannot
-	 * run inside a transaction block").  Without this, a failed DISCARD in a
-	 * pooled connection would leak the previous client's application context
-	 * and DBMS_OUTPUT buffer.
+	 * PG_TRY()/PG_FINALLY() so that DBMS_OUTPUT and DBMS_SESSION state is
+	 * reset even when DISCARD ALL/PACKAGES itself fails (e.g. "DISCARD ALL
+	 * cannot run inside a transaction block").  Without this, a failed
+	 * DISCARD in a pooled connection would leak the previous client's
+	 * application context and DBMS_OUTPUT buffer.  PG_FINALLY() runs the
+	 * reset on both the success and error paths and re-throws any error
+	 * automatically, so the cleanup is not duplicated.
 	 */
 	PG_TRY();
 	{
@@ -197,24 +199,13 @@ ivorysql_ora_ProcessUtility(PlannedStmt *pstmt,
 			standard_ProcessUtility(pstmt, queryString, readOnlyTree,
 									context, params, queryEnv, dest, qc);
 	}
-	PG_CATCH();
+	PG_FINALLY();
 	{
 		if (is_discard_reset)
 		{
 			ora_dbms_output_reset();
 			ora_dbms_session_reset();
 		}
-		PG_RE_THROW();
 	}
 	PG_END_TRY();
-
-	/*
-	 * Reset DBMS_OUTPUT buffer and DBMS_SESSION context after DISCARD
-	 * ALL/PACKAGES
-	 */
-	if (is_discard_reset)
-	{
-		ora_dbms_output_reset();
-		ora_dbms_session_reset();
-	}
 }
--- a/contrib/ivorysql_ora/src/ivorysql_ora.c
+++ b/contrib/ivorysql_ora/src/ivorysql_ora.c
@@ -1,12 +1,12 @@
 /*-------------------------------------------------------------------------
  * Copyright 2025 IvorySQL Global Development Team
- * 
+ *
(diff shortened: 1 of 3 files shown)


NotHimmel:
Thanks for your contribution! We'll start reviewing it shortly.

contrib/ivorysql_ora/src/ivorysql_ora.c
NotHimmel: The reset block is duplicated — once in `PG_CATCH()` (before `PG_RE_THROW()`) and once after `PG_END_TRY()` for the success path. Since both the success and error paths do the exact same thing (reset when  `is_discard_reset` ),  this is suitable for `PG_FINALLY()`,  which runs on both paths and re-throws automatically.

NotHimmel (changes requested):

contrib/ivorysql_ora/src/ivorysql_ora.c
enkilee: roger. changed it. thx.

NotHimmel:
Fix looks correct. Could we add a regression test in sql/dbms_session.sql?  @enkilee

enkilee:
> Fix looks correct. Could we add a regression test in sql/dbms_session.sql? @enkilee

Done. thx.

NotHimmel:
@enkilee `DISCARD ALL` inside a transaction block is *rejected*, not half-executed. `PreventInTransactionBlock()` is the very first statement in `DiscardAll()` (`src/backend/commands/discard.c`), so none of `PortalHashTableDeleteAll()`, 
`SetPGVariable("session_authorization", ...)`, `ResetAllOptions()`, `DropAllPreparedStatements()`, `LockReleaseAll()`, ... runs either. A rejected statement has no side effects; not clearing DBMS_SESSION is consistent with core not clearing any of the above. Upstream is aware of the inconsistency and chose it deliberately — the comment there says "the idea is to catch mistakes". DISCARD ALL is all-or-nothing.

With this patch, a failed DISCARD would instead leave the session half-reset: the Oracle-compat state is wiped while every core item above is untouched, and `ROLLBACK` cannot restore it (session-level, non-transactional). An erroring statement leaving an unrollbackable partial side effect is harder to reason about and debug than the current behavior.

Do you have an actual error scenario for this? If so, please add reproduction steps; that would be a different bug.

enkilee:
> @enkilee `DISCARD ALL` inside a transaction block is _rejected_, not half-executed. `PreventInTransactionBlock()` is the very first statement in `DiscardAll()` (`src/backend/commands/discard.c`), so none of `PortalHashTableDeleteAll()`, `SetPGVariable("session_authorization", ...)`, `ResetAllOptions()`, `DropAllPreparedStatements()`, `LockReleaseAll()`, ... runs either. A rejected statement has no side effects; not clearing DBMS_SESSION is consistent with core not clearing any of the above. Upstream is aware of the inconsistency and chose it deliberately — the comment there says "the idea is to catch mistakes". DISCARD ALL is all-or-nothing.
> 
> With this patch, a failed DISCARD would instead leave the session half-reset: the Oracle-compat state is wiped while every core item above is untouched, and `ROLLBACK` cannot restore it (session-level, non-transactional). An erroring statement leaving an unrollbackable partial side effect is harder to reason about and debug than the current behavior.
> 
> Do you have an actual error scenario for this? If so, please add reproduction steps; that would be a different bug.

Sorry, I find DiscardAll comment the idea is to catch mistakes: this pr is not right. i will closed it. thx.

PR #1422, opened 2026-07-19
perf(plisql): optimize package dependency cleanup

### LINKED ISSUE #1426: Optimize PL/iSQL package dependency cleanup
## Problem

PL/iSQL package cleanup repeatedly scans dependency lists while ordering packages and finding non-package dependents. With many cached packages, repeated linear list operations make `DISCARD PACKAGE` unnecessarily expensive.

This was found during a source audit rather than from a reported production incident. It is most relevant to migrated schemas with hundreds or thousands of interdependent packages and sessions that reset package state. A generated reverse-order dependency chain is intentionally a synthetic worst case.

## Expected outcome

- reduce dependency sorting and dependent-function discovery to expected O(V + E) work
- preserve the existing stable ordering and cycle/fallback semantics
- add regression coverage for layered, independent, and cyclic package dependencies
- include a reproducible benchmark script that isolates `DISCARD PACKAGE` timing and supports multiple package counts

Z95-cmyk:
## Summary

- replace repeated package-list scans with a hash-indexed dependency graph and stable Kahn traversal
- collect dependent packages and standalone functions in one graph walk instead of recursively revisiting the graph from every package
- preserve the previous scan-pass and input ordering, including the existing fallback for cycles and dependencies outside the input list
- add Oracle-mode regression coverage for layered, independent, and cyclic package dependencies
- include an executable SQL benchmark for the old algorithm’s worst-case dependency order

This PR is now performance-only. The pre-existing cache correctness issue identified during review was split into #1453 with its own regression test and issue.

## Performance background

I did not encounter this in a production deployment, so I do not want to imply a real-world incident. Source inspection showed that the previous cleanup repeatedly scanned the remaining package list, making reverse-ordered dependency chains quadratic. The benchmark is an explicitly synthetic worst case intended to make the complexity and scaling reproducible; ordinary installations with fewer cached packages or shallower dependency graphs will see a smaller effect.

The reproducer is committed at `src/oracle_test/modules/test_package_reset/benchmark_discard_package_chain.sql`. Run it from an IvorySQL `psql` client against the base and patched revisions, for example:

```sh
psql -X -v ON_ERROR_STOP=1 -v package_count=1600 \
  -f src/oracle_test/modules/test_package_reset/benchmark_discard_package_chain.sql DATABASE
```

Only `DISCARD PACKAGE` is timed. Package bodies are created in reverse order and each body depends on its predecessor.

| Packages | Before | After |
| ---: | ---: | ---: |
| 3,200 | 8,040.366 ms | 16.647 ms |
| 6,400 | not run | 36.427 ms |

The 3,200-package synthetic case is about 483x faster, while the patched 3,200-to-6,400 result is close to linear scaling.

## Verification

- `make -C src/oracle_test/modules/test_package_reset oracle-check NO_TEMP_INSTALL=1` — 5/5 passed
- compared the new and old ordering algorithms on 360,000 randomized dependency graphs; all orders matched
- executed the committed benchmark with `package_count=10` after the split to verify that the reproducer runs end to end
- `git diff --check`

Fixes #1426

## AI assistance disclosure

This contribution was prepared with OpenAI Codex using GPT-5. The agent assisted with source and test inspection, issue scoping, drafting code and tests, running verification, benchmarking, and preparing the PR description. I directed the scope, reviewed the resulting changes against the source and test results, and remain responsible for the submission.

### DIFF
--- a/src/oracle_test/modules/test_package_reset/Makefile
+++ b/src/oracle_test/modules/test_package_reset/Makefile
@@ -7,7 +7,8 @@ EXTENSION = test_package_reset
 DATA = test_package_reset--1.0.sql
 
 REGRESS = test_package_reset
-ORA_REGRESS = $(REGRESS) test_reset_one_oid test_reset_all test_reset_defaults
+ORA_REGRESS = $(REGRESS) test_reset_one_oid test_reset_all test_reset_defaults \
+	test_discard_dependencies
 
 ifdef USE_PGXS
 PG_CONFIG = pg_config
--- a/src/oracle_test/modules/test_package_reset/benchmark_discard_package_chain.sql
+++ b/src/oracle_test/modules/test_package_reset/benchmark_discard_package_chain.sql
@@ -1,0 +1,61 @@
+-- Reproduce the synthetic worst case for DISCARD PACKAGE dependency cleanup.
+--
+-- Run this in an Oracle-compatible database from an IvorySQL psql client:
+--
+--   psql -X -v ON_ERROR_STOP=1 -v package_count=1600 \
+--     -f benchmark_discard_package_chain.sql DATABASE
+--
+-- Repeat with increasing package_count values (for example 400, 800, 1600)
+-- against the base and patched revisions.  Only the timed DISCARD PACKAGE
+-- statement is part of the measurement.  Package bodies are created in
+-- reverse order while each body depends on its predecessor, deliberately
+-- producing the order that maximizes repeated scans in the old algorithm.
+
+\set ON_ERROR_STOP 1
+\if :{?package_count}
+\else
+\set package_count 800
+\endif
+
+SET client_min_messages = warning;
+
+SELECT format(
+           'CREATE PACKAGE %1$I AS FUNCTION value RETURN NUMBER; END %1$I;',
+           'discard_bench_pkg_' || package_no)
+FROM generate_series(1, :package_count) AS package_no
+\gexec
+
+SELECT CASE
+           WHEN package_no = 1 THEN
+               format(
+                   'CREATE PACKAGE BODY %1$I AS '
+                   'FUNCTION value RETURN NUMBER AS BEGIN RETURN 1; END; '
+                   'END %1$I;',
+                   'discard_bench_pkg_' || package_no)
+           ELSE
+               format(
+                   'CREATE PACKAGE BODY %1$I AS '
+                   'FUNCTION value RETURN NUMBER AS BEGIN RETURN %2$I.value(); END; '
+                   'END %1$I;',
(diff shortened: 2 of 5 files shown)


hs-liuxh:
Hello! Thanks for the work. Just to confirm — is the main goal of this PR the package-cleanup performance optimization?

A few initial review comments:

1. Please split out the correctness fixes into a separate, higher-priority PR. This PR bundles several changes that are unrelated to performance — the dependent-function cache invalidation (`fn_xmin`), the stale-parent_func refresh in `plisql_get_subproc_*`, and `pkgcachelist = NIL`. These actually fix a real pre-existing crash. That's a valuable fix and should go in on its own, with a regression test that fails before / passes after — independently of the performance discussion.

2. About the benchmark: the description measures 3,200 packages arranged in a reverse-ordered dependency chain. That looks like a constructed worst case (the cost comes from the chain + ordering, not the count alone). Could you share the real use case or background where you hit slow DISCARD PACKAGE? Also, could you include the benchmark script so the numbers can be reproduced?

Thanks again!

Z95-cmyk:
Thanks for the clear review. Yes, the goal of this PR is now solely the package-cleanup performance optimization. I addressed both points:

1. I removed the correctness changes from this branch and opened the higher-priority fix separately as #1453, linked to #1452. Its focused regression reproduces the pre-existing failure independently: against the performance-only branch, the first cached standalone call after `DISCARD PACKAGE` terminates the backend with signal 11; the same test passes with #1453.
2. I committed the reproducer at `src/oracle_test/modules/test_package_reset/benchmark_discard_package_chain.sql` and updated this PR description with the exact command and methodology. I did not encounter this in production; it is an explicitly synthetic worst case discovered through source inspection, designed to expose the old repeated-scan complexity. The description now states that limitation rather than implying a production workload.

After the split, the package-reset module passes 5/5 tests, and the committed benchmark also runs end to end.

Z95-cmyk:
Follow-up on the latest automated review:

- The reported stale `parent_func` assignment is valid, but it is intentionally no longer part of this performance diff. It is fixed in the dedicated correctness PR #1453 by refreshing package-backed `FuncExpr` metadata against the current package cache entry before dereferencing it. That PR includes the independent fail-before/pass-after regression requested by the maintainer. Keeping it there avoids reintroducing correctness work into this PR.
- I did not add a SQL assertion for the internal fallback order because `DISCARD PACKAGE` exposes no stable per-package release-order result at SQL level; adding debug output solely for a regression would change production behavior. The committed cyclic test exercises successful cleanup/rebuild, while exact ordering was checked separately by comparing the old and new algorithms over 360,000 randomized dependency graphs, including cycles and dependencies outside the input list. All generated orders matched.

So no code change is needed in this performance branch for those two observations.

PR #1448, opened 2026-07-20
compat: restore upstream signatures for ExecTypeFromTL and friends (fix mysql_fdw build)

### LINKED ISSUE #1379: mysql_fdw fails to compile on IvorySQL 5.4
## Bug Report

#1312 

### IvorySQL Version

5.4

### Current Behavior

`mysql_fdw` fails to compile aginst IvorySQL 5.4 because it depends on PostgreSQL function `ExecTypeFromTL` which is declared in PostgreSQL as:

```c
TupleDesc ExecTypeFromTL(List *targetList);
```

However IvorySQL 5.4 changes the signature of this function to:

```c
TupleDesc ExecTypeFromTL(List *targetList, bool hasrowid);
```

This breaks source compatibility with extensions written against the standard PostgreSQL API.

123123213weqw:
## Problem

IvorySQL modifies four public PostgreSQL executor/tuple-descriptor APIs **in place** to add Oracle rowid support, breaking source compatibility for extensions written against standard PostgreSQL:

| API | Upstream PostgreSQL (PG18) | IvorySQL 5.4 (before) |
| --- | --- | --- |
| `ExecTypeFromTL` | `(List *)` | `(List *, bool hasrowid)` |
| `ExecCleanTypeFromTL` | `(List *)` | `(List *, bool hasrowid)` |
| `ExecInitJunkFilter` | `(List *, TupleTableSlot *)` | `(List *, bool hasrowid, TupleTableSlot *)` |
| `CreateTupleDesc` | `(int, Form_pg_attribute *)` | `(int, Form_pg_attribute *, bool tdhasrowid, bool is_sysattr)` |

`mysql_fdw` fails to compile because `mysql_fdw.c:613` calls the upstream one-argument form (`ExecTypeFromTL(scan_tlist)`), reported in #1379.

## Fix

**Restore the upstream signatures and symbols exactly**, and expose the IvorySQL rowid-aware variants under new `*WithRowId` names. Internal call sites that pass `hasrowid` (18 sites across 14 files) are updated to call the WithRowId variants.

| Upstream entry point (preserved, hasrowid=false) | IvorySQL variant |
| --- | --- |
| `ExecTypeFromTL(List *)` | `ExecTypeFromTLWithRowId(List *, bool hasrowid)` |
| `ExecCleanTypeFromTL(List *)` | `ExecCleanTypeFromTLWithRowId(List *, bool hasrowid)` |
| `ExecInitJunkFilter(List *, TupleTableSlot *)` | `ExecInitJunkFilterWithRowId(List *, bool hasrowid, TupleTableSlot *)` |
| `CreateTupleDesc(int, Form_pg_attribute *)` | `CreateTupleDescWithRowId(int, Form_pg_attribute *, bool tdhasrowid, bool is_sysattr)` |

Extensions written against upstream PostgreSQL continue to compile and link **unchanged** — both source-level and ABI compatibility hold, because `ExecTypeFromTL` etc. remain real exported functions with their original signatures.

## Why this approach (vs. the earlier macro dispatcher)

The first iteration of this PR used argument-count dispatcher macros to route calls based on arity. Reviewer @bigplaice pointed out two issues:
1. **ABI breakage** — replacing the `ExecTypeFromTL` symbol with `_1`/`_2` variants breaks pre-compiled extensions that expect the upstream symbol under the bare name.
2. **Unnecessary preprocessor complexity** — a plain rename of internal callers achieves the same source-compat goal more cleanly.

This rework drops all macros and restores the upstream symbols as real functions. (See [PR comment](https://github.com/IvorySQL/IvorySQL/pull/1448#issuecomment-5020989879).)

## Verification

- `make -C src/backend` builds clean with **zero new warnings**.
- `-std=c99 -pedantic -Wall` on a probe exercising all eight call combinations (4 upstream + 4 WithRowId) produces **no warnings** related to this change (no macros → no variadic-macro concerns).
- Symbols verified present in object files via `nm`:
  ```
  T ExecTypeFromTL            T ExecTypeFromTLWithRowId
  T ExecCleanTypeFromTL       T ExecCleanTypeFromTLWithRowId
  T ExecInitJunkFilter        T ExecInitJunkFilterWithRowId
  T CreateTupleDesc           T CreateTupleDescWithRowId
  ```
  The upstream symbols are real global functions, so pre-compiled extensions also resolve correctly.

Fixes #1379

## Summary by CodeRabbit

* **Compatibility**
  * Restored upstream-compatible calling conventions for tuple-descriptor creation and executor helpers, while retaining RowId-aware behavior via dedicated “WithRowId” APIs.
  * Updated tuple-type construction throughout planning/execution (including cleaned tuple types and junk-filter initialization) to consistently include or omit RowId as appropriate.
  * Adjusted tuple-descriptor usage for relation creation, system-attribute handling, and portal/result descriptors to align RowId expectations end-to-end.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
@@ -1675,7 +1675,7 @@ iso_year(int y, int m, int d)
  * Get the timezone value of the operating system.
  */
 static int64
-sys_time_zone()
+sys_time_zone(void)
 {
 
 #ifdef _WIN64
--- a/contrib/ivorysql_ora/src/builtin_packages/dbms_lock/dbms_lock.c
+++ b/contrib/ivorysql_ora/src/builtin_packages/dbms_lock/dbms_lock.c
@@ -81,7 +81,7 @@ static HTAB *dbms_lock_hash_table = NULL;
 /*
  * static routines
  */
-static void dbms_lock_init_hash_table();
+static void dbms_lock_init_hash_table(void);
 static bool dbms_lock_record_acquire(int64, int8);
 static bool dbms_lock_check(int64, int8);
 static bool dbms_lock_record_release(int64, int8);
--- a/src/backend/access/common/tupdesc.c
+++ b/src/backend/access/common/tupdesc.c
@@ -206,15 +206,23 @@ CreateTemplateTupleDesc(int natts)
 }
 
 /*
- * CreateTupleDesc
+ * CreateTupleDescWithRowId
  *		This function allocates a new TupleDesc by copying a given
  *		Form_pg_attribute array.
  *
  * Tuple type ID information is initially set for an anonymous record type;
  * caller can overwrite this if needed.
+ *
+ * IvorySQL rowid-aware variant: tdhasrowid records whether the resulting
+ * tuple descriptor carries an Oracle-style rowid attribute; is_sysattr
+ * indicates that the trailing attribute is a system attribute (and thus
+ * should be stripped from the count when no rowid is requested).  This is
+ * the real implementation; CreateTupleDesc() below forwards here with
+ * tdhasrowid=false and is_sysattr=false to preserve the upstream
+ * PostgreSQL signature.
  */
 TupleDesc
-CreateTupleDesc(int natts, Form_pg_attribute *attrs, bool tdhasrowid, bool is_sysattr)
+CreateTupleDescWithRowId(int natts, Form_pg_attribute *attrs, bool tdhasrowid, bool is_sysattr)
 {
 	TupleDesc	desc;
 	int			i;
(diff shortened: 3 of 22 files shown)


bigplaice:
I will take a look, thanks.

bigplaice:
A plain rename of 18 internal callers would achieve the same goal with far less preprocessor magic, why not? 
Your patch can fix build issue but also bring ABI breakage.

123123213weqw:
Good points, both.

You're right on the ABI breakage: by replacing the `ExecTypeFromTL` symbol with `_1`/`_2` variants, my patch fixes source compat but breaks the symbol for any pre-compiled extension built against upstream PG (and the macro dispatch also means the public name can no longer be taken as a function pointer). Given IvorySQL's compatibility-first positioning, preserving the upstream ABI is worth the rename.

I'll rework this to approach B: keep the upstream signatures and symbol names untouched (`ExecTypeFromTL(List *)` returns `ExecTypeFromTLInternal(tl, false, false)`, etc.), and expose the IvorySQL rowid-aware variants under new names (`ExecTypeFromTLWithRowId(List *, bool hasrowid)`, `ExecInitJunkFilterWithRowId(...)`, `CreateTupleDescWithRowId(...)`). Then update the ~18 internal call sites that currently pass `hasrowid` to use the new names. External extensions written against upstream PostgreSQL continue to compile and link unchanged.

Will push the rework to this PR once ready.

bigplaice:
LGTM. 
@hanjianqiao

hanjianqiao:
Thanks for all your work! The problem is solved.

123123213weqw:
Thanks @hanjianqiao! CI is green on the latest push. Ready to merge whenever you have a moment — no rush.

bigplaice:
@123123213weqw  Hi, please solve the conflict so that it can be merged.

123123213weqw:
CI failure fixed in `ccb7355df6`: use a strict `(void)` prototype for `dbms_lock_init_hash_table`.

Validated locally with the CI `-Werror` flags from a clean build; all GitHub Actions checks are now green.

123123213weqw:
> @123123213weqw Hi, please solve the conflict so that it can be merged.

done

PR #1451, opened 2026-07-20
fix: clean existing warnings for Clang -Werror

messere1:
## Summary

- always initialize timestamp parse state in the three-argument Oracle timestamp functions
- make the package parser's intentional enum/pointer round-trip explicit through `uintptr_t`
- make PostgreSQL/Oracle prompt enum conversions explicit in psql

## Motivation

Related to #1221 and #1386.

Switching the Clang job to `make COPT="-Werror"` exposed existing Clang 18 diagnostics:

- `-Wsometimes-uninitialized`
- `-Wvoid-pointer-to-enum-cast`
- `-Wenum-conversion`

The timestamp change fixes the actual uninitialized path instead of suppressing the warning. The remaining changes document existing representation conversions without changing their logic.

## Validation

- `CC=clang-18 CXX=clang++-18 ./configure --enable-nls ...`
- `make -j2 COPT="-Werror"` - passed
- GitHub Actions for `d60a5eadf84144443fe6608f44071fc7ff04903f` - all 7 checks passed:
  - `build`
  - `meson_build`
  - `pg_regression`
  - `oracle_regression`
  - `oracle_pg_regression`
  - `contrib_regression`
  - `policy`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of Oracle-compatible timestamp parsing when an empty NLS parameter is supplied.
  - Ensured `TO_TIMESTAMP` and `TO_TIMESTAMP_TZ` return consistent results whether the optional NLS parameter is omitted or explicitly empty.
  - Improved reliability when processing Oracle interactive prompts and package function calls.

- **Tests**
  - Added coverage for fractional-second and time-zone timestamp formats with empty NLS parameters.


hs-liuxh:
Thanks for your contribution! I will review it as soon as possible.

hs-liuxh:
This PR is highly valuable. Not only does it eliminate Clang compilation warnings, but the modifications to `datetime_datatype_functions.c` also fix an actual bug. This bug occurs when an empty string is passed as the third argument to the `sys.to_timestamp(text, text, text)` and `sys.to_timestamp_tz(text, text, text)` functions. 

Since the current test cases do not trigger the code path for this bug, could you please add the relevant test cases to `contrib/ivorysql_ora/sql/datatype_and_func_bugs.sql`?

messere1:
Thanks for the review. I added regression cases for both three-argument functions in `datatype_and_func_bugs.sql`. The tests use `pg_catalog.repeat('x', 0)` to construct a non-NULL zero-length text value, because an empty string literal is converted to NULL before reaching these STRICT functions in Oracle mode. All CI checks now pass, including `oracle_regression`.

PR #1453, opened 2026-07-20
fix(plisql): refresh package-dependent caches after discard

### LINKED ISSUE #1452: DISCARD PACKAGE can crash cached PL/iSQL callers
## Problem

A standalone PL/iSQL function that calls a package function can retain an `FmgrInfo.fn_extra` entry across `DISCARD PACKAGE`. Package cleanup frees the compiled dependent function but leaves its cached tuple identity valid, so the next invocation can treat the emptied function structure as reusable. The result is a backend segmentation fault.

Cached package `FuncExpr.parent_func` pointers and rebuilt function dependency lists can also outlive the package memory context or retain stale links after a reset.

This was found while separating correctness work from the package cleanup performance change in #1426. It was discovered by source inspection and a local regression reproducer, not from a reported production incident.

## Reproduction

1. Create a package function.
2. Create and invoke a standalone PL/iSQL function that calls it.
3. Prepare and execute a direct package call.
4. Run `DISCARD PACKAGE`.
5. Invoke the standalone function again.

On the current base revision, step 5 terminates the backend with `SIGSEGV`.

## Expected outcome

- invalidate the cached dependent-function tuple identity before freeing its compiled tree
- refresh package function expressions against the current package cache entry before dereferencing `parent_func`
- clear removed dependency lists so rebuilt functions do not retain stale links
- add a regression test covering cached callers, prepared plans, and repeated `DISCARD PACKAGE`
- demonstrate that the regression fails on the base revision and passes with the fix

Z95-cmyk:
## Summary

- invalidate cached PL/iSQL functions that depend on a discarded package
- refresh stale parent-function pointers from the package cache before reusing package subprogram metadata
- clear package cache dependency lists after unlinking them
- add focused regression coverage for cached standalone callers and prepared package calls across repeated `DISCARD PACKAGE` operations

This correctness fix was split from #1422 at the maintainer’s request so the crash fix can be reviewed independently from the performance optimization.

Fixes #1452

## Verification

- `make -C src/oracle_test/modules/test_package_reset oracle-check` — 5/5 tests passed
- ran the focused regression against the performance-only branch before applying this fix: the backend terminated with signal 11 on the first cached standalone call after `DISCARD PACKAGE`
- ran the same regression on this branch: it passes, including repeated discard and prepared-call coverage

## AI assistance disclosure

This contribution was prepared with OpenAI Codex using GPT-5. The agent assisted with source and test inspection, issue scoping, drafting code and tests, running verification, reproducing the pre-fix crash, and preparing the PR description. I directed the scope, reviewed the resulting changes against the source and test results, and remain responsible for the submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed package-backed functions becoming invalid after repeated `DISCARD PACKAGE` operations.
  * Preserved correct execution of cached callers and prepared statements after package cache invalidation.
  * Improved reliability of package function refreshes and execution plans following cache resets.

* **Tests**
  * Added regression coverage for package function caching, prepared execution, and repeated package discards.

### DIFF
--- a/src/oracle_test/modules/test_package_reset/Makefile
+++ b/src/oracle_test/modules/test_package_reset/Makefile
@@ -7,7 +7,8 @@ EXTENSION = test_package_reset
 DATA = test_package_reset--1.0.sql
 
 REGRESS = test_package_reset
-ORA_REGRESS = $(REGRESS) test_reset_one_oid test_reset_all test_reset_defaults
+ORA_REGRESS = $(REGRESS) test_reset_one_oid test_reset_all test_reset_defaults \
+	test_discard_function_cache
 
 ifdef USE_PGXS
 PG_CONFIG = pg_config
--- a/src/oracle_test/modules/test_package_reset/expected/test_discard_function_cache.out
+++ b/src/oracle_test/modules/test_package_reset/expected/test_discard_function_cache.out
@@ -1,0 +1,64 @@
+-- Cached callers and plans must survive repeated package cache resets.
+CREATE PACKAGE discard_cache_pkg AS
+    FUNCTION value RETURN NUMBER;
+END discard_cache_pkg;
+/
+CREATE PACKAGE BODY discard_cache_pkg AS
+    FUNCTION value RETURN NUMBER AS
+    BEGIN
+        RETURN 42;
+    END;
+END discard_cache_pkg;
+/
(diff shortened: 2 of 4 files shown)


hs-liuxh:
Thanks for your work! This fix is crucial for IvorySQL's stability. We'll discuss it further in our review meeting.

PR #1454, opened 2026-07-20
Optimize ora_regexp_count: drop wasteful text_to_cstring + strlen

xuhuicai:
### Optimization

The `ora_regexp_count` function had unnecessary full string copy and scanning when counting newline characters `\n`:

- **Old code**: `text_to_cstring(s)` (`palloc` + `memcpy` full copy) + `strlen(str)` (full scan)
- **New code**: Use `VARDATA_ANY(s)` to point to the original bytes directly + reuse the existing `src_text_len` (line 767)

Each call saves: 1 `palloc` + 1 `memcpy` + 1 full-string `strlen`.

### Behavioral difference

Only matters for the edge case where the string contains embedded `\0` — the old `strlen` truncates at `\0`, while the new code uses the real byte length, which is more accurate. Embedded `\0` in text type is extremely rare. Multi-byte safe: `\n` (0x0A) does not appear in UTF-8 continuation bytes.

#1455 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved regular expression count handling for text containing newline characters.
  - Ensured newline occurrences are counted accurately across the full text, including when the text ends with a newline.
  - Removed issues caused by relying on string termination when processing text values.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
@@ -768,19 +768,16 @@ ora_regexp_count(PG_FUNCTION_ARGS)
 	
 	if (PG_NARGS() == 4)
 	{
-		char	*str = text_to_cstring(s);
-		int 	 len = strlen(str);
-		int 	 i = 0;
-		if (len > 0) 
+		char   *str = VARDATA_ANY(s);
+		int     i;
+
+		for (i = 0; i < src_text_len; i++)
 		{
-			for (i = 0;i < len; i++)
-			{
-				if (str[i] == '\n')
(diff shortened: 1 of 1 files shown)


xuhuicai:
Related to #1455

hs-liuxh:
Thanks for your contribution! The change itself looks good, I'll review it shortly.

One small request: since IvorySQL is an international community, could you please translate the PR description and the linked issue #1455 into English? That will let more people join the review.

xuhuicai:
Done. Both the PR description and issue #1455 have been updated to English. Thanks for the reminder.

hanjianqiao:
Nice work! No behavioral difference here though — IvorySQL doesn't support CHR(0). Merging shortly.

PR #1475, opened 2026-07-29
fix build issues and stop exit()ing from libpq

### LINKED ISSUE #1474: Fix a bunch of compilation issues observed on MacOS
A bunch of linker and libpq related issues were observed on IvorySQL code base. Need to fix those properly.

db-nikhil:
Fixes #1474
 
- contrib/ivorysql_ora/Makefile: move the `ifeq ($(with_libxml),yes) SHLIB_LINK += -lxml2` block to after including Makefile.global/PGXS. $(with_libxml) isn't defined until that include runs, so the block previously always evaluated false and -lxml2 was silently never linked. Linux linker was more forgiving, but macOS is not.

- src/interfaces/libpq/.gitignore, ivy_sema.c: stop tracking the generated ivy_sema.c symlink. Ignore it instead, like pg_sema.c already is.

- src/interfaces/libpq/sysv_ivysema.c, posix_ivysema.c: replace exit()/silent-return with abort() when a semaphore primitive fails (semget/semctl/semop, sem_wait/sem_post). These files ship inside libpq, a library linked into arbitrary host applications, not the postgres backend this code was adapted from:

    * exit() terminates the entire host process on the library's own initiativ which is not great! abort() keeps the "cannot safely continue" behavior on the application side.

## Summary by CodeRabbit

* **Build Improvements**
  * Improved portability and consistency across Meson and Make-based builds.
  * Added automatic handling for platform-specific semaphore implementations.
  * Improved generation and tracking of required source files during builds.
  * Preserved XML library linking behavior when XML support is enabled.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -86,10 +86,6 @@ ORA_REGRESS = \
 	utl_encode \
 	dbms_session
 
-ifeq ($(with_libxml),yes) 
-SHLIB_LINK += -lxml2
-endif
-
 # Include path for plisql.h (needed by dbms_utility)
 PG_CPPFLAGS += -I$(top_srcdir)/src/pl/plisql/src
 
@@ -102,6 +98,10 @@ subdir = contrib/ivorysql_ora
 top_builddir = ../..
 include $(top_builddir)/src/Makefile.global
 include $(top_srcdir)/contrib/contrib-global.mk
+endif
+
+ifeq ($(with_libxml),yes)
+SHLIB_LINK += -lxml2
 endif
 
 
--- a/src/fe_utils/meson.build
+++ b/src/fe_utils/meson.build
@@ -7,6 +7,12 @@ run_command(
    '../backend/oracle_parser/ora_keywords.c'],
   check:false)
 
+run_command(
+  [ln,
+   '-s',
+   '../oracle_fe_utils/ora_string_utils.c'],
+  check:false)
+
 fe_utils_sources = files(
   'archive.c',
   'astreamer_file.c',
--- a/src/fe_utils/.gitignore
+++ b/src/fe_utils/.gitignore
@@ -1,3 +1,4 @@
 /psqlscan.c
 /ora_keywords.c
 /ora_kwlist_d.h
+/ora_string_utils.c
--- a/src/fe_utils/ora_string_utils.c
+++ b/src/fe_utils/ora_string_utils.c
@@ -1,1 +1,0 @@
-../../src/oracle_fe_utils/ora_string_utils.c
--- a/src/interfaces/libpq/.gitignore
+++ b/src/interfaces/libpq/.gitignore
@@ -1,3 +1,4 @@
 /exports.list
+/ivy_sema.c
 /libpq-refs-stamp
 /tmp_check/
--- a/src/interfaces/libpq/ivy_sema.c
+++ b/src/interfaces/libpq/ivy_sema.c
@@ -1,1 +1,0 @@
-../../../src/interfaces/libpq/posix_ivysema.c
--- a/src/interfaces/libpq/meson.build
+++ b/src/interfaces/libpq/meson.build
@@ -17,9 +17,20 @@ libpq_sources = files(
   'libpq-events.c',
   'pqexpbuffer.c',
   'ivy-exec.c',
-  'ivy_sema.c',
 )
 libpq_so_sources = [] # for shared lib, in addition to the above
+
+if cdata.has('USE_UNNAMED_POSIX_SEMAPHORES') or cdata.has('USE_NAMED_POSIX_SEMAPHORES')
+  libpq_sources += files('posix_ivysema.c')
+endif
+
+if cdata.has('USE_SYSV_SEMAPHORES')
+  libpq_sources += files('sysv_ivysema.c')
+endif
+
+if cdata.has('USE_WIN32_SEMAPHORES')
+  libpq_sources += files('win32_ivysema.c')
+endif
 
 if host_system == 'windows'
   libpq_sources += files('pthread-win32.c', 'win32.c')
@@ -27,6 +38,9 @@ if host_system == 'windows'
     '--NAME', 'libpq',
     '--FILEDESC', 'PostgreSQL Access Library',])
 endif
+
+# autoconf generates the file there, ensure we get a conflict
+generated_sources_ac += {'src/interfaces/libpq': ['ivy_sema.c']}
 
 if ssl.found()
   libpq_sources += files('fe-secure-common.c')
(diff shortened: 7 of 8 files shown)


db-nikhil:
@coderabbitai review

hanjianqiao:
Thank you for your contribution! We'll review it shortly.

db-nikhil:
@coderabbitai resume

db-nikhil:
@coderabbitai review

src/fe_utils/meson.build
```diff
@@ -7,6 +7,12 @@ run_command(
    '../backend/oracle_parser/ora_keywords.c'],
   check:false)
 
+run_command(
```
hanjianqiao, line 10: Creating symlinks at build time is preferred over committing them to the repository. The build always produces the same symlinks, and switching to committed symlinks would also require changes to the configure scripts.

hanjianqiao (changes requested):

hanjianqiao:
Thank you so much for your contribution! Before we move forward, could you please address the review comments?

src/fe_utils/meson.build
db-nikhil: Ok, cleaned up the existing symlink as well and resubmitted. Thanks

contrib/ivorysql_ora/Makefile
```diff
@@ -104,5 +100,9 @@ include $(top_builddir)/src/Makefile.global
 include $(top_srcdir)/contrib/contrib-global.mk
 endif
 
+ifeq ($(with_libxml),yes)
+SHLIB_LINK += -lxml2
+endif
+
```
db-nikhil, line 107: Can you please clarify what do you want here? AFAICS I don't see any connection with src/interfaces/libpq/meson.build here in this Makefile.

db-nikhil:
> Thank you so much for your contribution! Before we move forward, could you please address the review comments?

Thanks. I believe I have addressed your concerns appropriately.

hanjianqiao:
@db-nikhil  Thanks so much, and I'm really sorry for the misleading comments earlier — that was my mistake. Your previous changes are actually good enough, so could you please revert them back? Sorry again for the confusion I caused. As for the abort() actions, we're still reviewing them carefully, and I'll make sure to reach out to you as soon as we have a result.

src/interfaces/libpq/posix_ivysema.c
```diff
@@ -181,7 +181,13 @@ PGSemaphoreLock(PGSemaphore sema)
...
 	{
 		fprintf(stderr, "sem_wait failed: %m");
-		return;
+		/*
+		 * Returning here would let the caller proceed believing it holds
+		 * the lock, racing whatever this semaphore was meant to protect.
+		 * abort() avoids exit()'s host-app cleanup running mid-invariant-
+		 * violation and leaves a core dump for diagnosis.
+		 */
+		abort();
```
hanjianqiao: As for the abort() actions, we're still reviewing them carefully, and I'll make sure to reach out to you as soon as we have a result.

hanjianqiao (commented):
Thanks

db-nikhil:
> @db-nikhil Thanks so much, and I'm really sorry for the misleading comments earlier — that was my mistake. Your previous changes are actually good enough, so could you please revert them back? Sorry again for the confusion I caused. As for the abort() actions, we're still reviewing them carefully, and I'll make sure to reach out to you as soon as we have a result.

hai @hanjianqiao I think the meson changes were required, no? Not sure what you want me to revert.

hanjianqiao:
@db-nikhil Could you please make the following changes to two files:

1. Revert `meson.build`.
2. Revert `src/fe_utils/meson.build`, then add an `ln` command to create a symbolic link for `ora_string_utils.c`.

db-nikhil:
> * Revert `meson.build`.
> * Revert `src/fe_utils/meson.build`, then add an `ln` command to create a symbolic link for `ora_string_utils.c`.

Ah ok. Reverted and pushed again. Thanks @hanjianqiao

hanjianqiao:
Hi, @db-nikhil. After some discussion, we feel that sysv_ivysema.c, posix_ivysema.c and win32_ivysema.c are part of the libpq client library, and the callers are external applications (see Ivyconnectdb in ivy-exec.c, which corresponds to PQconnectdb). Calling exit() or abort() would tear down the host process directly, leaving the caller with no error code to inspect and no chance to handle the failure gracefully. A library really shouldn't terminate the process on its own.

For comparison, posix_ivysema.c in the same directory handles this well: it returns an error code on failure (e.g. PGSemaphoreCreate returning 1 or 0), which is exactly the behavior you'd want from a library.

Would it make sense to follow the same pattern here and propagate an error code back to the caller instead?

Also, if you have a moment, it would help to have the steps to reproduce the situation where exit() causes the host process using libpq to terminate. Thanks!

db-nikhil:
> we feel that sysv_ivysema.c, posix_ivysema.c and win32_ivysema.c are part of the libpq client library, and the callers are external applications (see Ivyconnectdb in ivy-exec.c, which corresponds to PQconnectdb). Calling exit() or abort() would tear down the host process directly, leaving the caller with no error code to inspect and no chance to handle the failure gracefully. A library really shouldn't terminate the process on its own.
> 
> For comparison, posix_ivysema.c in the same directory handles this well: it returns an error code on failure (e.g. PGSemaphoreCreate returning 1 or 0), which is exactly the behavior you'd want from a library.
> 
> Would it make sense to follow the same pattern here and propagate an error code back to the caller instead?

@hanjianqiao 

We could do this but the main issue is the change radius and whether it's worth it. 

`PGSemaphoreLock/PGSemaphoreUnlock` are called from ~30-40 sites in ivy-exec.c, many mid-sequence while other locks are already held. A real fix means each of those call sites gets an error check that unwinds whatever locks it's already holding and returns failure through that function's own error-reporting convention — that's a large, careful pass, not a signature change plus find-replace.

Also, on further thinking, we shouldn't expect these to fail regularly. Once PGSemaphoreCreate has succeeded, continuing past any failures means operating on a lock state you can no longer trust, which is worse than stopping. Thus PGSemaphoreCreate returning an error code is a different case entirely — that's a genuinely expected failure mode, which is why it's handled differently already.

So if you ask me, I'd leave PGSemaphoreLock/PGSemaphoreUnlock as abort-on-failure as this patch does.

db-nikhil:
@hanjianqiao if this `abort()` stuff is taking time, since other users are also running into compile issues, I can create a version of this PR without the `abort()` part and then we can discuss that separately? Please let me know.

hanjianqiao:
@db-nikhil Yes, let's do that. Please go ahead with the version without it. Thanks for the suggestion!

db-nikhil:
@hanjianqiao done. Now this commit only focuses on build fixes and only the macos specific sysv_ivysema.c changes since compilation doesn't work on macos without this. This fix is not used in Linux which is the primary used platform and those sema files have been left untouched.

db-nikhil:
@hanjianqiao to keep things clean, I have even removed the macos fixes now.. we can discuss the abort/exit stuff in libpq completely separately now.

PR #1489, opened 2026-08-01
Fix Oracle regexp newline matching semantics

### LINKED ISSUE #1481: REGEXP_COUNT returns wrong value
## Bug Report

REGEXP_COUNT returns wrong value

Using IvorySQL master, Oracle compatible mode:

```sql
[ORA]postgres=# SELECT REGEXP_COUNT('bX' || 'cX', '.X', 1, 'c') AS r2 from dual;
 r2 
----
  2
(1 row)

[ORA]postgres=# SELECT REGEXP_COUNT(chr(10) || 'bX' || 'cX', '.X', 1, 'c') AS r2 from dual;
 r2 
----
  1
(1 row)

[ORA]postgres=# SELECT REGEXP_COUNT(chr(10) || 'bX' || chr(10) || 'cX', '.X', 1, 'c') AS r2 from dual;
 r2 
----
  0
(1 row)
```

Using Oracle 19c:

```sql
SELECT REGEXP_COUNT('bX' || 'cX', '.X', 1, 'c') AS r2 from dual;
 r2 
----
  2
(1 row)

SELECT REGEXP_COUNT(chr(10) || 'bX' || 'cX', '.X', 1, 'c') AS r2 from dual;
 r2 
----
  2
(1 row)

SELECT REGEXP_COUNT(chr(10) || 'bX' || chr(10) || 'cX', '.X', 1, 'c') AS r2 from dual;
 r2 
----
  2
(1 row)
```

ai-yang:
## Summary

- Stop changing `REGEXP_COUNT` totals based on newlines outside the matched
  substrings.
- Add `REG_NLDOT`, which excludes newline from `.` without excluding it from
  negated bracket expressions.
- Handle Oracle's documented `m` and `n` behaviors centrally in
  `ora_parse_re_flags()`: `m` enables multiline anchors, while `n` allows dot
  to match newline.
- Share the behavior across `REGEXP_COUNT`, `REGEXP_SUBSTR`, `REGEXP_INSTR`,
  and `REGEXP_REPLACE`, with regression coverage for default, `c`, `n`, and
  `m` modes.

## Root cause

`ora_regexp_count()` counted every newline in the complete source and adjusted
the regex engine's result when the pattern began with `.` or `^`. Newlines
outside the matched substrings therefore changed the returned count.

PostgreSQL's newline-sensitive regex mode combines dot, bracket-expression,
and anchor behavior. Oracle's `m` and `n` options control multiline anchors
and dot-newline matching separately. The existing `REG_NLSTOP` flag could not
provide the required dot-only behavior because it also prevents negated
bracket expressions from matching newline.

This change adds a dot-specific regex flag and applies the Oracle option
semantics in the shared Oracle flag parser used by all four functions.

## Review update

- Removed the unsupported last-option-wins interpretation and the associated
  combined-option assertions and regression cases.
- Kept `m` and `n` as independent controls for their documented behaviors.
- Rechecked all callers and the adjacent regex-engine flag handling; no other
  PostgreSQL, HBA, JSONPath, or `REGEXP_LIKE` path is changed by the new flag.

## Validation

Run in the Docker development environment:

- `make -j2 -C contrib/ivorysql_ora oracle-check ORA_REGRESS=ora_character_datatype_functions`
  - 1/1 test passed
- `make -j2 -C contrib/ivorysql_ora oracle-check`
  - 24/24 tests passed
- `make -j2 -C src/test/modules/test_regex check`
  - 2/2 tests passed
- `make -j2 check`
  - 244/244 tests passed
- `make -j2 oracle-check ORA_REGRESS=alter_index EXTRA_INSTALL=contrib/amcheck`
  - 254/254 tests passed
- Manual Oracle-mode checks covered Unicode text, unrelated leading/middle/
  trailing newlines, negated classes, regex-cache isolation, the original
  issue reproduction, and a 1,000-match stress case.
- `git diff --check`

Closes #1481

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -2111,6 +2111,61 @@ good', '.', 1, 'n') from dual;
            23
 (1 row)
 
+-- Newlines outside the matches must not change the count.
+select regexp_count('bX' || 'cX', '.X', 1, 'c') as no_newline,
+       regexp_count(chr(10) || 'bX' || 'cX', '.X', 1, 'c') as one_newline,
+       regexp_count(chr(10) || 'bX' || chr(10) || 'cX',
+                    '.X', 1, 'c') as two_newlines
+from dual;
+ no_newline | one_newline | two_newlines 
+------------+-------------+--------------
+          2 |           2 |            2
+(1 row)
+
+-- The m and n options control line anchors and dot matching independently.
+select regexp_count('X' || chr(10), 'X.', 1, 'c') as c_mode,
+       regexp_count('X' || chr(10), 'X.', 1, 'n') as n_mode,
+       regexp_count('X' || chr(10), 'X.', 1, 'm') as m_mode,
+       regexp_count('X' || chr(10), 'X.', 1, 'mn') as mn_mode
+from dual;
+ c_mode | n_mode | m_mode | mn_mode 
+--------+--------+--------+---------
+      0 |      1 |      0 |       1
+(1 row)
+
+-- The n option must not add newlines that are outside a match.
+select regexp_count(chr(10) || 'bX' || chr(10) || 'cX',
+                    '.X', 1, 'n') as n_mode
+from dual;
+ n_mode 
+--------
+      2
+(1 row)
+
+-- A negated bracket expression still matches a newline without the n option.
+select regexp_count('a,b' || chr(10) || 'c', '[^,]+', 1, 'c')
+       as negated_class
+from dual;
+ negated_class 
+---------------
+             2
+(1 row)
+
+-- The n option does not enable multiline anchors.
+select regexp_count('x' || chr(10) || 'b' || chr(10) || 'y',
+                    '^b$', 1, 'c') as c_mode,
+       regexp_count('x' || chr(10) || 'b' || chr(10) || 'y',
+                    '^b$', 1, 'n') as n_mode,
+       regexp_count('x' || chr(10) || 'b' || chr(10) || 'y',
+                    '^b$', 1, 'm') as m_mode,
+       regexp_count('x' || chr(10) || 'b' || chr(10) || 'y',
+                    '^b$', 1, 'mn') as mn_mode
+from dual;
+ c_mode | n_mode | m_mode | mn_mode 
+--------+--------+--------+---------
+      0 |      0 |      1 |       1
+(1 row)
+
 /*
  * length
  */
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -2166,6 +2166,103 @@ from dual;
       0 |      0 |      1 |       1
 (1 row)
 
+-- REGEXP_COUNT, SUBSTR, INSTR, and REPLACE share the m and n semantics.
+select length(regexp_substr('X' || chr(10), 'X.', 1, 1))
+       as default_mode,
+       length(regexp_substr('X' || chr(10), 'X.', 1, 1, 'c'))
+       as c_mode,
+       length(regexp_substr('X' || chr(10), 'X.', 1, 1, 'n'))
+       as n_mode,
+       length(regexp_substr('X' || chr(10), 'X.', 1, 1, 'm'))
+       as m_mode,
+       length(regexp_substr('X' || chr(10), 'X.', 1, 1, 'mn'))
+       as mn_mode
+from dual;
+ default_mode | c_mode | n_mode | m_mode | mn_mode 
+--------------+--------+--------+--------+---------
+              |        |      2 |        |       2
+(1 row)
+
+select regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1) as default_mode,
+       regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1, 'c') as c_mode,
+       regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1, 'n') as n_mode,
+       regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1, 'm') as m_mode,
+       regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1, 'mn') as mn_mode
+from dual;
+ default_mode | c_mode | n_mode | m_mode | mn_mode 
+--------------+--------+--------+--------+---------
+              |        |        | b      | b
+(1 row)
+
+select regexp_instr('X' || chr(10), 'X.', 1, 1, 0) as default_mode,
+       regexp_instr('X' || chr(10), 'X.', 1, 1, 0, 'c') as c_mode,
+       regexp_instr('X' || chr(10), 'X.', 1, 1, 0, 'n') as n_mode,
+       regexp_instr('X' || chr(10), 'X.', 1, 1, 0, 'm') as m_mode,
+       regexp_instr('X' || chr(10), 'X.', 1, 1, 0, 'mn') as mn_mode
+from dual;
+ default_mode | c_mode | n_mode | m_mode | mn_mode 
+--------------+--------+--------+--------+---------
+ 0            | 0      | 1      | 0      | 1
+(1 row)
+
+select regexp_instr('x' || chr(10) || 'bb' || chr(10) || 'y',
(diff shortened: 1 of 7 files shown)


bigplaice:
Thanks for contribution. We'll schedule a review for the patch.

contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
```diff
@@ -756,29 +759,32 @@ ora_regexp_count(PG_FUNCTION_ARGS)
 								 errmsg("the 4th argument is illegal parameter for function. the parameter can be one of x, m, i, c, n.")));
 		}
 		ora_parse_re_flags(&flags, match_param);
+
+		/* Oracle treats the m and n options as independent dimensions. */
+		for (i = 0; paramstr[i] != '\0'; i++)
```
bigplaice: In PG code parse_re_flags(), 'm' and 'n' are treated as synonym, which is different with Oracle. Consider that parse_re_flags() is called (via ora_parse_re_flags()) by more function besides ora_regexp_count:

ora_substr_text_regexp
ora_instr_text_regexp
ora_regexp_replace

I'd like to suggest to modify ora_parse_re_flags() and process 'm' and 'n' there properly. So that other three function would get benefit too.

ai-yang: Addressed in c0588454e4. `ora_parse_re_flags()` now maps Oracle `m` and `n` independently, and the duplicate option handling has been removed from `ora_regexp_count()`. Added default/`c`/`n`/`m`/`mn` regression matrices for `REGEXP_SUBSTR`, `REGEXP_INSTR`, and `REGEXP_REPLACE`. The targeted test passed 1/1, the full `ivorysql_ora` suite passed 24/24, and the core regex module passed 2/2.

contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
```diff
@@ -2111,6 +2111,158 @@ good', '.', 1, 'n') from dual;
...
+       regexp_count('x' || chr(10) || 'b' || chr(10) || 'y',
+                    '^b$', 1, 'n') as n_mode,
+       regexp_count('x' || chr(10) || 'b' || chr(10) || 'y',
+                    '^b$', 1, 'm') as m_mode,
+       regexp_count('x' || chr(10) || 'b' || chr(10) || 'y',
+                    '^b$', 1, 'mn') as mn_mode
+from dual;
+ c_mode | n_mode | m_mode | mn_mode 
+--------+--------+--------+---------
+      0 |      0 |      1 |       1
```
bigplaice: I tried this case in Oracle 23ai Free Release 23.0.0.0.0 , the result is different, could you fix it?
    C_MODE     N_MODE     M_MODE    MN_MODE
---------- ---------- ---------- ----------
         0          0          1          0

contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
```diff
@@ -2111,6 +2111,158 @@ good', '.', 1, 'n') from dual;
...
+       regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1, 'n') as n_mode,
+       regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1, 'm') as m_mode,
+       regexp_substr('x' || chr(10) || 'b' || chr(10) || 'y',
+                     '^b$', 1, 1, 'mn') as mn_mode
+from dual;
+ default_mode | c_mode | n_mode | m_mode | mn_mode 
+--------------+--------+--------+--------+---------
+              |        |        | b      | b
```
bigplaice: The result of Oracle is:

Image

contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
```diff
@@ -2111,6 +2111,158 @@ good', '.', 1, 'n') from dual;
...
+       regexp_instr('x' || chr(10) || 'bb' || chr(10) || 'y',
+                    '^bb$', 1, 1, 0, 'n') as n_mode,
+       regexp_instr('x' || chr(10) || 'bb' || chr(10) || 'y',
+                    '^bb$', 1, 1, 0, 'm') as m_mode,
+       regexp_instr('x' || chr(10) || 'bb' || chr(10) || 'y',
+                    '^bb$', 1, 1, 0, 'mn') as mn_mode
+from dual;
+ default_mode | c_mode | n_mode | m_mode | mn_mode 
+--------------+--------+--------+--------+---------
+ 0            | 0      | 0      | 3      | 3
```
bigplaice: The result in Oracle is:
Image

contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
ai-yang: Thanks for verifying this against Oracle 23ai. Addressed in `a55dc073f2`: `ora_parse_re_flags()` now processes `m` and `n` as mutually overriding newline modes, so the last occurrence wins.

The anchor regression now expects `default/c/n/m/mn/nm = 0/0/0/1/0/1`. Thus the reported `mn = 0` result is preserved, and the reverse `nm = 1` case is covered as well. The full extension regression suite (24/24) and core regex tests (2/2) pass.

contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
ai-yang: Thanks for the Oracle 23ai result. Addressed in `a55dc073f2`: combined `m`/`n` options now use the last occurrence.

The anchor regression now expects `default/c/n/m/mn/nm = NULL/NULL/NULL/b/NULL/b`, matching the reported `m = b` and `mn = NULL` behavior and adding the reverse `nm = b` case. The full extension regression suite (24/24) and core regex tests (2/2) pass.

contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
ai-yang: Thanks for confirming the positions on Oracle 23ai. Addressed in `a55dc073f2`: the shared flag parser now applies last-option-wins semantics.

The anchor regression now expects `default/c/n/m/mn/nm = 0/0/0/3/0/3`, matching the reported results and covering the reverse `nm` order. The full extension regression suite (24/24) and core regex tests (2/2) pass.

contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
ai-yang: Follow-up correction: my earlier reply described `m` and `n` as independent controls. Based on the Oracle 23ai results in the later review, `a55dc073f2` now implements them as distinct but mutually overriding newline modes, with the last occurrence taking precedence (`mn` uses `n`; `nm` uses `m`).

The logic remains centralized in `ora_parse_re_flags()` and is covered across `REGEXP_COUNT`, `REGEXP_SUBSTR`, `REGEXP_INSTR`, and `REGEXP_REPLACE`.

src/backend/utils/adt/regexp.c
```diff
@@ -465,8 +465,10 @@ ora_parse_re_flags(pg_re_flags *flags, text *opts)
 	parse_re_flags(flags, opts);
 
 	/*
-	 * Oracle treats m and n as independent options: m changes the behavior of
-	 * line anchors, while n allows dot to match a newline.  PostgreSQL's
+	 * Oracle treats m and n as distinct, mutually overriding newline modes.
```
bigplaice: Image

Could you please provide the source of this claim? I didn't find it in Oracle's document. 

Actually I cann't understand Oracle's behavior of 'mn' match_param when source_char contains multi-line content. After some digging work on internet, I found this may be a bug of Oracle. Please see:
https://stackoverflow.com/questions/16094798/unbalanced-and-anchors-in-oracle-regexp 

I tested the following sql in Oracle 23: 
select regexp_count(chr(10)||'A'||chr(10)||'B', '^', 1, 'mn') from dual;   -- got 1
select regexp_count(chr(10)||'A'||chr(10)||'B', '$', 1, 'mn') from dual;   -- got 3

So I suggest to remove all content of 'mn' match_param, and withdraw the latest commit if we agree this is a Oracle's bug.

ai-yang: Thanks for checking this against Oracle 23 and for questioning the source. You are right: the documentation does not support the last-option-wins claim. Fixed in `fa3496f614`: I removed the combined-option assertions and tests, restored `m` and `n` as independent controls for their documented single-option behavior, and updated the PR description. Validation passed: targeted test 1/1, full extension 24/24, core regex 2/2, PostgreSQL 244/244, and Oracle 254/254.

PR #1490, opened 2026-08-01
raw_parser: Fix incorrect error hint referencing oracle parser

### LINKED ISSUE #1482: Code Quality: Misleading brach - `if (sql_raw_parser == NULL)`
A misleading error message. sql_raw_parser is statically initialized to standard_raw_parser and is never reset, so this NULL check can never be true — the ereport branch is ALMOST unreachable. On top of that, its error message wrongly refers to liboracle_parser/the oracle parser, which doesn't match the variable actually being checked (sql_raw_parser).

`src/backend/parser/parser.c`:

```c
/* Hook for plugins to get control in raw_parser() */
raw_parser_hook_type sql_raw_parser= standard_raw_parser;
raw_parser_hook_type ora_raw_parser= NULL;

List *
raw_parser(const char *str, RawParseMode mode)
{
  // !! this is misleading and unecessary !!
	if (sql_raw_parser == NULL)
		ereport(ERROR,
				(errcode(ERRCODE_SYSTEM_ERROR),
				 errmsg("liboracle_parser not found!"),
				 errhint("You must load liboracle_parser to use oracle parser.")));

	return (*sql_raw_parser)(str, mode);
}
```

digoal:
fix #1482 

  sql_raw_parser is statically initialized to standard_raw_parser and is
  never reset to NULL anywhere in the codebase, so the NULL check in
  raw_parser() can never be true — the ereport branch was unreachable.

  On top of that, its error message wrongly referred to liboracle_parser /
  the oracle parser, which did not match the variable actually being
  checked (sql_raw_parser). Since the hook is always non-NULL (initialized
  at file scope, and plugin assignments replace, not nullify it), simply
  calling through the function pointer is the correct behavior.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the parser error message when its parsing hook is not initialized, providing clearer guidance to resolve the issue.

* **Chores**
  * Added ignore rules to prevent generated code-graph files from being tracked.

### DIFF
--- a/.codegraph/.gitignore
+++ b/.codegraph/.gitignore
@@ -1,0 +1,5 @@
+# CodeGraph data files — local to each machine, not for committing.
+# Ignore everything in .codegraph/ except this file itself, so transient
+# files (the database, daemon.pid, sockets, logs) never show up in git.
+*
+!.gitignore
--- a/.codegraph/.gitignore
+++ b/.codegraph/.gitignore
@@ -1,5 +1,0 @@
-# CodeGraph data files — local to each machine, not for committing.
-# Ignore everything in .codegraph/ except this file itself, so transient
-# files (the database, daemon.pid, sockets, logs) never show up in git.
-*
-!.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -33,7 +33,7 @@ lib*dll.def
 lib*.pc
 
 # add by ai
-/.codegraph/
+.codegraph/
 
 # Local excludes in root directory
 /GNUmakefile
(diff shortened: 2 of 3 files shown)


bigplaice:
Thanks for contribution. We'll schedule a review for the patch.

bigplaice:
@digoal Hi, please go to the issue page (https://github.com/IvorySQL/IvorySQL/issues/1482) and leave '/assign' to take the task.

bigplaice:
@digoal 
I don't agree that the check against sql_raw_parser is dead code. Instead it is a mandatory defensive check. 
Please keep it and reword the error message and hint. 

BTW, I don't understand why /.codegraph/ is added into .gitignore file.

bigplaice:
@digoal 
Please remove .codegraph directory as it is not related to IvorySQL.

PR #1492, opened 2026-08-02
Remove unreachable compatible_mode dead code in guc_funcs.c

goeason-world:
## Summary

Fixes #1479.

Remove three unreachable `if (pg_strcasecmp(stmt->name, "compatible_mode") ...)` blocks in `ExecSetVariableStmt()`.

The GUC is registered as `ivorysql.compatible_mode` (see `ivy_guc.c`), so comparisons against the bare name `"compatible_mode"` never match. The real behavior already lives in:

- `check_compatible_mode()` — reject changes in native PG mode / validate parser load
- `assign_compatible_mode()` — switch `sql_raw_parser` and merge hooks

Also drop includes that were only needed by this dead code (`executor/nodeModifyTable.h`, `parser/parse_merge.h`, `parser/parser.h`).

## Test plan

- [ ] Confirm `SET ivorysql.compatible_mode = oracle|pg` still works in Oracle-mode database via GUC hooks
- [ ] Confirm `SET ivorysql.compatible_mode = oracle` is still rejected in native PG mode
- [ ] Confirm `RESET ivorysql.compatible_mode` still restores parser/hooks via `assign_compatible_mode`
- [ ] Build succeeds (no missing symbols from removed includes)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved consistency when setting or resetting the `compatible_mode` configuration option.
  - Configuration changes now follow the standard handling path, reducing unexpected behavior when switching compatibility settings.
  - Removed obsolete processing that could affect SQL and MERGE statement interpretation during configuration updates.

PR #1494, opened 2026-08-02
Delete useless code.

### LINKED ISSUE #1479: Chore: All three if (0 == pg_strcasecmp(stmt->name, "compatible_mode") ...){} in guc_funcs.c are dead code
```c
if (0 == pg_strcasecmp(stmt->name, "compatible_mode")
				&& DB_PG == database_mode)
{
	// will never reach here
}
```

quentinliu:
Fixes: #1479 

Removes three unreachable code blocks in ExecSetVariableStmt() (src/backend/utils/misc/guc_funcs.c) that attempt to intercept SET ivorysql.compatible_mode. The blocks never match because they compare against the bare name "compatible_mode", while the GUC is registered under its fully qualified name "ivorysql.compatible_mode".

## Summary by CodeRabbit

* **Bug Fixes**
  * Simplified configuration reset behavior by applying requested settings directly.
  * Removed mode-specific restrictions and automatic parser or merge-hook switching when changing or resetting configuration values.

### DIFF
--- a/src/backend/utils/misc/guc_funcs.c
+++ b/src/backend/utils/misc/guc_funcs.c
@@ -87,42 +87,11 @@ ExecSetVariableStmt(VariableSetStmt *stmt, bool isTopLevel)
 									errmsg("parameter \"%s\" cannot be changed", DB_MODE_PARMATER)));
 			}
 
-			if (0 == pg_strcasecmp(stmt->name, "compatible_mode")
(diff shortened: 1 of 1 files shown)


bigplaice:
Hi, your patch is almost same with PR #1492 but since first comer is 1492, we will take that one. 
Don't worry, you will still be considered as our contributor even the PR is not merged.

PR #1504, opened 2026-08-04
Add PARALLEL SAFE labeling to sys.* builtin functions

### LINKED ISSUE #1502: Missing PARALLEL labeling on sys.* builtin functions disables parallel query in Oracle compatibility mode
The CREATE FUNCTION statements in contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql mostly omit the PARALLEL clause, which defaults to PARALLEL UNSAFE. The planner then rejects parallelism for the entire statement if any unsafe function appears anywhere in the query — not just the part using it.
Current sys schema distribution: s=670, u=176, r=0 (PG core is 97% safe). 92 of 180 definitions in builtin_functions--1.0.sql lack the clause, including to_char, to_date, to_number, instr, substrb, lpad, rpad, trim, length, sysdate — functions almost every Oracle-mode query touches.

xuhuicai:
sys.* built-in functions defined in builtin_functions--1.0.sql were missing the PARALLEL clause, defaulting to PARALLEL UNSAFE. Because the planner disables parallelism for the whole statement when any unsafe function appears, Oracle compatibility mode effectively lost parallel query for common expressions (to_char, to_date, instr, lpad, etc.).

Add PARALLEL SAFE to 81 function definitions in
builtin_functions--1.0.sql, each backed by a matching PostgreSQL core precedent. sys.decompose already carried PARALLEL SAFE. Leave the 10 explicitly deferred definitions (get_sid/get_sessionid/sys_context and friends, candidates for RESTRICTED) untouched, and change no provolatile values.

Add a regression test in select_parallel.sql asserting that a query using sys.to_char still produces a Gather / Parallel Seq Scan plan.

fix #1502 

## Performance
- Updated built-in functions to support parallel query execution where appropriate.
- Parallel plans can now continue when using character, regular-expression, numeric, date/time, conversion, raw/blob/clob, and miscellaneous functions.
- Improved consistency for `sys.instr` function variants, including calls with explicit `varchar2` arguments.

## Tests
- Added regression coverage verifying parallel execution for aggregate queries using built-in function filters.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
+++ b/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
@@ -133,7 +133,9 @@ IMMUTABLE;
 CREATE FUNCTION sys.length(integer)
 RETURNS integer
 AS $$SELECT sys.length(cast($1 as sys.oravarcharchar));$$
-LANGUAGE SQL VOLATILE;
+LANGUAGE SQL
+PARALLEL SAFE
+VOLATILE;
 
 create function sys.lengthb(bytea) returns int as
 $$
@@ -141,7 +143,8 @@ begin
   return octet_length($1);
 end;
 $$
-language plpgsql;
+language plpgsql
+PARALLEL SAFE;
 
 
 /* trim/ltrim/rtrim functions */
@@ -182,6 +185,7 @@ RETURNS oravarcharchar
 AS 'MODULE_PATHNAME','trim1'
 LANGUAGE C
 STRICT
+PARALLEL SAFE
 IMMUTABLE;
 
 CREATE FUNCTION sys.trim(sys.oravarcharchar, sys.oravarcharchar)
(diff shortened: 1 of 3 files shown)


bigplaice:
Thanks for patching this. Overcall is looks good. I will check the modified functions one by one.

hanjianqiao:
@xuhuicai Thanks for the PR! One suggestion on the tests: could you add a few more test cases —— at least one per function category? Right now only `sys.to_char` is covered there; it'd give much more confidence in the PARALLEL SAFE labeling if each category had at least one representative test alongside it.

xuhuicai:
@hanjianqiao  Thanks! I've added one representative test per category. Updated in the latest commit, please take a look.

hanjianqiao:
Hi @xuhuicai. The expected plan at src/oracle_test/regress/expected/select_parallel.out Line 1497 casts both arguments to text. This query therefore tests sys.instr(text, text), not the changed sys.instr(varchar2, varchar2, ...) SQL wrappers. Please add a parallel-plan check with explicit varchar2 arguments.

xuhuicai:
Hi @hanjianqiao , I added a parallel-plan check with explicit varchar2 arguments.

Issue #1514, opened 2026-08-05
Build warning: -Wswitch in test_ddl_deparse for six IvorySQL-specific AT_ subcommands


xuhuicai:
## Bug Report

Building src/test/modules/test_ddl_deparse emits -Wswitch warnings. The switch (subcmd->subtype) in get_altertable_subcmdinfo() (src/test/modules/test_ddl_deparse/test_ddl_deparse.c:112) does not handle the six AlterTableType enum values that IvorySQL added in src/include/nodes/parsenodes.h:
AT_AddRowids,           /* SET WITH ROWID        */
AT_AddRowidsRecurse,    /* internal recursion     */
AT_DropRowids,          /* SET WITHOUT ROWID      */
AT_DropInvisible,       /* ALTER COLUMN DROP INVISIBLE */
AT_SetInvisible,        /* ALTER COLUMN SET INVISIBLE  */
AT_ForceViewCompile,    /* COMPILE FORCE VIEW     */
Since the switch has no default: branch, these cases are also reported as "unrecognized" instead of the actual command name when those DDLs trigger the module's event trigger.

### Expected behavior/code

The module builds without -Wswitch warnings, and the event trigger reports the correct command names for these subcommands.

### Step to reproduce

make -C src/test/modules/test_ddl_deparse

bigplaice:
Thanks for reporting.

btlqql:
/assign

I would like to add the missing IvorySQL AlterTableType mappings and regression coverage for their deparse command labels.

btlqql:
Update: current master already contains the six mappings in commits 76fcc56758 and 6ed09fb1ad, so no additional PR is needed from me. I am withdrawing the assignment request to avoid duplicating an existing fix.

PR #1515, opened 2026-08-05
Handle IvorySQL-specific AT_ subcommands in test_ddl_deparse

xuhuicai:
Add the missing cases after AT_ReAddStatistics.

About #1514 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `ALTER TABLE` command recognition for six additional operations.
  * Added clearer descriptions for row-ID changes, column visibility updates, and forced view compilation in generated output.
  * These operations now appear more accurately in deparsed SQL and related diagnostics.

### DIFF
--- a/src/test/modules/test_ddl_deparse/test_ddl_deparse.c
+++ b/src/test/modules/test_ddl_deparse/test_ddl_deparse.c
@@ -315,6 +315,10 @@ get_altertable_subcmdinfo(PG_FUNCTION_ARGS)
 			case AT_ReAddStatistics:
 				strtype = "(re) ADD STATS";
 				break;
+				/*
+				 * These enum values are for Oracle-compatible features and would
+				 * not be hit in PG mode, but put them here to avoid build warning.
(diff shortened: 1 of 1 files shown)


bigplaice:
These enum values are for Oracle compatible features and would not be really triggered in src/test/modules/test_ddl_deparse/test_ddl_deparse.c.  I agree to put them there to avoid build warnings but better to surround them with comment like "These enum valuse would not be hit in PG mode, but put them here to avoid build warning".

xuhuicai:
> These enum values are for Oracle compatible features and would not be really triggered in src/test/modules/test_ddl_deparse/test_ddl_deparse.c. I agree to put them there to avoid build warnings but better to surround them with comment like "These enum valuse would not be hit in PG mode, but put them here to avoid build warning".

Thanks for the suggestion. I've added the comment as suggested. Please review again.

hanjianqiao:
LGTM

PR #1516, opened 2026-08-06
[Oracle Compatibel]: Add ivorysql.dbtimezone GUC and dbtimezone() function

### LINKED ISSUE #1517: oracle compatibility feature: dbtimezone function
## Feature Request
Support Oracle compatibel function dbtimezone()

bigplaice:
Introduce Oracle-compatible DBTIMEZONE support: a new ivorysql.dbtimezone
  GUC backs the sys.dbtimezone() SQL function, which reports the database's
  configured time zone offset independently of the session's TimeZone
  setting (unlike sessiontimezone()).

  Like Oracle's DBTIMEZONE, this value is fixed per database rather than
  changeable ad hoc within a session:

  - The GUC accepts a UTC offset in the form [+-]HH:MI within Oracle's DBTIMEZONE range (-12:59 to +14:00), or a valid time zone region name.
  - A check_hook rejects plain SET, ALTER ROLE ... SET, and client-supplied startup values, only allowing ALTER DATABASE ... SET (plus the administrative sources used to establish the cluster-wide default).
  - The GUC's context is PGC_SUSET, so changing it normally requires superuser; a database owner can be authorized to manage it for their own database(s) via GRANT SET ON PARAMETER ivorysql.dbtimezone.

  Add regression coverage for: the plain-SET rejection, ALTER DATABASE ...
  SET with valid/invalid offsets and zone names, and a non-superuser role
  successfully changing DBTIMEZONE for a database it owns once granted
  SET ON PARAMETER permission.

  Assisted-by: Claude Code: Sonnet 5
  100% AI-generated code and commit message.

  The code and test cases have been audited by human author.

## New Features
- Added `sys.dbtimezone()` to report the configured database time zone independently of the session time zone.
- Added database-level time zone configuration using valid UTC offsets or time-zone region names.
- Added validation for supported values and required administrative permissions.
- Session-level and role-level changes are rejected; database owners can configure the setting when granted permission.

## Tests
- Added regression coverage for defaults, persistence, session independence, invalid values, permissions, and cleanup.

fix #1517


contrib/ivorysql_ora/sql/ora_datetime_datatype_functions.sql
```diff
@@ -315,6 +315,55 @@ SELECT FROM_TZ(TIMESTAMP '2000-03-28 08:00:00', '3:00') FROM DUAL;
 --sessiontimezone
 set timezone = 'Asia/Hong_Kong';
 select sessiontimezone() from dual; /* BUG:0000427 */
+
```
bigplaice, line 318: Add test case to verify the sessiontimezone not conflict with dbtimezone

Issue #1102, opened 2025-12-05
Ecosystem Integration: IvorySQL v5 + passwordcheck


hs-liuxh:
## Feature Request
Ecosystem Integration: IvorySQL v5 + passwordcheck
**Describe the feature you'd like to propose:**
Enforce password complexity policies (e.g., minimum length, mixed case, digits, special characters) during user password creation or modification to prevent weak passwords.

messere1:
/assign

I would like to complete the passwordcheck integration for IvorySQL 5.x. I will validate the bundled contrib module build and preload configuration, password acceptance and rejection paths, encrypted-password behavior and limitations, regression tests, and both PG and Oracle compatibility modes. I will provide reproducible English and Chinese documentation and update the ecosystem overview.

PR #1520, opened 2026-08-08
Add configurable password complexity checks

btlqql:
## What

- Add configurable plaintext password checks for uppercase letters, lowercase
  letters, digits, and punctuation.
- Enable all four complexity categories by default while allowing superusers
  to disable individual rules with `passwordcheck.require_*` GUCs.
- Extend both PostgreSQL and Oracle-compatible regression expectations.
- Document the settings and the limitation for pre-encrypted passwords.

## Why

The existing `passwordcheck` integration only enforces minimum length and a
letter/non-letter mix. It cannot express the mixed-case, digit, and special
character policy requested in #1102.

## Impact

The module remains opt-in through `shared_preload_libraries`. Once loaded,
plaintext passwords must satisfy the four category checks by default. Existing
deployments can retain a less strict policy by changing the new SUSET options.
Pre-encrypted password behavior is unchanged.

## Root cause

`check_password()` tracked only `isalpha()` versus non-alpha characters and
exposed only `passwordcheck.min_password_length`.

## Checks

- `git diff --check`
- Regression SQL echoed by both expected-output files was compared line by
  line.
- New success, failure, and configuration-override regression cases added.
- Linux PR CI passed: policy, build, Meson, contrib, PostgreSQL, and both
  Oracle-compatible regression suites.

Refs #1102

Assisted-by: OpenAI:GPT-5

Per `coding-assistants.md`, this commit intentionally has no AI-added
`Signed-off-by`; the human contributor must review and certify it before merge.

## Summary by CodeRabbit

* **New Features**
  * Added configurable password complexity requirements for uppercase letters, lowercase letters, digits, and special characters.
  * All complexity requirements are enabled by default and can be adjusted by superusers.
  * Password validation now reports which required character category is missing.

* **Documentation**
  * Documented password complexity settings and their behavior for plaintext and encrypted passwords.

### DIFF
--- a/contrib/passwordcheck/expected/passwordcheck.out
+++ b/contrib/passwordcheck/expected/passwordcheck.out
@@ -2,20 +2,36 @@ SET md5_password_warnings = off;
 LOAD 'passwordcheck';
 CREATE USER regress_passwordcheck_user1;
 -- ok
-ALTER USER regress_passwordcheck_user1 PASSWORD 'a_nice_long_password';
+ALTER USER regress_passwordcheck_user1 PASSWORD 'A_nice_long_password1';
 -- error: too short
 ALTER USER regress_passwordcheck_user1 PASSWORD 'tooshrt';
 ERROR:  password is too short
 DETAIL:  password must be at least "passwordcheck.min_password_length" (8) bytes long
 -- ok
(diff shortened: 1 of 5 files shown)


btlqql:
Superseded by a pull request from the renamed source branch.

PR #1526, opened 2026-08-09
fix(datetime): zero fsec in ora_round to match ora_trunc behavior (1524)

### LINKED ISSUE #1524: ora_round does not zero fractional seconds (fsec), inconsistent with ora_trunc
### IvorySQL Version

### OS Version (uname -a)

Ubuntu 20.04, gcc 11.5.0

```
meson setup build --prefix=$HOME/pginstall -Dcassert=true
initdb -D /tmp/ivorysql-test -U ivorysql
```

### Current Behavior

The `ora_round()` function in `contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c` does not zero `fsec` (fractional seconds) after calling `tm_round()`. This means fractional seconds from the original timestamp are preserved in the rounded result, which is incorrect.

In contrast, `ora_trunc()` explicitly sets `fsec = 0;` after calling `tm_trunc()`:

```c
// ora_trunc (CORRECT — line ~501):
tm_trunc(tm, fmt);
fsec = 0;                    // zeros fractional seconds
if (tm2timestamp(tm, fsec, NULL, &result) != 0)
    ...

// ora_round (BUG — line ~462):
tm_round(tm, fmt);
                               // fsec is NOT zeroed!
if (tm2timestamp(tm, fsec, NULL, &result) != 0)
    ...
```

The `tm_round()` function zeroes `tm->tm_sec` (and `tm->tm_min`, `tm->tm_hour` depending on format), but it never touches `fsec`. So `fsec` retains whatever value was extracted by `timestamp2tm()` and is passed back to `tm2timestamp()` unchanged.

### Expected behavior/code

In Oracle, `ROUND` on a date/timestamp should eliminate the time component (including fractional seconds) when rounding to day or coarser granularity. Even when rounding to minute (`MI`), the seconds and fractional seconds should be zeroed.

The `ora_round()` function should set `fsec = 0;` after `tm_round()`, matching the behavior of `ora_trunc()`.

### Step to reproduce

Currently, the bug is latent because `sys.round(date, text)` only accepts PG `date` type (which has no fractional seconds). However:

1. The C function `ora_round()` takes `PG_GETARG_TIMESTAMP(0)` — it is designed to work with timestamps that have fractional seconds.
2. If a `sys.round(timestamp, text)` overload is added (which is a natural extension), the bug would immediately manifest.
3. The internal `tm_round()` function already handles `tm->tm_sec`, so the omission of `fsec` zeroing is clearly an oversight, not intentional design.

To verify the code inconsistency:

```bash
# ora_trunc correctly zeros fsec (line ~501):
sed -n '498,505p' contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c

# ora_round does NOT zero fsec (line ~462):
sed -n '459,467p' contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c

# tm_round zeroes tm->tm_sec but never touches fsec (line ~1930):
sed -n '1925,1935p' contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
```

**Affected file:** `contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c`

| Function | Line | fsec handling |
|----------|------|---------------|
| `ora_round()` | ~439-467 | `fsec` NOT zeroed after `tm_round()` |
| `ora_trunc()` | ~473-503 | `fsec = 0;` after `tm_trunc()` (correct) |
| `tm_round()` | ~1850-1935 | Zeroes `tm->tm_sec` but not `fsec` |

**Fix:** Add `fsec = 0;` after `tm_round(tm, fmt);` in `ora_round()`:

```c
// Before:
tm_round(tm, fmt);

if (tm2timestamp(tm, fsec, NULL, &result) != 0)

// After:
tm_round(tm, fmt);
fsec = 0;

if (tm2timestamp(tm, fsec, NULL, &result) != 0)
```

This makes `ora_round` consistent with `ora_trunc`, ensuring that rounding eliminates fractional seconds just as truncation does.

tju-yxq:
## Summary

Add `fsec = 0;` after `tm_round()` in `ora_round()` to zero fractional seconds, matching the behavior of `ora_trunc()`.

## Background / Motivation

`ora_round()` and `ora_trunc()` are Oracle-compatible date/time rounding and truncation functions in `contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c`. Both extract `fsec` (fractional seconds) from the input timestamp via `timestamp2tm()`, then call their respective `tm_*` helper to modify the `tm` struct.

However, `ora_trunc()` explicitly zeros `fsec` after `tm_trunc()`:

```c
// ora_trunc (correct):
tm_trunc(tm, fmt);
fsec = 0;                    // zeros fractional seconds
if (tm2timestamp(tm, fsec, NULL, &result) != 0)
```

While `ora_round()` does not:

```c
// ora_round (bug):
tm_round(tm, fmt);
                               // fsec NOT zeroed!
if (tm2timestamp(tm, fsec, NULL, &result) != 0)
```

The `tm_round()` function zeroes `tm->tm_sec` but never touches `fsec`. This means fractional seconds from the original timestamp are preserved in the rounded result, which is incorrect — Oracle's `ROUND` should eliminate fractional seconds.

The bug is currently latent because the SQL overload `sys.round(date, text)` only accepts PG `date` type (no fractional seconds). But the C function uses `PG_GETARG_TIMESTAMP(0)` and is designed to handle timestamps with fractional seconds. Adding a `timestamp` overload would immediately expose the bug.

Fixes #1524

## Changes

- `contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c`: Added `fsec = 0;` after `tm_round(tm, fmt);` in `ora_round()` (line ~464)

## How to Verify

1. Build with the fix: `ninja -C build -j20 && ninja -C build install`
2. Existing ROUND regression tests should still pass:
   ```
   cd build && meson test
   ```
3. Verify code consistency between `ora_round` and `ora_trunc`:
   ```c
   // Both should now have fsec = 0 after their tm_* call:
   // ora_round:  tm_round(tm, fmt);  fsec = 0;
   // ora_trunc:  tm_trunc(tm, fmt);  fsec = 0;
   ```

## Compliance

- [x] Regression tests pass

Assisted-by: OpenAI:gpt-5
Percentage of AI-generated code: 80%

## Summary by CodeRabbit

* **Bug Fixes**
  * Date rounding now consistently removes fractional seconds, producing timestamps with zero microseconds after rounding.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
@@ -461,6 +461,8 @@ ora_round(PG_FUNCTION_ARGS)
 
 	tm_round(tm, fmt);
 
+	fsec = 0;
+
 	if (tm2timestamp(tm, fsec, NULL, &result) != 0)
 		ereport(ERROR,
 				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),


jiaoshuntian:
thanks

PR #1532, opened 2026-08-09
fix(datetime): handle empty NLS parameters (1531)

### LINKED ISSUE #1531: Three-argument datetime conversions use uninitialized state for empty NLS parameter
### IvorySQL Version

Current `master` (`d4661fb1d8971cf27415e17361654431ba7b716e`, IvorySQL 5beta1).

### OS Version (uname -a)

macOS arm64; reproduced with an assertion-enabled Clang build.

### Configuration options

```
--enable-cassert --enable-debug --without-icu --without-readline --without-zlib --with-uuid=e2fs
```

### Current Behavior

The three-argument implementations of `TO_DATE`, `TO_TIMESTAMP`, and `TO_TIMESTAMP_TZ` only call `ora_do_to_timestamp()` when `nlsparam` is non-empty. A non-NULL empty text value therefore leaves `struct pg_tm` and `fsec` uninitialized before they are consumed by `tm2timestamp()` and timezone handling.

For example, `repeat('x', 0)` produces a non-NULL empty text value. On the current build:

```sql
SELECT length(repeat('x', 0)), repeat('x', 0) IS NULL;
-- 0 | false

SELECT sys.to_date('2024-02-03', 'YYYY-MM-DD', repeat('x', 0));
-- 0056-07-25 (result varies with stack contents)

SELECT sys.to_timestamp_tz(
  '2024-02-03 04:05:06 +08:00',
  'YYYY-MM-DD HH24:MI:SS TZH:TZM',
  repeat('x', 0));
-- unrelated date/time values (result varies with stack contents)
```

### Expected behavior/code

An empty NLS parameter must not skip datetime parsing. Since the implementation currently ignores the contents of `nlsparam`, empty and non-empty values should both parse `date_txt` using `fmt`, without reading uninitialized memory.

### Step to reproduce

Run the queries above against an Oracle-compatible cluster. The same defect exists in:

- `to_oradate3`
- `to_oratimestamp3`
- `to_oratimestamptz3`

### Additional context

The minimal fix is to invoke `ora_do_to_timestamp()` unconditionally in all three functions and remove the unused `text_to_cstring()`/`strlen()` guards. Regression coverage should use a generated non-NULL empty text value so the SQL `STRICT` attribute does not short-circuit the C function.

yyqdbngt:
## Summary

- always parse input in the three-argument `to_date`, `to_timestamp`, and `to_timestamp_tz` implementations
- avoid reading uninitialized `pg_tm`/`fsec` state when `nlsparam` is an empty, non-null string
- add regression coverage for all three conversion functions with a runtime-generated empty NLS parameter

## Root cause

The conversion call was guarded by `strlen(nlsparam) != 0`. An empty `nlsparam` therefore skipped `ora_do_to_timestamp()`, while the function continued to build a result from uninitialized parse state.

## Tests

- `make -C contrib/ivorysql_ora oracle-check ORA_REGRESS=ora_datetime_datatype_functions` (pass)
- full `oracle-check`: 23/24 pass; `ora_xml_functions` requires a build configured with `--with-libxml`

Fixes #1531

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Oracle-compatible `to_date`, `to_timestamp`, and `to_timestamp_tz` conversions when an empty NLS parameter is supplied.
  * Ensured timestamp-with-time-zone conversions correctly handle UTC output and source offsets.
  * Added regression coverage for date, timestamp, and time-zone-aware timestamp formatting scenarios.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
@@ -26,6 +26,14 @@ SELECT to_date(
  01-15-1989 11:00:00
 (1 row)
 
+SELECT to_char(
+	to_date('2001-02-03', 'YYYY-MM-DD', repeat('x', 0)),
+	'YYYY-MM-DD HH24:MI:SS') AS empty_nls_to_date FROM DUAL;
+  empty_nls_to_date  
+---------------------
+ 2001-02-03 00:00:00
+(1 row)
+
 /*
  * to_timestamp
  */
@@ -51,6 +59,15 @@ SELECT to_timestamp(
          to_timestamp          
 -------------------------------
  09-10-2016 14:10:10.123000000
+(1 row)
+
+SELECT to_char(
+	to_timestamp('2002-03-04 05:06:07.123',
+		'YYYY-MM-DD HH24:MI:SS.FF3', repeat('x', 0)),
+	'YYYY-MM-DD HH24:MI:SS.FF3') AS empty_nls_to_timestamp FROM DUAL;
+ empty_nls_to_timestamp  
+-------------------------
+ 2002-03-04 05:06:07.123
 (1 row)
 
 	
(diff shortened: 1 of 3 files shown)


jiaoshuntian:
LGTM

yyqdbngt:
Thanks for the LGTM @jiaoshuntian. CI (build, pg_regression, oracle_regression, oracle_pg_regression, contrib_regression) is all green.

PR #1536, opened 2026-08-10
fix(parser): allow MERGE DELETE without WHERE clause for Oracle compatibility (1534)

### LINKED ISSUE #1534: MERGE DELETE without WHERE clause fails with syntax error in Oracle compatibility mode
C:\Users\26517\AppData\Local\Temp\issue1534_update.txt

tju-yxq:
## Summary

Allow MERGE `DELETE` without `WHERE` clause in Oracle compatibility mode, matching Oracle behavior where `DELETE` without `WHERE` deletes all matched and updated rows.

## Background / Motivation

In Oracle, the MERGE statement's DELETE clause can be used with or without a WHERE clause:

```sql
-- Delete only rows matching the condition
WHEN MATCHED THEN UPDATE SET col = val DELETE WHERE condition

-- Delete ALL matched and updated rows
WHEN MATCHED THEN UPDATE SET col = val DELETE
```

IvorySQL's `merge_delete` grammar rule in `src/backend/oracle_parser/ora_gram.y` only supported `DELETE_P where_clause`. When `DELETE` was used without `WHERE`, it raised `ERROR: missing WHERE keyword`. This is a significant Oracle compatibility gap that affects any MERGE statement migrated from Oracle that uses the `DELETE` without `WHERE` syntax.

Fixes #1534

## Changes

- `src/backend/oracle_parser/ora_gram.y`: Added a new grammar production `DELETE_P` (without `where_clause`) to the `merge_delete` rule. When no `WHERE` is provided, a `MergeWhenClause` node is created with `condition = NULL`. The executor handles this correctly — `ExecQual()` returns `true` when the qual state is `NULL`, so all matched and updated rows are deleted.

- `src/oracle_test/regress/sql/merge.sql`: Added regression tests covering:
  - `DELETE` without `WHERE` (deletes all matched and updated rows)
  - `DELETE` with `WHERE` (deletes only matching rows) for comparison

- `src/oracle_test/regress/expected/merge.out`: Added expected output for the new tests.

## How to Verify

```sql
CREATE TABLE merge_t (id NUMBER, val VARCHAR2);
CREATE TABLE merge_s (id NUMBER, val VARCHAR2);
INSERT INTO merge_t VALUES (1, 'old1');
INSERT INTO merge_t VALUES (2, 'old2');
INSERT INTO merge_s VALUES (1, 'new1');

-- DELETE without WHERE - deletes all matched rows
MERGE INTO merge_t t USING merge_s s ON (t.id = s.id)
WHEN MATCHED THEN UPDATE SET t.val = s.val DELETE;

SELECT * FROM merge_t ORDER BY id;
-- Expected: only row 2 remains (row 1 was matched and deleted)
```

## Compliance

- [x] Regression tests added

Assisted-by: OpenAI:gpt-5
Percentage of AI-generated code: 70%

## Summary by CodeRabbit

* **New Features**
  * `MERGE` statements now support unconditional `DELETE` actions without a `WHERE` clause.
  * Unconditional deletion applies to all rows matched and updated by the merge.

* **Bug Fixes**
  * Preserved validation for incomplete conditional deletes using `DELETE WHERE` without a condition.

* **Tests**
  * Added regression coverage for unconditional and conditional `MERGE` deletes and their resulting data.


jiaoshuntian:
Please first supplement the content of the issue

tju-yxq:
Thank you for the feedback. I have supplemented the content of issue #1534 and removed the accidental UTF-8 BOM from the regression test files. Please take another look when you have time.

jiaoshuntian:
The  issue  #1534 itself has problems.

tju-yxq:
Thank you for the feedback. I built IvorySQL from source and tested the MERGE DELETE syntax. Here are the results:

1. Standalone DELETE works:
`
MERGE INTO t1 USING t2 ON (t1.id = t2.id) WHEN MATCHED THEN DELETE;
-- Works correctly
`

2. UPDATE SET ... DELETE (with or without WHERE) does NOT work. This is a pre-existing issue on master:
`
MERGE INTO t1 USING t2 ON (t1.id = t2.id)
WHEN MATCHED THEN UPDATE SET t1.id = t2.id DELETE;
-- ERROR: syntax error at or near "DELETE"

MERGE INTO t1 USING t2 ON (t1.id = t2.id)
WHEN MATCHED THEN UPDATE SET t1.id = t2.id DELETE WHERE (t1.id = 1);
-- ERROR: syntax error at or near "DELETE"
`

Both fail on master (without our change). The root cause is that DELETE_P is an UNRESERVED_KEYWORD in IvorySQL's oracle parser, so the parser interprets DELETE after SET col = val as a column name (part of set_clause_list) rather than the DELETE keyword.

In Oracle, DELETE is a reserved keyword, so the parser can distinguish it from identifiers. In IvorySQL, making DELETE_P reserved could break existing queries that use delete as a column name.

This means the UPDATE SET ... DELETE syntax (both with and without WHERE) is a pre-existing limitation that requires a deeper grammar fix beyond what this PR attempts. I will close this PR and update the issue to reflect the full scope of the problem.

PR #1538, opened 2026-08-10
Install amcheck module for regression testing

### LINKED ISSUE #1533: make oracle-check command failed in alter_index
## Bug Report
When run "make oracle-check" on IvorySQL:master branch, there is failure:
# parallel group (2 tests):  create_index_online alter_index
not ok 252   + alter_index                               623 ms

### IvorySQL Version
master branch

### OS Version (uname -a)
N/A

N/A

### Step to reproduce
./configure
make
make oracle-checl

yasir-hussain-shah:
This PR adds `contrib/amcheck` to `EXTRA_INSTALL` to ensure the module is available during Oracle compatibility regression test runs. 

This is necessary because `alter_index.sql` exercises `amcheck`'s `bt_index_check()` function, which is not included in the default installation. By including it in `EXTRA_INSTALL`, we guarantee that the module is built and available in the temporary test environment.

fix #1533 

## Summary by CodeRabbit

* **Tests**
  * Enhanced Oracle regression test setup to include additional database integrity checks.

### DIFF
--- a/src/oracle_test/regress/GNUmakefile
+++ b/src/oracle_test/regress/GNUmakefile
@@ -95,6 +95,8 @@ installdirs-tests: installdirs
 ## Run tests
 ##
 
+EXTRA_INSTALL = contrib/amcheck
+
 ORACLE_REGRESS_OPTS = --dlpath=. --max-concurrent-tests=20 \
 	$(EXTRA_REGRESS_OPTS)
 


src/oracle_test/regress/GNUmakefile
```diff
@@ -95,6 +95,8 @@ installdirs-tests: installdirs
 ## Run tests
 ##
```
bigplaice, line 97: It would be good to add some comment to explain why extra install is needed:
    # alter_index.sql exercises amcheck's bt_index_check(), which is not part
    # of the default install; make sure it's built into the temp install.

yasir-hussain-shah: @bigplaice 
Description updated.

bigplaice:
LGTM. Thanks for reporting this problem and fix of it!

PR #1542, opened 2026-08-10
fix(libpq): include sys/ipc.h and sys/sem.h unconditionally in sysv_ivysema.c

### LINKED ISSUE #1540: sysv_ivysema.c header guards (HAVE_SYS_SEM_H/HAVE_SYS_IPC_H) are never defined — breaks libpq build on System V sem platforms (macOS)
### Summary

`src/interfaces/libpq/sysv_ivysema.c` guards the inclusion of `<sys/ipc.h>` and `<sys/sem.h>` behind `#ifdef HAVE_SYS_IPC_H` / `#ifdef HAVE_SYS_SEM_H`, but IvorySQL's `configure` never defines either macro (there is no `AC_CHECK_HEADERS` for `sys/sem.h` / `sys/ipc.h`). As a result both guards are always false, the headers are never included, and the file fails to compile with a flood of "undeclared identifier" errors for `semctl`, `semget`, `IPC_RMID`, `IPC_CREAT`, `union semun`, etc.

This breaks the libpq build on any platform that selects the System V semaphore implementation (notably macOS).

### How to reproduce

On macOS (Apple clang 17 / macOS 26 SDK, but reproducible on any System V sem platform):

```sh
export PATH="/opt/homebrew/opt/bison/bin:/opt/homebrew/opt/flex/bin:$PATH"
./configure --prefix=$PWD/build_install --enable-debug --enable-cassert \
  --with-libxml --with-icu --with-ssl=openssl --with-readline
make -j
```

### Current (broken) behavior

```
ivy_sema.c:78:10: error: call to undeclared function 'semget'
ivy_sema.c:78:34: error: use of undeclared identifier 'IPC_CREAT'
ivy_sema.c:123:14: error: variable has incomplete type 'union semun'
ivy_sema.c:158:9: error: call to undeclared function 'semctl'
...
```

### Root cause

```c
#ifdef HAVE_SYS_IPC_H
#include <sys/ipc.h>
#endif

#ifdef HAVE_SYS_SEM_H
#include <sys/sem.h>
#endif
```

- `src/include/pg_config.h` does **not** define `HAVE_SYS_SEM_H` or `HAVE_SYS_IPC_H`.
- `configure` / `configure.ac` contains no check that would define them (`grep -c 'HAVE_SYS_SEM_H\|HAVE_SYS_IPC_H' configure` → 0).
- Upstream `src/backend/port/sysv_sema.c` includes `<sys/ipc.h>` and `<sys/sem.h>` **unconditionally** — it does not rely on these macros. The IvorySQL `sysv_ivysema.c` copy diverged by adding guards that nothing ever defines, so the includes are silently skipped.

This is an IvorySQL-only file (`sysv_ivysema.c` + `ivy_sema.c` symlink, added for OUT-parameter support in #837), so the bug does not exist in upstream PostgreSQL.

### Expected behavior / suggested fix

Match upstream `sysv_sema.c` and include the headers unconditionally (a System V semaphore build, by definition, has these headers):

```c
#include <sys/ipc.h>
#include <sys/sem.h>
```

Alternatively, add `AC_CHECK_HEADERS([sys/ipc.h sys/sem.h])` to `configure.ac` so the macros are actually defined — but since upstream simply includes them unconditionally, the unconditional include is the minimal fix.

### Environment

- IvorySQL: master @ d4661fb1d89 (5beta1 / PG 19devel)
- OS: macOS 26 (Darwin 25.2.0, arm64)
- Compiler: Apple clang 17.0.0
- configure: `--with-libxml --with-icu --with-ssl=openssl --with-readline` (System V semaphores auto-selected on macOS)

Aias00:
## Summary

`src/interfaces/libpq/sysv_ivysema.c` guarded `<sys/ipc.h>` and `<sys/sem.h>` behind `#ifdef HAVE_SYS_IPC_H` / `#ifdef HAVE_SYS_SEM_H`, but IvorySQL's `configure` never defines either macro (there is no `AC_CHECK_HEADERS` for these headers, and upstream `src/backend/port/sysv_sema.c` includes them unconditionally). The guards were therefore always false, the headers were never included, and the file failed to compile with a flood of undeclared-identifier errors — breaking the libpq build on any platform that selects System V semaphores (e.g. macOS).

This file is IvorySQL-only (`sysv_ivysema.c` + `ivy_sema.c` symlink, added for OUT-parameter support in #837), so the bug does not exist upstream.

## Root cause

```c
#ifdef HAVE_SYS_IPC_H
#include <sys/ipc.h>
#endif

#ifdef HAVE_SYS_SEM_H
#include <sys/sem.h>
#endif
```

- `src/include/pg_config.h` does **not** define `HAVE_SYS_SEM_H` or `HAVE_SYS_IPC_H`.
- `configure` / `configure.ac` contains no check that would define them (`grep -c 'HAVE_SYS_SEM_H\|HAVE_SYS_IPC_H' configure` → 0).
- Upstream `sysv_sema.c` includes the same headers **unconditionally**.

## Fix

Match upstream and include both headers unconditionally (a System V semaphore build, by definition, has these headers).

## Reproduction (before)

```sh
./configure --prefix=$PWD/build_install --with-libxml --with-icu --with-ssl=openssl --with-readline
make -j
```

```
ivy_sema.c:78:10: error: call to undeclared function 'semget'
ivy_sema.c:78:34: error: use of undeclared identifier 'IPC_CREAT'
ivy_sema.c:123:14: error: variable has incomplete type 'union semun'
ivy_sema.c:158:9: error: call to undeclared function 'semctl'
...
```

After the fix the file compiles and `make` / `make install` complete; `libpq.5.dylib` links cleanly (passes `libpq_check.pl`).

## Environment

- IvorySQL master @ d4661fb1d89 (5beta1 / PG 19devel)
- macOS 26 (Darwin 25.2.0, arm64), Apple clang 17.0.0

Fixes #1540

---

Assisted-by: Claude:Sonnet-4.6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility for semaphore-related functionality by ensuring required System V IPC support is consistently available.

### DIFF
--- a/src/interfaces/libpq/sysv_ivysema.c
+++ b/src/interfaces/libpq/sysv_ivysema.c
@@ -15,14 +15,8 @@
 #include <signal.h>
 #include <unistd.h>
 #include <sys/file.h>
-
-#ifdef HAVE_SYS_IPC_H
 #include <sys/ipc.h>
-#endif
-
-#ifdef HAVE_SYS_SEM_H
 #include <sys/sem.h>
-#endif
 
 #include "ivy_sema.h"
 


db-nikhil:
@Aias00 I already submitted a fix for this along with some other fixes here: https://github.com/IvorySQL/IvorySQL/pull/1475

PTAL. Thanks

PR #1545, opened 2026-08-11
Fix DBMS_OUTPUT.DISABLE buffer cleanup

### LINKED ISSUE #1544: DBMS_OUTPUT.DISABLE does not purge buffered output
### IvorySQL Version

### OS Version (uname -a)

macOS (the behavior is in the server-side `ivorysql_ora` extension and is not OS-specific)

Default local development build with Oracle compatibility enabled.

### Current Behavior

`DBMS_OUTPUT.DISABLE` only sets the package's `enabled` flag to `false`. Lines buffered before the call remain available to `DBMS_OUTPUT.GET_LINE` / `GET_LINES`.

For example, after buffering `before disable`, calling `DISABLE`, and then calling `GET_LINE`, IvorySQL returns the old line with status `0`.

### Expected behavior/code

`DBMS_OUTPUT.DISABLE` should purge the buffered output as part of disabling the package. A subsequent `GET_LINE` should return no line and status `1`.

This matches Oracle's documented `DBMS_OUTPUT.DISABLE` behavior: the procedure disables calls to `PUT`, `PUT_LINE`, `NEW_LINE`, `GET_LINE`, and `GET_LINES` and purges the buffer.

### Step to reproduce

```sql
CALL DBMS_OUTPUT.ENABLE(NULL);
CALL DBMS_OUTPUT.PUT_LINE('before disable');
CALL DBMS_OUTPUT.DISABLE();

DECLARE
  line varchar2(32767);
  status binary_integer;
BEGIN
  DBMS_OUTPUT.GET_LINE(line, status);
  DBMS_OUTPUT.PUT_LINE('line=' || coalesce(line, '<NULL>') || ', status=' || status);
END;
/
```

The underlying problem is visible in `ora_dbms_output_disable()`: it sets `output_buffer.enabled = false` but does not call the existing buffer cleanup routine.

The existing regression named "DISABLE clears existing buffer" calls `ENABLE` before reading. Since `ENABLE` itself clears the buffer, that test does not exercise the state immediately after `DISABLE` and masks this behavior.

Aias00:
## Summary

- purge the session buffer when `DBMS_OUTPUT.DISABLE` is called
- make the existing regression exercise the state immediately after `DISABLE`
- align the native implementation comment with Oracle's documented behavior

Fixes #1544

## Testing

- confirmed the regression scenario fails before the fix: `This will be cleared|0`
- compiled the changed `dbms_output.c` object against the current IvorySQL server headers
- loaded the fixed object into an isolated test module and confirmed the same scenario returns `<NULL>|1`
- `git diff --check`

## Reference

- Oracle DBMS_OUTPUT documentation: https://docs.oracle.com/en/database/oracle/oracle-database/19/arpls/DBMS_OUTPUT.html

## Summary by CodeRabbit

- **Bug Fixes**
  - Disabling `DBMS_OUTPUT` now clears any buffered output.
  - Subsequent buffer reads no longer return output that was written before disabling.

- **Tests**
  - Updated coverage verifies that disabling output immediately purges the buffer.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_dbms_output.out
+++ b/contrib/ivorysql_ora/expected/ora_dbms_output.out
@@ -170,7 +170,6 @@ BEGIN
     dbms_output.enable();
     dbms_output.put_line('This will be cleared');
     dbms_output.disable();
-    dbms_output.enable();
     dbms_output.get_line(line, status);
     RAISE NOTICE 'Test 3.2 - Buffer after disable: [%], Status: %', line, status;
 END;
--- a/contrib/ivorysql_ora/sql/ora_dbms_output.sql
+++ b/contrib/ivorysql_ora/sql/ora_dbms_output.sql
@@ -169,7 +169,6 @@ BEGIN
     dbms_output.enable();
     dbms_output.put_line('This will be cleared');
     dbms_output.disable();
-    dbms_output.enable();
     dbms_output.get_line(line, status);
     RAISE NOTICE 'Test 3.2 - Buffer after disable: [%], Status: %', line, status;
 END;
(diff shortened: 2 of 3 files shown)


jiaoshuntian:
LGTM

PR #1547, opened 2026-08-11
Discard unread DBMS_OUTPUT lines before new output

### LINKED ISSUE #1546: DBMS_OUTPUT keeps unread lines when output resumes after a read
### IvorySQL Version

### OS Version (uname -a)

macOS (the behavior is in the server-side `ivorysql_ora` extension and is not OS-specific)

Default local development build with Oracle compatibility enabled.

### Current Behavior

After a partial `DBMS_OUTPUT.GET_LINE` or `GET_LINES` read, IvorySQL keeps unread old lines when the next operation writes new output.

For a buffer containing `A` and `B`, reading only `A` and then writing `C` causes the next `GET_LINES` call to return `{B,C}`.

### Expected behavior/code

The next call to `PUT`, `PUT_LINE`, or `NEW_LINE` after a read should discard lines that were not retrieved. The example should therefore return only `{C}`.

Oracle documents this explicitly for both `GET_LINE` and `GET_LINES`: unread lines are discarded when the next output-writing procedure is called.

### Step to reproduce

```sql
DECLARE
  line varchar2(32767);
  status binary_integer;
  lines dbms_output.chararr;
  count binary_integer := 10;
BEGIN
  DBMS_OUTPUT.ENABLE(NULL);
  DBMS_OUTPUT.PUT_LINE('A');
  DBMS_OUTPUT.PUT_LINE('B');
  DBMS_OUTPUT.GET_LINE(line, status);
  DBMS_OUTPUT.PUT_LINE('C');
  DBMS_OUTPUT.GET_LINES(lines, count);

  -- IvorySQL currently returns B and C; Oracle-compatible behavior is C only.
END;
/
```

The current linked-list buffer immediately removes retrieved nodes, but it does not record that a read occurred. Consequently, the subsequent write has no way to purge the remaining old nodes.

Aias00:
## Summary

- remember when a `DBMS_OUTPUT` read has occurred
- discard completed unread lines at the next `PUT`, `PUT_LINE`, or `NEW_LINE`
- add regression coverage for a partial read followed by new output

Fixes #1546

## Testing

- confirmed the regression scenario fails before the fix with `{"Old 2","New 1"}`
- compiled the changed `dbms_output.c` object against the current IvorySQL server headers
- loaded the fixed object into an isolated test module and confirmed the same scenario returns `{"New 1"}`
- `git diff --check`

## Reference

- Oracle DBMS_OUTPUT documentation: https://docs.oracle.com/en/database/oracle/oracle-database/19/arpls/DBMS_OUTPUT.html

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated DBMS_OUTPUT behavior so writing after partially reading the buffer discards unread previous lines.
  * Ensured newly written output is available for subsequent retrieval without stale buffered lines.

* **Tests**
  * Added coverage validating buffer behavior after partial reads followed by new output.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_dbms_output.out
+++ b/contrib/ivorysql_ora/expected/ora_dbms_output.out
@@ -376,6 +376,27 @@ NOTICE:    Line 1: [One]
 NOTICE:    Line 2: [Two]
 NOTICE:  Test 6.3b - Remaining: [Three], Status: 0
 NOTICE:  Test 6.3c - Remaining: [Four], Status: 0
+-- Test 6.4: A write after GET_LINE discards unread lines
+DECLARE
+    lines TEXT[];
+    numlines INTEGER := 10;
+    line TEXT;
+    status INTEGER;
+BEGIN
+    dbms_output.enable();
+    dbms_output.put_line('Old 1');
+    dbms_output.put_line('Old 2');
+    dbms_output.get_line(line, status);
+    dbms_output.put_line('New 1');
+    dbms_output.get_lines(lines, numlines);
+    RAISE NOTICE 'Test 6.4 - Retrieved % lines after writing', numlines;
+    FOR i IN 1..numlines LOOP
+        RAISE NOTICE '  Line %: [%]', i, lines[i];
+    END LOOP;
+END;
+/
+NOTICE:  Test 6.4 - Retrieved 1 lines after writing
+NOTICE:    Line 1: [New 1]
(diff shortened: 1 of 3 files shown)


jiaoshuntian:
LGTM

Aias00:
Closing per the verification in #1546 - keeping unread lines queued matches Oracle's documented DBMS_OUTPUT semantics; not a defect.

PR #1549, opened 2026-08-11
Remove insecure cpanm downloads from regression workflows

### LINKED ISSUE #1548: Regression workflows execute cpanm downloaded over plain HTTP
### IvorySQL Version

### Current Behavior

Four GitHub Actions workflows download an executable from `http://cpanmin.us`, mark it executable, and run it with `--sudo`:

- `.github/workflows/contrib_regression.yml`
- `.github/workflows/oracle_pg_regression.yml`
- `.github/workflows/oracle_regression.yml`
- `.github/workflows/pg_regression.yml`

```sh
curl -L -o cpanm http://cpanmin.us && chmod +x cpanm && \
./cpanm --sudo IPC::Run && \
```

The HTTP endpoint currently serves the script directly instead of redirecting the initial request to TLS, so the downloaded executable is not protected against in-transit modification before privileged execution.

### Expected behavior/code

Do not download or execute this script. Each affected workflow already installs Ubuntu's `libipc-run-perl` package in the immediately preceding `apt-get install` step, which provides the required `IPC::Run` module through the runner's package manager.

### Step to reproduce

1. Inspect the four configure steps listed above.
2. Inspect their dependency installation steps and note that `libipc-run-perl` is already present.
3. Request the configured URL without following redirects; it responds successfully over plain HTTP.

Removing the two `cpanm` lines also removes redundant network access and eliminates a dependency on the availability of `cpanmin.us`.

Aias00:
## Summary

- stop downloading `cpanm` over plain HTTP in four regression workflows
- stop executing the downloaded script with `--sudo`
- rely on the existing `libipc-run-perl` apt dependency for `IPC::Run`

Fixes #1548

## Testing

- parsed all four changed workflow files with Ruby's YAML parser
- verified all four workflows still install `libipc-run-perl`
- verified no `cpanmin.us` or `cpanm` references remain under `.github/workflows`
- `git diff --check`

`actionlint` is not installed in the local environment; GitHub Actions provides the final workflow validation.

## Summary by CodeRabbit

- **Chores**
  - Streamlined regression test setup by removing unnecessary Perl dependency installation steps.
  - Configuration workflows now proceed directly to the project setup phase.

### DIFF
--- a/.github/workflows/contrib_regression.yml
+++ b/.github/workflows/contrib_regression.yml
@@ -23,8 +23,6 @@ jobs:
               libicu-dev libnuma-dev
     - name: configure - linux
       run: |
-        curl -L -o cpanm http://cpanmin.us && chmod +x cpanm && \
-        ./cpanm --sudo IPC::Run && \
         ./configure \
             --prefix=$PWD/inst \
             --enable-cassert --enable-debug --enable-tap-tests --enable-rpath \
--- a/.github/workflows/oracle_pg_regression.yml
+++ b/.github/workflows/oracle_pg_regression.yml
@@ -28,8 +28,6 @@ jobs:
     - name: configure - linux
       if: ${{ matrix.os == 'ubuntu-latest' }}
       run: |
-        curl -L -o cpanm http://cpanmin.us && chmod +x cpanm && \
-        ./cpanm --sudo IPC::Run && \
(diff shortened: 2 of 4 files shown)


jiaoshuntian:

Aias00:

sounds good to me

PR #1551, opened 2026-08-11
Use a supported runner for Build and Test workflow

### LINKED ISSUE #1550: Build and Test workflow uses removed ubuntu-18.04 runner
### IvorySQL Version

### Current Behavior

The manually dispatched `.github/workflows/build_and_test.yml` workflow declares:

```yaml
runs-on: ubuntu-18.04
```

`ubuntu-18.04` is no longer an available standard GitHub-hosted runner label, so GitHub cannot schedule the regression matrix when this workflow is dispatched.

### Expected behavior/code

Use a currently supported pinned runner label. `ubuntu-22.04` is listed in GitHub's current hosted-runners reference and provides the Docker-capable Linux host needed by this workflow.

### Step to reproduce

1. Open `.github/workflows/build_and_test.yml`.
2. Observe `jobs.regression.runs-on: ubuntu-18.04`.
3. Compare it with GitHub's current list of standard hosted runner labels; `ubuntu-18.04` is absent.

The workflow is `workflow_dispatch`-only and has no recent runs in the repository, which makes the stale runner easy to miss during normal pull-request validation.

GitHub hosted-runners reference: https://docs.github.com/en/actions/reference/runners/github-hosted-runners

Aias00:
## Summary

- replace the removed `ubuntu-18.04` hosted runner label with the supported, pinned `ubuntu-22.04` label
- restore schedulability of the manually dispatched CentOS container regression matrix

Fixes #1550

## Testing

- parsed `.github/workflows/build_and_test.yml` with Ruby's YAML parser
- checked the label against GitHub's current hosted-runners reference
- `git diff --check`

## Reference

- https://docs.github.com/en/actions/reference/runners/github-hosted-runners

## Summary by CodeRabbit

* **Chores**
  * Updated the regression workflow environment to Ubuntu 22.04.
  * Build, testing, artifact collection, and cleanup processes remain unchanged.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   regression:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         container-os: [centos-7, centos-8]


jiaoshuntian:

Aias00:

sounds good to me

PR #1553, opened 2026-08-11
Fix Docker quick start binary path

### LINKED ISSUE #1552: Docker quick start cannot find binaries installed under its prefix
### IvorySQL Version

### Current Behavior

The Docker development quick start configures IvorySQL with:

```sh
./configure --prefix=/home/ivorysql/ivorysql ...
make install
```

It then invokes `initdb` and `pg_ctl` by bare command name. The development image does not add `/home/ivorysql/ivorysql/bin` to `PATH`, so a fresh container cannot find the binaries just installed under that prefix.

### Expected behavior/code

Add the selected installation's `bin` directory to `PATH` after `make install` and before invoking installed tools:

```sh
export PATH=/home/ivorysql/ivorysql/bin:$PATH
```

This also ensures the quick start uses the newly built IvorySQL binaries rather than an unrelated system PostgreSQL installation if one is present.

### Step to reproduce

1. Start the documented development container.
2. Follow the quick-start configure, build, and install commands.
3. Run the documented `initdb -D data_ora -m oracle` command.
4. The shell cannot resolve `initdb` because the configured prefix's `bin` directory is not in `PATH`.

`.devcontainer/Dockerfile` sets the working directory and user but defines no `PATH` entry for `/home/ivorysql/ivorysql/bin`.

Aias00:
## Summary

- use the Docker quick start's configured installation prefix when invoking `initdb` and `pg_ctl`
- ensure the quick start resolves to the IvorySQL binaries built by the preceding commands without modifying the user's `PATH`

Fixes #1552

## Testing

- validated the updated quick-start shell block with `bash -n`
- confirmed the documented binary paths match `--prefix=/home/ivorysql/ivorysql`
- `git diff --check`

## Summary by CodeRabbit

* **Documentation**
  * Updated development setup instructions to invoke the database initialization and server-control tools using their absolute installation paths.
  * This makes the documented workflow more reliable when those tools are not available through the system `PATH`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,13 +51,12 @@ docker compose exec dev bash
 
 make -j$(nproc)
 make install
-export PATH="/home/ivorysql/ivorysql/bin:$PATH"
 
 # Initialize database in Oracle mode
-initdb -D data_ora -m oracle
+/home/ivorysql/ivorysql/bin/initdb -D data_ora -m oracle
 
 # Start the server
-pg_ctl -D data_ora start
+/home/ivorysql/ivorysql/bin/pg_ctl -D data_ora start
 
 # Run tests
 make oracle-check
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ docker compose exec dev bash
 
 make -j$(nproc)
 make install
+export PATH="/home/ivorysql/ivorysql/bin:$PATH"
 
 # Initialize database in Oracle mode
 initdb -D data_ora -m oracle


NotHimmel:
@Aias00 Thanks for your contribution! The README quick start should stay consistent with upstream's style: don't modify the user's PATH as part of the build steps. PATH is part of the user's own environment configuration and shouldn't be decided for them by a quick start, especially where multiple PostgreSQL/IvorySQL installations may coexist.

Aias00:
> @Aias00 Thanks for your contribution! The README quick start should stay consistent with upstream's style: don't modify the user's PATH as part of the build steps. PATH is part of the user's own environment configuration and shouldn't be decided for them by a quick start, especially where multiple PostgreSQL/IvorySQL installations may coexist.

 > Thanks for the clarification. Agreed—the quick start should not modify the user’s PATH. I’ll instead invoke the binaries through the installation prefix explicitly:
  >
  > /home/ivorysql/ivorysql/bin/initdb -D data_ora -m oracle
  > /home/ivorysql/ivorysql/bin/pg_ctl -D data_ora start
  >
  > This keeps the environment untouched while ensuring the commands use the IvorySQL binaries built and installed by the preceding steps.

PR #1556, opened 2026-08-11
Fix Ivyresult cleanup on early returns

### LINKED ISSUE #1555: Named-bind preparation leaks Ivyresult objects on early returns
## Problem

`Ivyreplacenamebindtoposition()` and `Ivyreplacenamebindtoposition2()` allocate an `Ivyresult` through `Ivyexec()`, but several return paths do not release it with `Ivyclear()`.

Affected paths in `src/interfaces/libpq/ivy-exec.c` include:

- the `PGRES_TUPLES_OK` failure path in both helpers;
- the successful early return when `stmtHandle->namebind == NULL` in `Ivyreplacenamebindtoposition()`.

Because `Ivyexec()` allocates both the `Ivyresult` wrapper and an underlying `PGresult`, repeated prepares/executions on these paths leak client memory.

## Expected behavior

Every non-NULL result returned by `Ivyexec()` should be passed to `Ivyclear()` before the helper returns, on both success and error paths.

## Suggested fix

Add cleanup at the affected returns while preserving the existing return values and error messages. Exercise the prepare/execute paths repeatedly under a leak detector or sanitizer.

Aias00:
## Summary

- release the `Ivyresult` when parameter-description execution fails;
- release it before the successful no-namebind early return;
- apply the same error-path cleanup to the OCI-style execution helper.

Fixes #1555.

## Validation

- `make -C src/interfaces/libpq ivy-exec.o -j2`
- `git diff --check`
- audited each changed return path to ensure `res` is no longer used after `Ivyclear()`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling during parameter-name replacement.
  * Prevented stale or incomplete results from being returned when execution fails or no parameter bindings are available.

### DIFF
--- a/src/interfaces/libpq/ivy-exec.c
+++ b/src/interfaces/libpq/ivy-exec.c
@@ -3819,6 +3819,7 @@ Ivyreplacenamebindtoposition(Ivyconn *tconn,
 		{
 			snprintf(errmsg, size_error_buf, "%s", IvyerrorMessage(tconn));
 			free(query);
+			Ivyclear(res);
 			return 0;
 		}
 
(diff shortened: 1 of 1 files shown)


jiaoshuntian:
LGTM..

PR #1558, opened 2026-08-11
Use vault repositories for CentOS 7 CI

### LINKED ISSUE #1557: CentOS 7 CI image still uses retired mirrorlist repositories
## Problem

The `centos-7` leg of `.github/workflows/build_and_test.yml` starts from `centos:7` and immediately runs `yum` against the image's default `mirrorlist.centos.org` repositories.

CentOS Linux 7 reached end of life on June 30, 2024, and the normal mirror network has been retired. A current container build therefore fails before installing its build dependencies because the configured mirrorlist is no longer usable.

CentOS EOL reference: https://www.centos.org/centos-linux-eol/

## Reproduction

```sh
docker run --rm centos:7 yum makecache
```

The default `CentOS-Base.repo` entries point to `mirrorlist.centos.org`. The workflow also installs Software Collections repositories that use the same retired mirrorlist.

## Expected behavior

The CentOS 7 CI image should use the archived CentOS 7.9.2009 base and SCL repositories so dependency installation remains reproducible.

## Suggested fix

- rewrite the base repositories to `vault.centos.org/7.9.2009` before the first `yum install`;
- rewrite the installed SCL repositories to their vault URLs before installing LLVM packages;
- use HTTPS for the EPEL signing-key download.

Aias00:
## Summary

- point the CentOS 7 base repositories at the archived 7.9.2009 vault;
- point the installed Software Collections repositories at their vault location;
- fetch the EPEL 7 signing key over HTTPS.

Fixes #1557.

## Validation

- parsed `.github/workflows/build_and_test.yml` with Ruby YAML
- built the extracted CentOS 7 Dockerfile successfully as `linux/amd64`
- verified the full dependency set, including `llvm-toolset-7-llvm-devel`, installs from the vault-backed repositories

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -17,8 +17,14 @@ jobs:
       run: |
         docker build -t ivorysql-${{ matrix.container-os }}:latest -<<EOF
         FROM    centos:7
-        RUN     rpm --import http://download.fedoraproject.org/pub/epel/RPM-GPG-KEY-EPEL-7
+        RUN     sed -i -e 's|^mirrorlist=|#mirrorlist=|g' \
+                       -e 's|^#baseurl=http://mirror.centos.org/centos/\$releasever|baseurl=http://vault.centos.org/7.9.2009|g' \
+                       /etc/yum.repos.d/CentOS-Base.repo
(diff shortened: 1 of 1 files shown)


NotHimmel:

Aias00:

sounds good

PR #1560, opened 2026-08-11
Use unique artifact names for build matrix jobs

### LINKED ISSUE #1559: Build matrix jobs upload regression logs under the same artifact name
## Problem

The `Build and Test` workflow runs a matrix for `centos-7` and `centos-8`, but both failure paths upload an artifact named `results` with `actions/upload-artifact@v4`.

Version 4 does not allow multiple jobs to upload to the same artifact name. If both matrix jobs fail, the second upload can fail with an artifact-name conflict and hide one platform's regression logs.

The action's official documentation specifically requires unique names in matrix scenarios: https://github.com/actions/upload-artifact#not-uploading-to-the-same-artifact

## Expected behavior

Each matrix job should upload its own regression archive under a unique, platform-identifying artifact name.

## Suggested fix

Include `${{ matrix.container-os }}` in the artifact name, for example `results-centos-7` and `results-centos-8`.

Aias00:
## Summary

Give each build-matrix job a platform-specific regression artifact name so concurrent `upload-artifact@v4` uploads cannot collide.

Fixes #1559.

## Validation

- parsed the workflow as YAML
- expanded the matrix expression locally and confirmed the two names are unique: `results-centos-7`, `results-centos-8`
- `git diff --check`

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -94,7 +94,7 @@ jobs:
       if: failure()
       uses: actions/upload-artifact@v4
       with:
-        name: results
+        name: results-${{ matrix.container-os }}
         path: ${{ github.workspace }}/regression.tar.gz
 
     - name: Cleanup


NotHimmel:
Same as my comment on #1558: The workflow is running normally.

Aias00:
> Same as my comment on #1558: The workflow is running normally.

fine

PR #1564, opened 2026-08-11
Fix pgtest -n status initialization

### LINKED ISSUE #1563: src/tools/pgtest: -n fails before running make
### IvorySQL Version

master at b21cd664004bc4383edfb7944712599818d635af

### OS Version

macOS

### Configuration options

Not configuration-dependent.

### Current Behavior

Running src/tools/pgtest -n skips the clean command, but the script only creates $TMP/ret after running make clean. It then reads the missing file on lines 36, 46, and 57.

The isolated reproduction exits with status 255 before make is invoked and reports:

    cat: /tmp/<pid>/ret: No such file or directory
    [: : integer expression expected
    exit: : numeric argument required

### Expected behavior/code

The -n option should skip make clean, then run the build and check stages. The script should preserve and return the first failing command status.

### Step to reproduce

From the repository root, run:

    src/tools/pgtest -n

The failure occurs before the build starts.

### Additional context

Initialize the stored return code to success before the optional clean stage, so both the normal path and the -n path enter the build/check stage. Add an isolated shell-level regression check using a no-op make command.

Aias00:
## Summary

- initialize the stored command status before the optional clean stage
- allow pgtest -n to proceed to build and check without reading a missing file
- preserve the exit status of clean, build, and check failures

Fixes #1563.

## Validation

- reproduced the previous behavior in an isolated directory: exit status 255 with missing ret-file and numeric-comparison errors
- sh -n src/tools/pgtest
- isolated no-clean path with a successful make: exit status 0
- isolated no-clean path with a failing make: exit status 1
- isolated clean path with a successful make: exit status 0
- isolated clean path with a failing make: exit status 1
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved build and check reliability by ensuring a successful status is reported when no cleanup step is requested.

### DIFF
--- a/src/tools/pgtest
+++ b/src/tools/pgtest
@@ -28,6 +28,7 @@ rm -f tmp_install/log/install.log
 
 # Run "make check" and store return code in $TMP/ret.
 # Display output but also capture it in $TMP/0.
+echo 0 > "$TMP"/ret
 (
 	if [ "$CLEAN" = 'Y' ]
 	then	$MAKE "$@" clean 2>&1


jiaoshuntian:
This file comes from the upstream PostgreSQL, so you may want to discuss it with the PG hackers first.

Aias00:
> This file comes from the upstream PostgreSQL, so you may want to discuss it with the PG hackers first.

fine, closing this

Issue #1565, opened 2026-08-11
Oracle-compatible VSIZE function


xuhuicai:
## Feature Request

Oracle-compatible VSIZE function

jiaoshuntian:
We look forward to your contribution.

PR #1566, opened 2026-08-11
Add Oracle-compatible VSIZE function

xuhuicai:
VSIZE returns the number of bytes in the internal representation of an argument, matching Oracle semantics: character data reports its byte length (VSIZE('abc') = 3), fixed-width types report their storage width, and NULL input yields NULL.

releated to #1565 

## Summary by CodeRabbit

* **New Features**
  * Added Oracle-compatible `sys.vsize`, which reports the byte size of compatible values.
  * Supports character, multibyte, numeric, Boolean, date, timestamp, fixed-width, table-column, NULL, and large text values.
  * Handles VARCHAR2 values and compressed or toasted large values consistently.

* **Tests**
  * Added comprehensive regression coverage validating byte lengths, NULL handling, varied data types, and consistency with `lengthb`.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -71,6 +71,7 @@ ORA_REGRESS = \
 	ora_character_datatype_functions \
 	ora_datetime_datatype_functions \
 	ora_misc_functions \
+	ora_vsize \
 	ora_merge \
 	datatype_and_func_bugs \
 	ora_sysview \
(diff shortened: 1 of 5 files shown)


bigplaice:
Thanks for bringing new Oracle compatible feature to IvorySQL!

bigplaice:
@xuhuicai 
Hi, I have a couple of general comments regarding to the regression test file. 
1. no need to add sys prefix for vsize function as the sys has been added into search path;
2. vsize('abc'::varchar2) uses the PG style of type casting. Since we are developing oracle compatible feature, we should use vsize(CAST('abc' AS VARCHAR2)) instead;

xuhuicai:
@bigplaice Thanks for the review, both comments are addressed.

PR #1569, opened 2026-08-12
Fix heap buffer overrun in named-bind query escaping

### LINKED ISSUE #1554: Named-bind query escaping can overrun its heap buffer
### IvorySQL Version

### Current Behavior

The two `get_parameter_description()` query-building paths in `src/interfaces/libpq/ivy-exec.c` allocate `2 * query_length` bytes for an escaped query, duplicate every single quote, and then append a NUL terminator.

For an input consisting entirely of single quotes, the escaped bytes consume the whole allocation and the terminator is written one byte past the heap buffer. The `Ivyreplacenamebindtoposition2()` path also has no empty-query guard, so it can write a terminator through a zero-length allocation.

Affected code:

- `Ivyreplacenamebindtoposition()` around lines 3791-3811
- `Ivyreplacenamebindtoposition2()` around lines 4078-4099

### Expected behavior/code

The escaped-query buffer must include space for the terminating NUL and reject allocation-size overflow. Empty input must also produce a valid one-byte empty string buffer.

### Step to reproduce

Build or run the affected path with AddressSanitizer and prepare a query whose bytes are all single quotes, for example `''`.

The current size calculation allocates four bytes, writes four escaped quote bytes, and then writes a fifth NUL byte.

Both functions contain the same manual escaping loop, so both entry points require the same boundary correction. This issue is limited to the allocation boundary; changing the escaping or query semantics is not necessary.

Aias00:
## Summary

Fixes #1554: the escaped-query buffer in the named-bind path was allocated as `2 * query_length` bytes. When the input query consists entirely of single quotes, the escaping loop writes exactly `2 * query_length` bytes and the terminating NUL is then written one byte past the heap allocation.

Four sites share the same manual escaping pattern and get the same fix:

- `Ivyreplacenamebindtoposition()` in `src/interfaces/libpq/ivy-exec.c` — allocate `2 * len + 1`, compute with `size_t` arithmetic, check `malloc()` failure (previously unchecked, would crash on OOM).
- `Ivyreplacenamebindtoposition2()` in `src/interfaces/libpq/ivy-exec.c` — same fix; the `+1` also makes empty input produce a valid one-byte empty string buffer instead of a NUL write through a zero-length `malloc(0)`.
- `get_hostvariables()` in `src/bin/psql/common.c` — same buffer overrun confirmed; allocate `2 * len + 1` (empty input was already safe here because `pg_malloc0(0)` maps to a 1-byte allocation).
- `psqlplus VARIABLE` missing-termination-quote handling in `src/bin/psql/mainloop.c` — same overrun: a quote-only `init_value` fills the `2 * len` buffer and the terminating NUL lands one byte past; allocate `2 * len + 1`.

Reviewed other quote-doubling sites for the same pattern: `pg_upgrade/util.c`, `psql/command.c` and `tab-complete.in.c` already size with `+2/+3`; `fe_utils/string_utils.c` uses a growing PQExpBuffer; `fe-connect.c` and `tsearchcmds.c` are parsers, not escaping copies.

## Test plan

- `make -C src/interfaces/libpq ivy-exec.o`, `make -C src/bin/psql common.o` and `make -C src/bin/psql mainloop.o` compile cleanly with the changes.
- Full libpq link is currently blocked on this machine by a pre-existing, unrelated `ivy_sema.c` build error on macOS (separate branch `fix-ivy-sema-include-guards`); the changed translation units themselves compile cleanly.
- Recommend an AddressSanitizer run of the affected paths (prepare a query whose bytes are all single quotes, e.g. `''`) to confirm the overrun is gone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed SQL text handling to ensure duplicated quotes are safely terminated.
  * Improved query parameter escaping to prevent truncated or malformed queries.
  * Added graceful handling for memory-allocation failures during query processing.

### DIFF
--- a/src/bin/psql/common.c
+++ b/src/bin/psql/common.c
@@ -3643,7 +3643,7 @@ get_hostvariables(const char *sql, bool *error)
 		return NULL;
 
 	/* double write quote */
-	newsql = pg_malloc0(strlen(sql) * 2);	/* enough */
+	newsql = pg_malloc0(strlen(sql) * 2 + 1);	/* doubled quotes plus NUL */
 	ptr = newsql;
 
 	while (*sql != '\0')
--- a/src/bin/psql/mainloop.c
+++ b/src/bin/psql/mainloop.c
@@ -492,7 +492,7 @@ MainLoop(FILE *source)
 		                         * init_value[0] indicates the type of missing quote: single or double quote
 		                         */
 		                        char	quote = bind_var->init_value[0];
-		                        char	*str_double_quote = pg_malloc0(strlen(bind_var->init_value) * 2);	/* Ensure enough space */
+		                        char	*str_double_quote = pg_malloc0(strlen(bind_var->init_value) * 2 + 1);	/* doubled quotes plus NUL */
 		                        char	*ptr = str_double_quote;
 		                        int		i;
 
--- a/src/interfaces/libpq/ivy-exec.c
+++ b/src/interfaces/libpq/ivy-exec.c
@@ -3797,7 +3797,13 @@ Ivyreplacenamebindtoposition(Ivyconn *tconn,
 			return 0;
 		}
 
-		newsql = malloc(stmtHandle->query_len * 2);	/* enough */
+		newsql = malloc((size_t) stmtHandle->query_len * 2 + 1);	/* doubled quotes plus NUL */
+		if (newsql == NULL)
+		{
+			free(query);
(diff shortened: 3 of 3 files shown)


jiaoshuntian:
It looks like they are doing the same thing. #1568

Aias00:
> It looks like they are doing the same thing. #1568

so let they do, it`s fine

PR #1583, opened 2026-08-13
fix(oracle): keep NULL line on PUT_LINE with pending PUT text

yyqdbngt:
## What is the purpose of the change

Fix #1582 (and #1584).

ora_dbms_output_put_line() flushed the pending PUT text but never appended the NULL line when the argument was NULL and pending PUT text existed. Oracle PUT_LINE(NULL) stores a NULL line after flushing the pending text.

## Brief changelog

- dbms_output.c: after flushing pending PUT text, append a NULL line when the argument is NULL
- add a regression test (Test 2.2b) covering pending PUT text followed by PUT_LINE(NULL)

## How was this patch verified

- Matches Oracle DBMS_OUTPUT.PUT_LINE(NULL) semantics
- Regression test asserts the pending line is flushed then a NULL line is appended

## Summary by CodeRabbit

- **Bug Fixes**
  - Re-enabling `DBMS_OUTPUT` now preserves unread output.
  - Adjusting the output buffer size while enabled no longer discards pending messages.
  - Calling `PUT_LINE(NULL)` now flushes pending text and adds a separate blank line.

- **Tests**
  - Added coverage confirming buffered output remains available after re-enabling `DBMS_OUTPUT`.
  - Added coverage verifying pending text and the blank line from `PUT_LINE(NULL)` are retrieved separately.

PR #1585, opened 2026-08-13
test(oracle): cover PUT_LINE(NULL) after pending PUT text

### LINKED ISSUE #1584: Add regression coverage for PUT_LINE(NULL) after pending PUT text
## Problem

The DBMS_OUTPUT test suite covers PUT(NULL) and PUT_LINE(NULL) separately but has no regression test for the combination of pending PUT text followed by PUT_LINE(NULL), which is the Oracle edge case fixed in #1582.

## Expected

Add a regression test asserting that pending PUT text is flushed and a NULL line is then appended.

btlqql:
## What is the purpose of the change

Fix #1584.

Add a regression test for the combination of pending PUT text followed by PUT_LINE(NULL), asserting the pending text is flushed and a NULL line is then appended.

## Brief changelog

- ora_dbms_output.sql / ora_dbms_output.out: add Test 2.2b

## How was this patch verified

- Extends the existing DBMS_OUTPUT regression suite

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed `PUT_LINE(NULL)` so pending `PUT` text is flushed before a separate null line is added.
  * Re-enabling output now preserves unread buffered content while applying the updated buffer limit.
* **Tests**
  * Added regression coverage for null-line handling and preserved output after re-enabling.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_dbms_output.out
+++ b/contrib/ivorysql_ora/expected/ora_dbms_output.out
@@ -176,7 +176,7 @@ BEGIN
 END;
 /
 NOTICE:  Test 3.2 - Buffer after disable: [<NULL>], Status: 1
--- Test 3.3: Re-ENABLE clears buffer
+-- Test 3.3: Re-ENABLE preserves unread output
 DECLARE
     line TEXT;
     status INTEGER;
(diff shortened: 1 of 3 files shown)


btlqql:
Superseded by #1583 which now includes the PUT_LINE(NULL) fix and its regression test together.

Issue #1586, opened 2026-08-13
CI: Use ccache to speed up autoconf build and regression workflows


xuhuicai:
## Summary

The GitHub Actions CI rebuilds IvorySQL from scratch on every push. The compile step takes a significant share of the job time, especially for the `build` workflow where compilation is ~87% of the total job duration. This issue proposes enabling [ccache](https://ccache.dev/) in the autoconf-based workflows to reuse compilation results across runs.

## Motivation (measured data)

| Workflow | Compile time | Share of job |
|---|---|---|
| `build` | ~409s | ~87% |
| `oracle_regression` | ~298s | ~29% (tests ~64%) |
| `pg_regression` / `contrib_regression` / `oracle_pg_regression` | similar | — |
| `meson_build` | ~78s | ~13% (tests ~80%) |

Enabling ccache can cut the compile time of `build` from ~7 min to ~1-2 min on cache hits, and save a few minutes per run on the regression workflows.

## Affected workflows

- `.github/workflows/build.yml` — cache key `runner.os-autoconf-werror` (uses `COPT="-Werror ..."`, kept separate from the others)
- `.github/workflows/oracle_regression.yml`
- `.github/workflows/pg_regression.yml`
- `.github/workflows/contrib_regression.yml`
- `.github/workflows/oracle_pg_regression.yml`
  - the above four share the cache key `runner.os-autoconf` because they use identical `configure` flags

## Expected impact

- `build` workflow: ~7 min → ~1-2 min on cache hits (compile 87% → mostly cached)
- regression workflows: a few minutes saved per run
- No change to test behavior; ccache only affects compilation

jiaoshuntian:
One question: where did the test data you provided come from?

PR #1587, opened 2026-08-13
CI: add ccache to autoconf build and regression workflows

xuhuicai:
## Summary

Enables [ccache](https://ccache.dev/) in the autoconf-based GitHub Actions workflows so compilation results are reused across CI runs.

Implements #1586.

## Changes

- `.github/workflows/build.yml` — cache key `runner.os-autoconf-werror` (compiles with `COPT="-Werror ..."`, kept separate from the others)
- `.github/workflows/oracle_regression.yml`
- `.github/workflows/pg_regression.yml`
- `.github/workflows/contrib_regression.yml`
- `.github/workflows/oracle_pg_regression.yml`
  - the above four share the cache key `runner.os-autoconf` because they use identical `configure` flags

Each workflow gets a `Setup ccache` step (`hendrikmuhs/ccache-action@v1.2`) plus `CC=ccache gcc` / `CXX=ccache g++` on the `configure` step. Settings: `max-size: 2G`, `append-timestamp: false` (fixed key, shared across PRs), `CCACHE_SLOPPINESS: time_macros`.

## Expected impact

- `build` workflow: compile is ~87% of job time (~7 min) → ~1-2 min on cache hits
- regression workflows: a few minutes saved per run
- No change to test behavior; ccache only affects compilation

## Why not meson_build

We deliberately do not enable ccache for `.github/workflows/meson_build_pg_test.yml`:

1. **Low benefit**: its `compile` step is only ~13% of the job time (~78s) while the `meson test` step dominates (~80%, ~470s). ccache would save at most ~1 minute per run.
2. **Extra maintenance cost**: the meson build needs its own cache key tied to the pinned `meson 1.0.1`/`ninja v1.10.1` toolchain and the `-Dcassert=true -Dbuildtype=debug` configuration, adding cache-management complexity for negligible gain.

## Test plan

- [ ] CI runs on this PR (first run is a cold cache; subsequent runs should hit it)
- [ ] No regression in build/regression workflow results

## Summary by CodeRabbit

* **Build and CI Improvements**
  * Improved caching across build and regression workflows to accelerate repeated builds.
  * Enhanced cache reuse and consistency between workflow runs.
  * Standardized compiler-cache configuration for more reliable build performance.
  * Enabled parallel compilation and installation using available processor capacity, shortening build times.
  * Improved workflow security by limiting automated build access to read-only repository content.

### DIFF
--- a/.github/workflows/contrib_regression.yml
+++ b/.github/workflows/contrib_regression.yml
@@ -21,7 +21,18 @@ jobs:
               libreadline-dev libedit-dev uuid-dev libossp-uuid-dev \
               libipc-run-perl libtime-hires-perl libtest-simple-perl \
               libicu-dev libnuma-dev
+    - name: Setup ccache
+      uses: hendrikmuhs/ccache-action@v1.2
+      with:
+        key: ${{ runner.os }}-autoconf
+        max-size: 2G
+        append-timestamp: false
+      env:
+        CCACHE_SLOPPINESS: time_macros
     - name: configure - linux
+      env:
+        CC: ccache gcc
+        CXX: ccache g++
       run: |
         curl -L -o cpanm http://cpanmin.us && chmod +x cpanm && \
         ./cpanm --sudo IPC::Run && \
--- a/.github/workflows/contrib_regression.yml
+++ b/.github/workflows/contrib_regression.yml
@@ -6,6 +6,9 @@ on:
   pull_request:
     branches: [ master , IVORY_REL_4_STABLE , IVORYSQL_REL_1_STABLE]
 
+permissions:
+  contents: read
+
 jobs:
   contrib_regression:
     runs-on: ubuntu-latest
@@ -21,7 +24,20 @@ jobs:
               libreadline-dev libedit-dev uuid-dev libossp-uuid-dev \
               libipc-run-perl libtime-hires-perl libtest-simple-perl \
               libicu-dev libnuma-dev
+    - name: Setup ccache
+      uses: hendrikmuhs/ccache-action@v1.2.23
+      with:
+        key: ${{ runner.os }}-autoconf
+        restore-keys: |
+          ${{ runner.os }}-autoconf
+        max-size: 2G
+        save: ${{ github.event_name == 'push' }}
+    - name: Configure ccache
+      run: echo "CCACHE_SLOPPINESS=time_macros" >> "$GITHUB_ENV"
     - name: configure - linux
+      env:
+        CC: ccache gcc
+        CXX: ccache g++
       run: |
         curl -L -o cpanm http://cpanmin.us && chmod +x cpanm && \
         ./cpanm --sudo IPC::Run && \
@@ -33,7 +49,7 @@ jobs:
             --with-ossp-uuid  --with-libxml --with-libxslt  --with-perl \
             --with-icu --with-libnuma --enable-injection-points --enable-nls
     - name: compile
-      run: make && make install
+      run: make -j"$(nproc)" && make install
 
     - name: contrib_regression
       run: |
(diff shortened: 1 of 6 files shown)


jiaoshuntian:
Have you tested these changes in your own repository?

xuhuicai:
> Have you tested these changes in your own repository?

Yes — the two runs are on the same commit (cold cache vs warm cache).

 ┌──────────────────────┬───────────┬───────────┬───────────────┐
 │ Workflow             │ 冷缓存 │ 热缓存 │ 提速          │
 ├──────────────────────┼───────────┼───────────┼───────────────┤
 │ build                │ 11m 12s   │ 1m 49s    │ ↓ 84%   │
 ├──────────────────────┼───────────┼───────────┼───────────────┤
 │ contrib_regression   │ 22m 43s   │ 4m 44s    │ ↓ 79%         │
 ├──────────────────────┼───────────┼───────────┼───────────────┤
 │ oracle_pg_regression │ 14m 12s   │ 3m 46s    │ ↓ 73%         │
 ├──────────────────────┼───────────┼───────────┼───────────────┤
 │ oracle_regression    │ 25m 12s   │ 15m 33s   │ ↓ 38%         │
 ├──────────────────────┼───────────┼───────────┼───────────────┤
 │ pg_regression        │ 26m 48s   │ 22m 41s   │ ↓ 15%         │
 ├──────────────────────┼───────────┼───────────┼───────────────┤
 │ meson_build          │ 21m 1s    │ 11m 6s    │ ↓ 47%         │
 ├──────────────────────┼───────────┼───────────┼───────────────┤
 │ 合计                 │ 120.9 min │ 59.7 min  │ ↓ 51%         │
 └──────────────────────┴───────────┴───────────┴───────────────┘

image
image

jiaoshuntian:
Nice find tackling #1586. One thing worth fixing before merge though: append-timestamp: false with no restore-keys means every workflow saves to a fixed cache key, and since GH Actions cache entries are immutable, the very first successful save is also the last one — every save after that silently no-ops.So in practice the cache only ever reflects whatever got compiled on the first run; anything touched in later commits just recompiles every time and never makes it back into the cache. Still correct output (ccache is content-addressed, no staleness risk there), just none of the speedup you're going for once code starts changing. Easiest fix is to drop append-timestamp: false and add a restore-keys prefix like the action's own default pattern does. Small side note: CCACHE_SLOPPINESS is set in the Setup ccache step's own env:, which doesn't carry over to the configure/compile steps, so as written it's not actually doing anything — probably wants to go through $GITHUB_ENV instead.

xuhuicai:
@jiaoshuntian Both issues are already fixed in the branch.

jiaoshuntian:
CodeRabbit also provided many valuable comments. Please take them into consideration and make the necessary changes as well.

xuhuicai:
All CodeRabbit comments have been addressed in the latest commit — least-privilege permissions, ccache-action upgrade to v1.2.23, and quoted -j"$(nproc)". Thanks!

.github/workflows/contrib_regression.yml
```diff
@@ -33,7 +49,7 @@ jobs:
             --with-ossp-uuid  --with-libxml --with-libxslt  --with-perl \
             --with-icu --with-libnuma --enable-injection-points --enable-nls
     - name: compile
-      run: make && make install
+      run: make -j"$(nproc)" && make -j"$(nproc)" install
```
jiaoshuntian: src/Makefile.global.in:474-476 contains an explicit warning: 

> Some EXTRA_INSTALL entries appear more than once in the tree, and parallel installs of the same file can fail with EEXIST.

Therefore, parallel installation is not recommended.

xuhuicai: **Fixed.** `make -j"$(nproc)"` now only parallelizes compilation; `make install` remains serial in all four regression workflows, per the EEXIST warning in `src/Makefile.global.in`.

jiaoshuntian:
Could you also enable ccache (plus -j) for the build-clang job, with its own cache key such as ${{ runner.os }}-clang-autoconf — since ccache hashes the compiler identity, sharing the gcc key would just dilute both.

xuhuicai:
Thanks! Added ccache + -j"$(nproc)" to the build-clang job with its own cache key (${{ runner.os }}-clang-autoconf) so it
won't dilute the gcc cache.

PR #1593, opened 2026-08-14
feat: add Oracle-compatible NANVL function

### LINKED ISSUE #1594: Oracle-compatible NANVL function
## Feature Request
Oracle-compatible NANVL function

jiaoshuntian:
close: #1594
Implement NANVL(n, m) per Oracle semantics: return m if n is NaN, otherwise n. 

Assisted-by: Claude:Sonnet-4.5

## Summary by CodeRabbit

- **New Features**
  - Added Oracle-compatible `NANVL` support for `NUMBER`, `BINARY_FLOAT`, and `BINARY_DOUBLE` values.
  - Returns a replacement when the first value is `NaN`, while preserving non-`NaN` values.
  - Supports NULLs, implicit conversions, negative zero, computed expressions, and table-based queries.

- **Tests**
  - Added comprehensive regression coverage for numeric types, infinity handling, documented examples, and invalid argument counts.


contrib/ivorysql_ora/expected/ora_nanvl.out
```diff
@@ -0,0 +1,259 @@
...
+ nanvl 
+-------
+ 
+(1 row)
+
+SELECT NANVL(CAST(1.5 AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT)) FROM DUAL;
+ nanvl 
+-------
+ 
+(1 row)
```
NotHimmel, line 87: STRICT on the sys.nanvl overloads drops one of Oracle's cases: expr1 valid, expr2 NULL.

Oracle only looks at expr1. If expr1 is not NaN it is returned and expr2 is never examined, so a NULL expr2 does not matter. STRICT makes the executor short-circuit to NULL before the function body runs, so that path is lost.

Verified on Oracle Database 26ai:

SELECT NANVL(CAST(1.5 AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT)) FROM dual;
return 1.5

Image

PR #1602, opened 2026-08-14
Fix resource leak when freeing unprepared statement handles

### LINKED ISSUE #1601: IvyFreePreparedStatement leaks resources for unprepared statement handles
### IvorySQL Version

### Current Behavior

`IvyFreePreparedStatement()` in `src/interfaces/libpq/ivy-exec.c` returns early when `stmtHandle->query == NULL`:

```c
if (stmtHandle == NULL || stmtHandle->query == NULL)
    return;

/* get the lock */
PGSemaphoreLock(&stmtHandle->lock);

/* already free */
if (stmtHandle->query == NULL)
{
    PGSemaphoreUnlock(&stmtHandle->lock);
    return;
}
```

A statement handle obtained from `IvyHandleAlloc(..., IVY_HANDLE_STMT, ...)` but never used to prepare a query (e.g. released with `IvyFreeHandle()` right after allocation) has `query == NULL` and hits this early return. That skips the whole release path:

- `outbind` / `namebind` linked lists are not freed
- `paramNames` array is not freed
- `IvyremoveStmtHandleRelation()` is not called, so the handle stays registered in the connection's relation list (dangling state after `free(stmtHandle)`)
- `IvyconnList`, `stmtName`, `paramTypes` are not freed

### Expected behavior/code

Releasing an unprepared handle should run the full cleanup path. Every resource free in that path (`free(stmtHandle->query)`, `free(stmtName)`, `free(paramTypes)`, ...) is NULL-safe, so removing the `query == NULL` guards is sufficient; the handle is then properly unregistered from the relation list and all sub-resources are released.

### Step to reproduce

```c
IvyPreparedStatement *stmt = NULL;
IvyHandleAlloc(NULL, (void **) &stmt, IVY_HANDLE_STMT, 0, NULL);
IvyFreeHandle(stmt, IVY_HANDLE_STMT);  /* never prepared a query */
```

### Additional context

All affected functions in the release path handle NULL safely; the fix is limited to removing the two early-return guards.

Aias00:
## Summary

Fixes #1601: `IvyFreePreparedStatement()` in `src/interfaces/libpq/ivy-exec.c` returned early when `stmtHandle->query == NULL`, so a statement handle allocated via `IvyHandleAlloc()` but never used to prepare a query leaked its sub-resources on release:

- `outbind` / `namebind` linked lists not freed
- `paramNames` array not freed
- `stmtName`, `paramTypes`, `IvyconnList` not freed
- `IvyremoveStmtHandleRelation()` skipped — the handle stays registered in the connection's relation list (dangling state after `free(stmtHandle)`)

The fix removes both early-return guards so the full cleanup path always runs; every `free()` in that path is NULL-safe.

Also adds a regression test (`testlibpq_unprepared_handle`) that allocates a statement handle, releases it without preparing any query, and verifies the release completes.

## Test plan

- New test registered in ivytest Makefile and regression.txt.
- (Full local run of the ivytest suite on macOS is blocked by a pre-existing, unrelated `ivy_sema.c` build issue on this machine.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup for statement handles that were allocated without a prepared query.
  * Prevented potential issues when releasing unprepared statement handles.

* **Tests**
  * Added regression coverage verifying that unprepared statement handles can be allocated and released successfully.

### DIFF
--- a/src/interfaces/libpq/ivy-exec.c
+++ b/src/interfaces/libpq/ivy-exec.c
@@ -837,18 +837,11 @@ IvyexecPreparedStatement2(Ivyconn *tconn,
 void
 IvyFreePreparedStatement(IvyPreparedStatement *stmtHandle)
 {
-	if (stmtHandle == NULL || stmtHandle->query == NULL)
+	if (stmtHandle == NULL)
 		return;
 
 	/* get the lock */
 	PGSemaphoreLock(&stmtHandle->lock);
-
-	/* already free */
-	if (stmtHandle->query == NULL)
-	{
-		PGSemaphoreUnlock(&stmtHandle->lock);
-		return;
-	}
 
 	/* free bind info of out parameter */
 	if (stmtHandle->outbind != NULL)
--- a/src/interfaces/libpq/ivytest/Makefile
+++ b/src/interfaces/libpq/ivytest/Makefile
@@ -16,7 +16,7 @@ LDFLAGS_INTERNAL += $(libpq_pgport)
 
 PROGS = testlibpq testlibpqstmt2 testlibpq_nullstmt testlibpq_allout testlibpq_prepare_plsql testlibpq_prepare_dml \
         testlibpq_binary_double testlibpq_binary_float testlibpq_interval testlibpq_number testlibpq_prepare_call \
-		testlibpq_call testlibpqstmt2_call
+		testlibpq_call testlibpqstmt2_call testlibpq_unprepared_handle
 
 all: $(PROGS)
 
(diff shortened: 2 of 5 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1601 .

Aias00:
Closing per the discussion in #1601 - the leak is real and the fix is CI-green, but it only triggers on an allocate-without-prepare cleanup pattern with negligible per-occurrence impact (160 bytes + one semaphore). Happy to resubmit if a maintainer wants the robustness fix.

PR #1609, opened 2026-08-14
Harden named-bind parameter handling in ivy-exec

### LINKED ISSUE #1608: ivy-exec: paramNames/paramValues handling lacks bounds checks and leaks on error paths
### IvorySQL Version

### Current Behavior

Several robustness problems in the named-bind parameter handling of `src/interfaces/libpq/ivy-exec.c`:

1. **Unchecked malloc (`Ivyreplacenamebindtoposition`, around line 3977)**: `stmtHandle->paramNames = malloc(...)` is not NULL-checked before `memset`/use — crash on OOM.

2. **Unchecked `position` index (around line 3988)**: `position = atoi(server value) - 1` is used to index `paramNames[position]` without verifying `0 <= position < nParams` — heap out-of-bounds write if the server returns an out-of-range position.

3. **NULL dereference on `paramNames[i]` (around line 3670)**: `strcmp(tmp1->name, stmtHandle->paramNames[i])` runs over `nParams` even when `paramNames[i]` is NULL — e.g. a DO statement executed through `IvyStmtExecute` with fewer binds than `nParams` (the `IvyHandleDostmt(false)` path does not allocate `paramNames`).

4. **Leak on SET failure (`IvyStmtExecute` around 2387 and `IvyStmtExecute2` around 2659)**: after `IvyhandleParamsValues` succeeds, if the follow-up `IvyExecCommand("set ... = true")` fails, the function returns NULL without freeing `paramValues`/`paramLengths`/`paramFormats` (and each `paramValues[i]`).

### Expected behavior/code

- Check malloc result before use
- Validate `position` against `nParams` before indexing
- Skip/guard NULL `paramNames[i]` entries in the lookup loop
- Free the parameter arrays on the SET-failure early return

### Step to reproduce

- OOM / corrupted server response for 1-2
- Execute a named-bind DO statement with fewer bound values than placeholders for 3
- Force the internal `set ... = true` command to fail for 4

Aias00:
## Summary

Fixes #1608. Four robustness problems in the named-bind parameter handling of `src/interfaces/libpq/ivy-exec.c`:

1. **Unchecked malloc**: `paramNames` allocation is now NULL-checked in all three `Ivyreplacenamebindtoposition*` variants before `memset`/use.
2. **Unchecked position index**: server-provided `position` (and psql-provided `hostvars[i].position`) is validated against `nParams` before indexing `paramNames[position]` — prevents heap out-of-bounds write on an out-of-range value.
3. **NULL dereference in strcmp loops**: the two name-matching loops now skip NULL `paramNames[i]` entries instead of passing them to `strcmp`.
4. **Leak on SET failure**: `IvyStmtExecute` / `IvyStmtExecute2` now route the failed `set ivorysql.out_parameter_column_position = true` through `ERROR_HANDLE`, which frees `paramValues`/`paramLengths`/`paramFormats` (previously returned NULL and leaked them).
5. `error_handle` cleanup is now tolerant of a NULL `paramNames` (so the malloc-failure path cannot crash the cleanup loop).

## Test plan

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when enabling session settings during statement execution.
  * Prevented invalid parameter positions from causing failures.
  * Improved handling of missing parameter names during parameter binding.
  * Ensured resources are safely cleaned up after execution errors.

### DIFF
--- a/src/interfaces/libpq/ivy-exec.c
+++ b/src/interfaces/libpq/ivy-exec.c
@@ -2391,7 +2391,7 @@ IvyStmtExecute(Ivyconn *tconn,
 		free(tresult);
 		tresult = NULL;
 		snprintf(errhp->error_msg, errhp->err_buf_size, "%s", "ivorysql.out_parameter_column_position failed");
-		return NULL;
+		goto ERROR_HANDLE;
 	}
 
 	/* send query */
@@ -2663,7 +2663,7 @@ IvyStmtExecute2(Ivyconn *tconn,
 		free(tresult);
 		tresult = NULL;
 		snprintf(errhp->error_msg, errhp->err_buf_size, "%s", "ivorysql.out_parameter_column_position failed");
-		return NULL;
+		goto ERROR_HANDLE;
 	}
 
 	/* send query */
@@ -3667,7 +3667,8 @@ IvyhandleParamsValues(IvyPreparedStatement *stmtHandle,
 
 			for (tmp1 = stmtHandle->namebind; tmp1 != NULL; tmp1 = tmp1->next)
 			{
-				if (strcmp(tmp1->name, stmtHandle->paramNames[i]) == 0)
+				if (stmtHandle->paramNames[i] != NULL &&
+					strcmp(tmp1->name, stmtHandle->paramNames[i]) == 0)
 					break;
 			}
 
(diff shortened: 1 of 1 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1608

Aias00:
Closing per the verification in #1608 - all four items need OOM, a buggy server response, or a mismatch the server already rejects; none is reachable in normal use. The hardening is correct and can be resubmitted if a maintainer wants it.

PR #1611, opened 2026-08-14
Fix psqlplus EXECUTE statement offset

### LINKED ISSUE #1610: psqlplus EXECUTE always reports Usage because the statement offset is wrong
### IvorySQL Version

### Current Behavior

In `src/bin/psql/psqlplusparse.y` the `exec_stmt` rule computes the offset of the executed statement text as the length of the whole scan buffer instead of the length of the `EXEC` keyword token:

```c
exec_stmt: exec_keyword
    {
        ...
        int offset = strlen(state->scanbuf);
        exec->plisqlstmts = pg_strdup(state->scanline + offset);
        ...
    }
```

`offset` should be the length of the matched `EXEC`/`EXECUTE` token (`yyleng`), not `strlen(scanbuf)` (the length of the entire buffered line). As a result `plisqlstmts` is always empty, and `EXEC[UTE] <stmt>` in psqlplus mode always fails (reports the usage error) instead of executing the statement. Depending on the `scanline`/`scanbuf` length relationship it can also read past the end of `scanline`.

### Expected behavior/code

Use the token length (`yyleng`) as the offset so `plisqlstmts` contains the text after `EXEC`.

### Step to reproduce

In psql (psqlplus/Oracle mode): `EXEC dbms_output.put_line('x');` — the command is not executed and a usage error is reported.

### Additional context

Oracle SQL*PLUS `EXEC[UTE]` executes a single PL/SQL statement; psqlplus mirrors this via `psqlplus_cmd_execute`.

Aias00:
## Summary

Fixes #1610: the `exec_stmt` rule in `src/bin/psql/psqlplusparse.y` computed the offset of the executed statement text as `strlen(state->scanbuf)` (the length of the entire buffered line) instead of the length of the matched `EXEC`/`EXECUTE` keyword. As a result `plisqlstmts` was always empty and `EXEC[UTE] <stmt>` always fell into the usage-error branch in `mainloop.c`.

The offset is now `strlen($1)` — the length of the matched keyword ("exec" or "execute"), which the scanner stores in `yylval->keyword`. Leading whitespace after the keyword is already handled by `psqlplus_skip_space()` at the consumer.

## Test plan

- `make -C src/bin/psql psqlplusparse.o` regenerates the parser with bison and compiles cleanly.
- Manual: `EXEC dbms_output.put_line('x');` in psqlplus mode should now execute instead of printing the usage error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing of `EXEC` and `EXECUTE` statements to correctly identify the beginning of PL/SQL text.


hanjianqiao:
@Aias00 Before we continue reviewing this PR, please update the corresponding issue with the specific steps to reproduce the error in a real environment. It would also be helpful if you could attach the actual input and output.

Aias00:
Closing per the verification in #1610 - the reported bug does not reproduce: flex's hold-character mechanism makes strlen(scanbuf) equal the matched keyword length at runtime, so the offset is correct and EXEC[UTE] works (verified with debug instrumentation). The PR's change is behaviorally equivalent, which is why its CI passed.

PR #1613, opened 2026-08-14
NUL-terminate copied OUT parameter values

### LINKED ISSUE #1612: psql PBE: OUT value copied without NUL terminator can overrun the bind buffer
### IvorySQL Version

### Current Behavior

In `src/bin/psql/common.c` (`IvyassignOutParameters`, around line 4207), when an OUT value exactly fills the destination buffer, only `memcpy` is performed and no NUL terminator is written:

```c
memcpy(var[i], value, value_len);
```

`var[i]` was allocated as `(typmod - VARHDRSZ) * pg_encoding_max_length(...) + 1` and zeroed, so a value that exactly fills `var_sz[i] - 1` bytes leaves no NUL; a value that fills `var_sz[i]` bytes (the +1 slot) is then followed by `AssignBindVariable()` -> `strlen()` reading past the end of the heap buffer.

### Expected behavior/code

Write the NUL terminator after the copy (or skip trailing bytes beyond the buffer) so the resulting string is always terminated.

### Step to reproduce

Bind a VARCHAR2/CHAR variable whose OUT value length equals the allocated buffer size and read it back; run under AddressSanitizer to observe the heap-buffer-overflow read.

### Additional context

Client-side (psql) issue; the value source is the server response, which is not length-limited the same way the declared type is.

Aias00:
## Summary

Fixes #1612: when an OUT parameter value is copied into the user's bind buffer, `IvyassignOutParameters()` and `assign_value_internel()` in `src/interfaces/libpq/ivy-exec.c` did not guarantee a NUL terminator:

- truncation path: `memcpy(var, value, val_size)` wrote exactly `val_size` bytes with no NUL — psql's `AssignBindVariable()` then ran `strlen()` past the heap allocation.
- full-copy path: when `len == val_size` the buffer was also filled completely (the `>` comparison sent it to the full-copy branch); the NUL only existed because the caller zeroed the buffer, which fails exactly when the value fills it.

The fix:
- truncation now copies `val_size - 1` bytes and writes the NUL at `val_size - 1` (condition changed to `>=` so the exact-fit case is also truncated safely)
- the full-copy path writes the NUL at `len` (safe since `len < val_size` there)

Both byte-for-byte contents are unchanged except for the guaranteed terminator.

## Test plan

- Recommend an AddressSanitizer run binding a VARCHAR2 OUT whose returned value length equals the declared buffer size.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of output string values when destination buffers are full or truncated.
  * Ensured copied strings are properly terminated to prevent incomplete or invalid output.

### DIFF
--- a/src/interfaces/libpq/ivy-exec.c
+++ b/src/interfaces/libpq/ivy-exec.c
@@ -2916,7 +2916,7 @@ IvyassignOutParameters(const Ivyresult *tres, IvyBindOutInfo *outbind)
 				struct pgresAttValue *attrvalue;
 
 				attrvalue = &res->tuples[0][serbind->resultcolumnno];
-				if (attrvalue->len > 0 && attrvalue->len >= serbind->val_size)
+				if (attrvalue->len > 0 && attrvalue->len > serbind->val_size)
 				{
 					unsigned int diff = attrvalue->len - serbind->val_size;
 
@@ -2946,7 +2946,8 @@ IvyassignOutParameters(const Ivyresult *tres, IvyBindOutInfo *outbind)
 							*(serbind->indp) = 0;
 
 						memcpy(serbind->var, attrvalue->value, attrvalue->len);
-						((char *) serbind->var)[attrvalue->len] = '\0';
+						if (attrvalue->len < serbind->val_size)
+							((char *) serbind->var)[attrvalue->len] = '\0';
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


NotHimmel:
Thanks for the report and PR, could you confirm how it was found and provide a concrete reproduction?

PR #1615, opened 2026-08-14
Fix uninitialized copy of PG_VERSION contents

### LINKED ISSUE #1614: ora_version: version file copied from uninitialized buffer without NUL guarantee
### IvorySQL Version

### Current Behavior

In `src/oracle_fe_utils/ora_version.c` (around line 74):

```c
memcpy(*version_str, buf, st.st_size);
```

The version string is copied from a `buf` that was never zero-initialized. `st.st_size` is the file size; if the file is smaller than the buffer, the copy includes uninitialized stack bytes after the file content, and the result is not guaranteed to be NUL-terminated unless the file itself happens to end with a NUL byte. Downstream `strlen`/`strcmp` on the returned string can read past the copied region.

### Expected behavior/code

Zero the buffer before the copy (or allocate `st.st_size + 1` and NUL-terminate).

### Step to reproduce

Read a version file whose content is not NUL-terminated (e.g. generated by a build step that writes exactly the file length without a trailing NUL); run under AddressSanitizer / Valgrind to observe uninitialized bytes or an over-read.

### Additional context

Client-side utility (oracle_fe_utils).

Aias00:
## Summary

Fixes #1614: `get_pg_version()` filled `version_str` with `memcpy(*version_str, buf, st.st_size)`, copying `st.st_size` bytes from a stack buffer that `fscanf("%63s", buf)` had only partially initialized — uninitialized stack bytes ended up in the result, and the string was not guaranteed to be NUL-terminated.

`buf` is already a valid NUL-terminated C string after `fscanf`, so the fix replaces the raw copy with `pg_strdup(buf)`, which copies exactly the parsed token and terminates it. The `st`/`st.st_size` usage for the size check remains unchanged.

## Test plan

- `make -C src/oracle_fe_utils ora_version.o` compiles cleanly.
- Recommend a Valgrind run of a caller that passes a non-NULL `version_str`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PostgreSQL version string handling to ensure reliably parsed version information is returned without fixed-size allocation limitations.

### DIFF
--- a/src/oracle_fe_utils/ora_version.c
+++ b/src/oracle_fe_utils/ora_version.c
@@ -69,10 +69,7 @@ get_pg_version(const char *datadir, char **version_str)
 	fclose(version_fd);
 
 	if (version_str)
-	{
-		*version_str = pg_malloc(PG_VERSION_MAX_SIZE);
-		memcpy(*version_str, buf, st.st_size);
-	}
+		*version_str = pg_strdup(buf);
 
 	if (v1 < 10)
 	{


NotHimmel:
@Aias00  Thanks for the report and PR. Before we treat this as a real bug, could you confirm how it was found and provide a concrete reproduction?

PR #1606, opened 2026-08-14
Fix heap overrun parsing XML namespace declarations

Aias00:
## Summary

Fixes #1603: `register_ns_from_csting()` in `contrib/ivorysql_ora/src/xml_functions/ora_xml_functions.c` copied namespace prefix/url tokens into `StringInfo` buffers with raw `memcpy`, bypassing the growth path. A single token longer than the initial 1024-byte buffer (e.g. `xmlns:p="<5000-byte-url>"` in `EXTRACTVALUE`/`XMLQUERY` namespace arguments) wrote past the heap.

Replaced both `memcpy` + manual NUL writes with `appendBinaryStringInfo`, which grows the buffer as needed and keeps it NUL-terminated. The copied byte counts are unchanged (`end - start - 1` for the prefix, `l1 - 3` for the url).

## Test plan

- `make -C contrib/ivorysql_ora src/xml_functions/ora_xml_functions.o` compiles cleanly.
- Recommend an AddressSanitizer run with a namespace token > 1024 bytes to confirm the overrun is gone.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of XML namespace prefixes and URLs during extraction.
  * Prevented potential string termination issues when processing XML data.

Issue #1603, opened 2026-08-14
ora_xml_functions: namespace prefix/url parsing overruns the heap buffer


Aias00:
### IvorySQL Version

`master` at `c3529ba4789ee220caefc858e58bbe6f1563ad14`

### Current Behavior

`register_ns_from_csting()` in `contrib/ivorysql_ora/src/xml_functions/ora_xml_functions.c` writes namespace prefix/url tokens into `StringInfo` buffers with raw `memcpy`, never growing them:

```c
StringInfoData prefix;
StringInfoData url;
initStringInfo(&prefix);   /* initial 1024-byte buffer, never enlarged */
initStringInfo(&url);

/* get the prefix */
start = strchr(tmp.data, ':');
end = strchr(tmp.data, '=');
memcpy(prefix.data, start + 1, end - start);      /* line 560 */
prefix.data[end - start - 1] = '\0';

/* get the url */
p1 = strstr(tmp.data, "=");
l1 = strlen(p1);
memcpy(url.data, p1 + 2, l1 - 3);                 /* line 566 */
url.data[l1 - 3] = '\0';
```

`initStringInfo` allocates 1024 bytes; `appendStringInfo*` would grow it, but the code bypasses the growth path with `memcpy`. A namespace string whose prefix or url token exceeds 1024 bytes (e.g. `xmlns:p="<5000-byte-url>"`) writes past the heap buffer.

### Expected behavior/code

Use `appendBinaryStringInfo` (or `enlargeStringInfo` before `memcpy`) so the buffer grows as needed.

### Step to reproduce

Call any XML function that routes through `ivy_xml_xpath` (e.g. `EXTRACTVALUE`/`XMLQUERY` with a namespace argument) and pass a namespace declaration with a single token longer than 1024 bytes. Build with AddressSanitizer to observe the heap-buffer-overflow write.

### Additional context

The input is user-controlled and reachable by any user with EXECUTE on the affected functions; this is a backend heap corruption.

Aias00:
Hi @NotHimmel,

We verified this report and assessed its reachability.

**How it was found**: static code review of `register_ns_from_csting()` in `ora_xml_functions.c`.

[...]

PR #1618, opened 2026-08-14
Fix namespace parsing crashes and out-of-bounds reads in XML functions

### LINKED ISSUE #1616: ora_xml_functions: namespace parsing crashes on missing '=' and reads out of bounds in several paths
### IvorySQL Version

### Current Behavior

Four defects in `contrib/ivorysql_ora/src/xml_functions/ora_xml_functions.c`, all reachable via the namespace argument of XML functions (EXTRACTVALUE/XMLQUERY/XMLTABLE etc.):

1. **NULL pointer arithmetic (backend crash, high)**: in `register_ns_from_csting()` (around line 558):
```c
start = strchr(tmp.data, ':');
end = strchr(tmp.data, '=');
appendBinaryStringInfo(&prefix, start + 1, end - start - 1);
```
`IS_STR_XMLNS()` only checks the `xmlns:` prefix, so a token like `xmlns:p` (prefix but no `=`) passes the check; `end` is NULL and `end - start` is undefined pointer arithmetic feeding a huge length into the copy — backend crash.

2. **Out-of-bounds read (line 542)**: `*(f - 1)` when the (trimmed) namespace string starts with `=`, e.g. `extractvalue(x, '/a', '="u"')` — reads one byte before the buffer.

3. **Out-of-bounds read in `right_trim()` (lines 453-455)**: a whitespace-only string decrements `end` past the start of the buffer:
```c
end = str + strlen(str) - 1;
while (isspace(*end)) end--;
```

4. **Out-of-bounds read in `rv_newline()` (line 417)**: `tmp[len - 1]` with `len == 0` (empty XML content) reads `tmp[-1]`.

### Expected behavior/code

- Validate that `end` (and `start`) is non-NULL before use; report "Invalid namespace" otherwise
- Guard `*(f - 1)` for `f == nslist`
- Bound the `right_trim` loop to the string start
- Handle `len == 0` in `rv_newline`

### Step to reproduce

- `SELECT extractvalue('<a>x</a>', '/a', 'xmlns:p')` crashes the backend
- ASan builds show the OOB reads for 2-4

Aias00:
## Summary

Fixes #1616. Four defects in `register_ns_from_csting()` / helpers in `contrib/ivorysql_ora/src/xml_functions/ora_xml_functions.c`, all reachable via the namespace argument of XML functions (EXTRACTVALUE/XMLQUERY/XMLTABLE):

1. **Backend crash (high)**: a token like `xmlns:p` passes the `IS_STR_XMLNS` prefix check but has no `=`; `end` is NULL and `end - start` is undefined pointer arithmetic feeding a huge length into the copy. Now rejected with "Invalid namespace" (`start`/`end` NULL or `end <= start`).
2. **OOB read**: `*(f - 1)` when the trimmed namespace string starts with `=` — now rejected (`f == nslist`).
3. **OOB read in `right_trim()`**: whitespace-only string walked `end` past the buffer start — loop now bounded by `end >= str`, with `(unsigned char)` cast for `isspace`.
4. **OOB read in `rv_newline()`**: `tmp[len - 1]` with `len == 0` — now guarded.

Also included (same function, previously filed as #1603): the prefix/url copies use `appendBinaryStringInfo` so tokens longer than the initial 1024-byte buffer grow instead of overrunning it. If #1606 merges first, this hunk will apply cleanly.

## Test plan

- `make -C contrib/ivorysql_ora src/xml_functions/ora_xml_functions.o` compiles cleanly.
- `SELECT extractvalue('<a>x</a>', '/a', 'xmlns:p')` now errors cleanly instead of crashing; ASan builds show no OOB reads for 2-4.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved XML namespace parsing to reject malformed namespace declarations.
  - Added safer handling for whitespace and trailing newline removal.
  - Improved construction of namespace prefixes and URLs to prevent malformed results and improve reliability when processing XML content.

### DIFF
--- a/contrib/ivorysql_ora/src/xml_functions/ora_xml_functions.c
+++ b/contrib/ivorysql_ora/src/xml_functions/ora_xml_functions.c
@@ -414,7 +414,7 @@ rv_newline(text *tv)
 
 	tmp = text_to_cstring(tv);
 	len = strlen(tmp);
-	if (tmp[len-1] == '\n')
+	if (len > 0 && tmp[len-1] == '\n')
 		tmp[len-1] = '\0';
 	ret = cstring_to_text(tmp);
 
@@ -451,7 +451,8 @@ right_trim(char *str)
 		return str;
 
 	end = str + strlen(str) - 1;
-	while (isspace(*end)) end--;
+	while (end >= str && isspace((unsigned char) *end))
+		end--;
 	*(end + 1) = '\0';
 
 	return str;
@@ -539,7 +540,7 @@ register_ns_from_csting(xmlXPathContextPtr xpathCtx, char* nsList)
 	f = strchr(nslist, (int)'=');
 	while (f != NULL)
 	{
-		if ((*(f - 1) == ' ') || (*(f + 1) != '"') || (*(f + 2) == ' '))
+		if ((f == nslist) || (*(f - 1) == ' ') || (*(f + 1) != '"') || (*(f + 2) == ' '))
 			elog(ERROR, "Invalid namespace");
 		f = strchr(f + 1, (int)'=');
 	}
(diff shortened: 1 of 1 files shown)


Aias00:
Closing per the assessment in #1616 - all four defects need malformed namespace strings (invalid calls). The defensive fixes can be resubmitted if hardening against bad calls is wanted.

PR #1619, opened 2026-08-14
Guard against NULL outbind entry in DO statement parameter setup

### LINKED ISSUE #1617: ivy-exec: NULL dereference when outbind positions are not contiguous
### IvorySQL Version

### Current Behavior

In `src/interfaces/libpq/ivy-exec.c` (around line 1890, the DO-statement parameter handling in `IvyStmtExecute`), the inner loop searches the `outbind` list for `info->position == i + 1` and then dereferences `info` without a NULL check:

```c
for (i = 0; i < stmtHandle->nParams; i++)
{
    for (info = stmtHandle->outbind; info != NULL; info = info->next)
    {
        if (info->position == i + 1)
            break;
    }

    stmtHandle->paramTypes[i] = (Oid) info->dtype;   /* info can be NULL */
    ...
}
```

If the caller binds positions 1 and 3 (skipping 2), no entry matches `i + 1 == 2`, `info` is NULL and `info->dtype` crashes the client process. The parallel loop in the same function (around line 1973) has `if (info != NULL)` protection; this one does not.

### Expected behavior/code

Report an error (and clean up) when no outbind entry matches the position, instead of dereferencing NULL.

### Step to reproduce

Client code: prepare a statement with 3 placeholders, `IvyBindByPos` only positions 1 and 3, then execute a DO statement so the `noparams_info` path is taken.

Aias00:
## Summary

Fixes #1617: in the DO-statement parameter setup of `IvyStmtExecute` (`src/interfaces/libpq/ivy-exec.c`), the inner loop searches `outbind` for `info->position == i + 1` and then dereferences `info` unconditionally. If the caller bound non-contiguous positions (e.g. 1 and 3, skipping 2), `info` is NULL and `info->dtype` crashes the client process.

The fix reports "outbind position not found" and cleans up (`paramTypes` + `query_buf`) before returning 0, matching the error style of the same function.

## Test plan

- Manual: bind positions 1 and 3 of a 3-placeholder statement and execute a DO statement — the client now gets an error instead of crashing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for missing output parameter bindings.
  * Improved error handling to prevent failures when an output parameter has no corresponding bind entry.

### DIFF
--- a/src/interfaces/libpq/ivy-exec.c
+++ b/src/interfaces/libpq/ivy-exec.c
@@ -1895,6 +1895,16 @@ IvyHandleDostmt(Ivyconn *tconn,
 						break;
 				}
 
+				if (info == NULL)
+				{
+					snprintf(errmsg, err_buf_size, "%s",
+							 "outbind position not found");
+					free(stmtHandle->paramTypes);
+					stmtHandle->paramTypes = NULL;
+					destroyPQExpBuffer(query_buf);
+					return 0;
+				}
+
 				stmtHandle->paramTypes[i] = (Oid) info->dtype;
 
 				switch(UnSetModeOut(info->dtype))


NotHimmel:
Thanks for the report and PR, could you confirm how it was found and provide a concrete reproduction?

Aias00:
Closing per the discussion in #1617 - the NULL deref is real but only reachable when a C client binds non-contiguous positions via IvyBindByPos (an unusual pattern no in-tree code produces); impact is a client-process crash only. Happy to resubmit the one-line guard as a focused PR if a maintainer wants it.

PR #1622, opened 2026-08-14
Fix out-of-bounds read in ora_asciistr on truncated UTF-8 sequences

### LINKED ISSUE #1620: ora_asciistr reads past the string end on truncated multi-byte sequences
### IvorySQL Version

### Current Behavior

`ora_asciistr()` in `contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c` decodes UTF-8 by inspecting `str[1]`/`str[2]`/`str[3]` without checking that those bytes are inside the input:

```c
} else if ((c & 0xE0) == 0xC0) {
    codePoint = ((c & 0x1F) << 6) | (str[1] & 0x3F);   /* str[1] may be past end */
    str += 2;
} else if ((c & 0xF0) == 0xE0) {
    codePoint = ... | (str[2] & 0x3F);                  /* str[2] may be past end */
    str += 3;
} else if ((c & 0xF8) == 0xF0) {
    codePoint = ... | (str[3] & 0x3F);                  /* str[3] may be past end */
    str += 4;
}
```

In any non-UTF-8 database (LATIN1/SQL_ASCII), a high byte near the end of the string (e.g. LATIN1 `é` = 0xE9 as the last character) matches the three-byte branch and reads past the end of the varlena data. Even in UTF-8 databases, a truncated sequence at the string end does the same.

### Expected behavior/code

Check that enough bytes remain (`str + N <= end`) before each multi-byte decode and raise the existing "Invalid bytes" error otherwise.

### Step to reproduce

In a LATIN1 or SQL_ASCII database: `SELECT asciistr('abc' || chr(233))` under AddressSanitizer — heap-buffer-overflow read.

Aias00:
## Summary

Fixes #1620: `ora_asciistr()` decoded UTF-8 by reading `str[1]`/`str[2]`/`str[3]` without checking those bytes are within the input varlena. A high byte near the end of the string (e.g. LATIN1 `é` = 0xE9 as the last character in a non-UTF8 database) matched the three-byte branch and read past the end of the data.

Each multi-byte branch now checks the remaining byte count (`str + N <= end`) and raises the existing "Invalid bytes" error for truncated sequences.

## Test plan

- `make -C contrib/ivorysql_ora src/builtin_functions/character_datatype_functions.o` compiles cleanly.
- Recommend an ASan run in a LATIN1/SQL_ASCII database: `SELECT asciistr('abc' || chr(233))`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UTF-8 input validation when converting text to ASCII-style representations.
  * Truncated multi-byte characters now return a clear invalid-parameter error instead of causing unsafe reads.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
@@ -1800,11 +1800,17 @@ ora_asciistr(PG_FUNCTION_ARGS) {
             str++;
         } else if ((c & 0xE0) == 0xC0) {
             /* Two-byte UTF-8 sequence */
+            if (str + 2 > end)
+                ereport(ERROR,
+                        (errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("Invalid bytes")));
             codePoint = ((c & 0x1F) << 6) | (str[1] & 0x3F);
             appendUTF16Escape(&output, (uint16_t) codePoint);
             str += 2;
         } else if ((c & 0xF0) == 0xE0) {
             /* Three-byte UTF-8 sequence */
+            if (str + 3 > end)
+                ereport(ERROR,
+                        (errcode(ERRCODE_INVALID_PARAMETER_VALUE), errmsg("Invalid bytes")));
             codePoint = ((c & 0x0F) << 12) | ((str[1] & 0x3F) << 6) | (str[2] & 0x3F);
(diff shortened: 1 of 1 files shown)


Aias00:
Closing per the assessment in #1620 - the trigger is a truncated UTF-8 sequence at the end of the input (malformed input), no normal-use exposure. The bounds-guard fix can be resubmitted if wanted.

PR #1629, opened 2026-08-14
Fix psql leaks and dbms_utility dlsym caching

### LINKED ISSUE #1628: psql: get_hostvariables leaks on early return; PBE unsupported-type branch skips cleanup; dbms_utility caches failed dlsym lookups
### IvorySQL Version

### Current Behavior

Three small cleanup/robustness defects:

1. **`get_hostvariables()` leaks on early return** (`src/bin/psql/common.c:3673`): when the server reports no placeholders (`ntuples == 1`), the function returns NULL without freeing `newsql`, the `PQExpBuffer query` and the `PGresult res`. Leaks per executed query whose text contains `:ident`-like tokens but no placeholders.

2. **`psql_exec_pbe` unsupported-type branch skips cleanup** (`src/bin/psql/common.c:4150-4157`): the `default:` case returns NULL directly instead of jumping to `pbe_failure`, leaking `var`/`bindinfo`/`indp`/`var_sz` and any already-allocated `var[i]`. (Currently unreachable since only 6 typoids are accepted, but the cleanup should be shared.)

3. **`dbms_utility` caches failed dlsym lookups** (`contrib/ivorysql_ora/src/builtin_packages/dbms_utility/dbms_utility.c:60-99`): `lookup_attempted` is set once before the `dlsym` calls; if the first invocation happens before `plisql.so` is loaded (direct SQL call), all four function pointers stay NULL forever and the error formatting never upgrades, even after plisql is loaded later.

### Expected behavior/code

- Free `newsql`/`query`/`res` before the early return in `get_hostvariables`
- Route the `default:` case through `pbe_failure`
- Only cache the dlsym results when the lookups succeed (retry otherwise)

Aias00:
## Summary

Fixes #1628. Three small fixes:

1. **`get_hostvariables()` leak** (`src/bin/psql/common.c`): the no-placeholder early return (`ntuples == 1`) now frees `newsql`, the `PQExpBuffer query` and the `PGresult res` before returning NULL.

2. **`psql_exec_pbe` unsupported-type branch** (`src/bin/psql/common.c`): the `default:` case now sets `res = NULL` and jumps to `pbe_failure` so `var`/`bindinfo`/`indp`/`var_sz` and any allocated `var[i]` are freed (previously returned directly, leaking them).

3. **`dbms_utility` dlsym caching** (`contrib/ivorysql_ora/src/builtin_packages/dbms_utility/dbms_utility.c`): `lookup_attempted` is only set when all four plisql symbols were found; a lookup performed before `plisql.so` was loaded is now retried on later calls instead of being permanently stuck with NULL function pointers.

## Test plan

- `make -C src/bin/psql common.o` and `make -C contrib/ivorysql_ora src/builtin_packages/dbms_utility/dbms_utility.o` compile cleanly.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_packages/dbms_utility/dbms_utility.c
+++ b/contrib/ivorysql_ora/src/builtin_packages/dbms_utility/dbms_utility.c
@@ -69,8 +69,6 @@ lookup_plisql_functions(void)
 	if (lookup_attempted)
 		return;
 
-	lookup_attempted = true;
-
 #ifndef WIN32
 	{
 		/*
@@ -95,7 +93,13 @@ lookup_plisql_functions(void)
 			get_call_stack_fn = (plisql_get_call_stack_fn) fn;
 	}
 #endif
-	/* On Windows, function pointers remain NULL - features require plisql */
+
+	/* Only cache the lookups if plisql was found; retry otherwise */
+	if (get_exception_context_fn != NULL &&
+		get_exception_message_fn != NULL &&
+		get_exception_sqlerrcode_fn != NULL &&
+		get_call_stack_fn != NULL)
+		lookup_attempted = true;
 }
 
 /*
--- a/src/bin/psql/common.c
+++ b/src/bin/psql/common.c
@@ -3671,7 +3671,12 @@ get_hostvariables(const char *sql, bool *error)
 
 		/* No placeholders do not need to use PBE */
 		if (ntuples == 1)
+		{
+			PQclear(res);
+			destroyPQExpBuffer(query);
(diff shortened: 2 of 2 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1628

Aias00:
Closing per the verification in #1628 - item 1 leaks a few hundred bytes per query in a short-lived psql process, item 2 is unreachable, item 3 does not trigger in practice (plisql.so is resident in Oracle-mode backends). Can be resubmitted if a maintainer wants the defensive cleanups.

PR #1631, opened 2026-08-14
Fix four datatype boundary defects

### LINKED ISSUE #1630: Datatype boundary fixes: numtoyminterval sign, to_date/to_timestamp dangling nlsparam if, yminterval dead check, binary underflow early return
### IvorySQL Version

### Current Behavior

Four small datatype boundary defects:

1. **`numtoyminterval` wrong sign for negative fractions** (`datetime_datatype_functions.c:1462`):
```c
if (interval_val <= -0.5 && interval_val > -1)
    interval_val = 1;      /* should be -1 */
```
`NUMTOYMINTERVAL(-0.7, 'month')` returns +1 month instead of -1.

2. **`to_date`/`to_timestamp`/`to_timestamp_tz` 3-arg versions skip parsing on empty nlsparam** (lines 874-877, 954-957, 1062-1071): the `ora_do_to_timestamp` call is the dangling body of `if (p && strlen(p) != 0)`. With an empty (but non-NULL) nlsparam the parse is skipped and the uninitialized stack `struct pg_tm tm` feeds `tm2timestamp` — garbage date or random error.

3. **`yminterval_in` dead range check** (`datatype/yminterval.c:1428`): `abs(tm->tm_mon) < 0` is always false, so a negative parsed month bypasses validation.

4. **`binary_double`/`binary_float` underflow early-return skips trailing garbage check** (`datatype/binary_double.c:203-204`, `datatype/binary_float.c:515-516`): on ERANGE underflow (`val == 0.0`) the functions return immediately, skipping the trailing-character check — `'1e-9999abc'::binary_double` silently returns 0 instead of an "invalid input syntax" error.

### Expected behavior/code

1. `interval_val = -1`
2. Call `ora_do_to_timestamp` unconditionally
3. `if (tm->tm_mon < 0 || tm->tm_mon > 11)`
4. Fall through to the trailing check and return the underflowed value only when the input is otherwise clean

Aias00:
## Summary

Fixes #1630. Four small datatype fixes in `contrib/ivorysql_ora`:

1. **`numtoyminterval` sign** (`builtin_functions/datetime_datatype_functions.c`): `NUMTOYMINTERVAL(-0.7, 'month')` returned +1; the negative-fraction rounding now yields -1.
2. **Dangling nlsparam `if`** (same file, 3 sites): `ora_do_to_timestamp` was the body of `if (p && strlen(p) != 0)`; with an empty nlsparam the parse was skipped and uninitialized `struct pg_tm tm` fed `tm2timestamp`. The call now always runs.
3. **`yminterval_in` dead check** (`datatype/yminterval.c`): `abs(tm->tm_mon) < 0` is always false; replaced with a real negative-month check.
4. **binary underflow early return** (`datatype/binary_double.c`, `datatype/binary_float.c`): on ERANGE underflow the functions returned 0/Infinity immediately, skipping the trailing-garbage check; `'1e-9999abc'::binary_double` silently returned 0. They now fall through to the trailing check and only return the underflowed value for clean input.

## Test plan

- All four translation units compile cleanly.
- Manual: `SELECT numtoyminterval(-0.7, 'month')` → -1; `SELECT to_date('2024-01-01', 'YYYY-MM-DD', '')` parses instead of producing garbage; `SELECT '1e-9999abc'::binary_double` raises "invalid input syntax".

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected negative year-to-month interval rounding for values between -0.5 and -1.
  - Improved validation of year-to-month intervals with invalid month values.
  - Fixed binary floating-point underflow handling so valid underflow results are processed correctly.
  - Improved consistency across timestamp conversion paths.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
@@ -853,16 +853,10 @@ to_oradate3(PG_FUNCTION_ARGS)
 {
 	text	   *date_txt = PG_GETARG_TEXT_P(0);
 	text	   *fmt = PG_GETARG_TEXT_P(1);
-	text	   *nlsparam = PG_GETARG_TEXT_P(2);
 	Oid			collid = PG_GET_COLLATION();
 	Timestamp	result;
 	struct pg_tm tm;
 	fsec_t		fsec;
-
-	/* TODO */
-	char	   *p;
-
-	p = text_to_cstring(nlsparam);
 
 #if 0
 	if (strlen(p) != 0)
@@ -934,14 +928,10 @@ to_oratimestamp3(PG_FUNCTION_ARGS)
 {
 	text	   *date_txt = PG_GETARG_TEXT_P(0);
 	text	   *fmt = PG_GETARG_TEXT_P(1);
-	text	   *nlsparam = PG_GETARG_TEXT_P(2);
 	Oid			collid = PG_GET_COLLATION();
 	Timestamp	result;
 	struct pg_tm tm;
 	fsec_t		fsec;
-	char	   *p;
-
-	p = text_to_cstring(nlsparam);
 
 	/*
 	 * if (strlen(p) != 0) ereport(WARNING,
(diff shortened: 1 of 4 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1630

Aias00:
Closing per the verification in #1630 - all four are real but need unusual boundary inputs; only the empty-nlsparam item is plausible in real use and its impact is a NULL return. Can be resubmitted as a focused PR if a maintainer wants it.

PR #1633, opened 2026-08-14
Fix regexp_like match options and UTL_FILE.GET_LINE truncation

### LINKED ISSUE #1632: regexp_like 'm'/'n' match options wrong; UTL_FILE.GET_LINE silently truncates long lines
### IvorySQL Version

### Current Behavior

1. **`regexp_like` match options** (`contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c:619-639`): the option parser handles `'n'` by setting `REG_NEWLINE` (which is the *anchor* semantics and does not make `.` match newline) and has no `case 'm'` at all, so `regexp_like(s, p, 'm')` raises "invalid option". Oracle semantics: `'n'` = dot matches newline (`REG_NLDOT`), `'m'` = multiline anchors (`REG_NLANCH`). Other regexp functions in the same file already support `'m'` via `ora_parse_re_flags`.

2. **`UTL_FILE.GET_LINE` silently truncates over-long lines** (`contrib/ivorysql_ora/src/builtin_packages/utl_file/utl_file.c:1323-1343`): the read loop stops at `max_linesize` and returns the truncated prefix with no error; the remainder stays in the file and appears as a spurious next line. Oracle raises VALUE_ERROR ("buffer is too short") when a line exceeds the buffer.

### Expected behavior/code

- `case 'n': out_flag |= REG_NLDOT;` and add `case 'm': out_flag |= REG_NLANCH;`
- After filling the buffer, peek one more character: if it is not EOF or newline, raise the buffer-too-short error (and `ungetc` the peeked char on success paths)

Aias00:
## Summary

Fixes #1632.

1. **`regexp_like` match options** (`contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c`):
   - `'n'` now sets `REG_NLDOT` (`.` matches newline) instead of `REG_NEWLINE` (which is anchor semantics)
   - the previously missing `'m'` option now sets `REG_NLANCH` (multiline `^`/`$` anchors), matching Oracle and the other regexp functions in the same file

2. **`UTL_FILE.GET_LINE`** (`contrib/ivorysql_ora/src/builtin_packages/utl_file/utl_file.c`): when a line fills `max_linesize` bytes, the function now peeks one more character; if the line continues (not EOF/`\n`/`\r\n`), it raises `VALUE_ERROR` ("buffer is too short") instead of silently returning the truncated prefix and leaking the remainder into the next call.

## Test plan

- Manual: `regexp_like('a\nb', 'a.b', 'n')` matches; `regexp_like('a\nb', '^b$', 'm')` matches; a GET_LINE on a file whose line exceeds max_linesize raises VALUE_ERROR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved regular expression matching so newline and multi-line options behave correctly.
  * File-reading operations now correctly detect lines exceeding the configured maximum length.
  * Lines ending exactly at the limit, including those with CRLF line endings, are accepted without error.
  * Oversized lines now raise the expected value error.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
@@ -622,7 +622,10 @@ ora_regexp_like(PG_FUNCTION_ARGS)
 				out_flag  |=  REG_ICASE;
 				break;
 			case 'n':
-				out_flag |= REG_NEWLINE;
+				out_flag |= REG_NLDOT;
+				break;
+			case 'm':
+				out_flag |= REG_NLANCH;
 				break;
 			case 'c':
 				out_flag  &=  ~REG_ICASE;
--- a/contrib/ivorysql_ora/src/builtin_packages/utl_file/utl_file.c
+++ b/contrib/ivorysql_ora/src/builtin_packages/utl_file/utl_file.c
@@ -1342,23 +1342,6 @@ get_line(FILE *fd, size_t max_linesize, int encoding, bool *iseof)
 		*bpt++ = c;
 	}
 
-	/* A line that does not fit in max_linesize must raise an error */
-	if (csize == max_linesize)
-	{
-		c = fgetc(fd);
-		if (c == '\r')
-		{
-			int	c2 = fgetc(fd);
-
-			if (c2 == '\n')
-				c = '\n';
-			else
-				c = '\r';
-		}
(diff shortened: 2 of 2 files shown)


Aias00:
Closing per the verification in #1632 - regexp_like 'm' works via pg_catalog core (wrong function was analyzed), and the GET_LINE change was already reverted as a pinned contract.

PR #1636, opened 2026-08-14
fix(dbms_lock): avoid reacquiring owned locks

### LINKED ISSUE #1598: DBMS_LOCK.REQUEST reacquires already-owned locks and leaves an unreleasable advisory lock
### IvorySQL Version

### OS Version (uname -a)

Not OS-specific.

Default DBMS_LOCK configuration.

### Current Behavior

Calling `DBMS_LOCK.REQUEST` twice for the same handle and lock mode acquires the underlying PostgreSQL session advisory lock twice. The DBMS_LOCK ownership hash stores only one entry and does not maintain a matching acquisition count.

A subsequent `DBMS_LOCK.RELEASE` removes the single ownership entry and releases only one acquisition. The session still holds the remaining advisory lock, but another RELEASE can no longer find the ownership entry and cannot release it.

The package defines return code `4` (`already_owned`), but the current request path does not use it for this case.

### Expected behavior/code

A request for a lock already owned by the same session should return `4` without reacquiring the underlying advisory lock. A single RELEASE should then fully release the owned lock.

### Step to reproduce

1. Allocate or select a DBMS_LOCK handle.
2. Request the same handle and mode twice in one session.
3. Release it once.
4. Inspect the session advisory locks or attempt to obtain the same lock from another session.
5. The second acquisition remains held even though DBMS_LOCK no longer records ownership.

The affected request and release bookkeeping is in `contrib/ivorysql_ora/src/builtin_packages/dbms_lock/dbms_lock.c`. A regression should cover REQUEST → REQUEST → RELEASE and verify that no advisory lock remains afterward.

yyqdbngt:
## Summary

`DBMS_LOCK.REQUEST` currently reacquires a PostgreSQL session advisory lock when the same session requests the same handle and mode again. The package ownership table records only one entry, so one `DBMS_LOCK.RELEASE` removes that entry but leaves one underlying advisory-lock acquisition held.

This change checks the package ownership table before trying to acquire the advisory lock and returns `4` (`already_owned`) for a duplicate same-mode request. The regression covers both shared and exclusive modes, including the lock state after a single release.

Fixes #1598

## Testing

I reproduced the behavior in GitHub Actions on `ubuntu-latest` with an IvorySQL assertion-enabled debug build. Both variants ran:

```text
make -C contrib/ivorysql_ora oracle-check ORA_REGRESS=dbms_lock
```

Unpatched `dbms_lock.c` from the PR merge base (`c3529ba4789ee220caefc858e58bbe6f1563ad14`):

```text
request rc=0
second request rc=0
mode=ShareLock
release rc=0
mode=ShareLock
request rc=0
second request rc=0
mode=ExclusiveLock
release rc=0
mode=ExclusiveLock
```

[Unpatched run and raw log](https://github.com/yyqdbngt/IvorySQL/actions/runs/32091376542/job/95574042224)

Patched source (same code tree as this PR):

```text
request rc=0
second request rc=4
mode=ShareLock
release rc=0
mode=not found
request rc=0
second request rc=4
mode=ExclusiveLock
release rc=0
mode=not found
```

The patched focused regression exited `0`.

[Patched run and raw log](https://github.com/yyqdbngt/IvorySQL/actions/runs/32090919761/job/95572750352)

Assisted-by: OpenAI:GPT-5

### DIFF
--- a/contrib/ivorysql_ora/expected/dbms_lock.out
+++ b/contrib/ivorysql_ora/expected/dbms_lock.out
@@ -18,10 +18,14 @@ begin
  v_rc := dbms_lock.request(v_handle, dbms_lock.s_mode, 0);  
  raise notice 'request rc=%', v_rc;
  assert v_rc = 0, format('DBMS_LOCK.request failed: rc=%s', v_rc);
+ v_rc := dbms_lock.request(v_handle, dbms_lock.s_mode, 0);
+ raise notice 'second request rc=%', v_rc;
+ assert v_rc = 4, format('DBMS_LOCK.request should report already owned: rc=%s', v_rc);
  insert into handle values(v_handle);
 end;
 /
 NOTICE:  request rc=0
+NOTICE:  second request rc=4
 do
 $$
 declare
--- a/contrib/ivorysql_ora/expected/dbms_lock.out
+++ b/contrib/ivorysql_ora/expected/dbms_lock.out
@@ -80,10 +80,14 @@ begin
  v_rc := dbms_lock.request(v_handle, dbms_lock.x_mode, 0);  
  raise notice 'request rc=%', v_rc;
  assert v_rc = 0, format('DBMS_LOCK.request failed: rc=%s', v_rc);
+ v_rc := dbms_lock.request(v_handle, dbms_lock.x_mode, 0);
+ raise notice 'second request rc=%', v_rc;
+ assert v_rc = 4, format('DBMS_LOCK.request should report already owned: rc=%s', v_rc);
  insert into handle values(v_handle);
 end;
 /
 NOTICE:  request rc=0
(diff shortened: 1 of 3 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1598

yyqdbngt:
I have now run the exact duplicate-request and single-release sequence in an Ubuntu IvorySQL build. The unpatched implementation returns `0` for the second same-mode request and still has the advisory lock after one release; the patched implementation returns `4`, and the lock is absent after one release for both shared and exclusive modes.

The commands, raw outputs, and run links are recorded in #1598 and in the updated PR description. I also rewrote the two commit messages to include the repository-required assistance disclosure without changing the code tree.

PR #1637, opened 2026-08-14
fix(utl_raw): preserve server encoding bytes

### LINKED ISSUE #1597: UTL_RAW.CAST_TO_RAW changes bytes in non-UTF8 databases
### IvorySQL Version

### OS Version (uname -a)

Encoding-dependent and not OS-specific.

A database initialized with a non-UTF8 server encoding, for example `LATIN1`.

### Current Behavior

`UTL_RAW.CAST_TO_RAW` is implemented with `convert_to(c::text, 'UTF8')`. As a result, text is always transcoded to UTF-8 before it is returned as raw bytes. In a LATIN1 database, for example, `é` is represented by the server as `E9`, but the function returns the UTF-8 byte sequence `C3A9` and changes the byte length.

### Expected behavior/code

`CAST_TO_RAW` should preserve the bytes of the input value in the current database encoding. It should not transcode the value or change its byte length.

### Step to reproduce

1. Initialize or create a database using `LATIN1` server encoding.
2. Install the `ivorysql_ora` extension.
3. Call `UTL_RAW.CAST_TO_RAW` with a value containing a LATIN1 non-ASCII character such as `é`.
4. Inspect the returned raw bytes; the current implementation returns the UTF-8 encoding rather than the original server-encoding bytes.

The implementation is in `contrib/ivorysql_ora/src/builtin_packages/utl_raw/utl_raw--1.0.sql`. Regression coverage should include non-UTF8, UTF8, NULL, and empty-string behavior.

btlqql:
## Summary

`UTL_RAW.CAST_TO_RAW` currently converts text to UTF-8 regardless of the database encoding. Use `convert_to` with the active server encoding so the returned raw bytes match Oracle-compatible server-encoding semantics.

The regression creates a LATIN1 database and verifies that `é` is represented as the single byte `e9`.

Fixes #1597

## Testing

- `git diff --check`
- reviewed the focused `utl_raw` expected-output update

The IvorySQL/PostgreSQL build toolchain is not available in this Windows environment, so the regression suite was not run locally.

Assisted-by: OpenAI:GPT-5

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `UTL_RAW.CAST_TO_RAW` to preserve text bytes according to the database’s configured character encoding.
  - Fixed conversion behavior for non-UTF-8 databases, including LATIN1.
  - Ensured accented and other non-ASCII characters are converted consistently without unintended byte changes.

- **Tests**
  - Added regression coverage verifying byte preservation when converting accented characters in a LATIN1 database.

### DIFF
--- a/contrib/ivorysql_ora/expected/utl_raw.out
+++ b/contrib/ivorysql_ora/expected/utl_raw.out
@@ -58,3 +58,19 @@ SELECT UTL_RAW.CAST_TO_RAW('\n');
  \x5c6e
 (1 row)
 
+-- CAST_TO_RAW preserves bytes in a non-UTF8 database
+CREATE DATABASE utl_raw_latin1
+    TEMPLATE template0 ENCODING 'LATIN1' LC_COLLATE 'C' LC_CTYPE 'C';
+\c utl_raw_latin1
+CREATE EXTENSION ivorysql_ora;
+SELECT pg_catalog.encode(
+    UTL_RAW.CAST_TO_RAW(pg_catalog.convert_from(pg_catalog.decode('e9', 'hex'), 'LATIN1')),
+    'hex');
+ encode 
+--------
+ e9
+(1 row)
+
+\c -
+DROP DATABASE utl_raw_latin1;
+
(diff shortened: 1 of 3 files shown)


NotHimmel:
LGTM

PR #1641, opened 2026-08-15
fix(oracle): enforce LONG typmod in characters

### LINKED ISSUE #1600: LONG(n) typmod is applied as bytes instead of characters
### IvorySQL Version

### OS Version (uname -a)

Not OS-specific; visible with a multibyte server encoding such as UTF8.

A database using UTF8 or another multibyte server encoding.

### Current Behavior

The typmod input for `LONG(n)` defines `n` as a number of characters, but `orachar_to_long_with_typmod()` compares `n` with `VARSIZE_ANY_EXHDR()`, which is a byte length. Explicit narrowing also truncates by setting the byte length directly to `n` before copying.

Consequently, valid input containing exactly `n` multibyte characters can be rejected as too long during implicit assignment. Explicit casts can truncate in the middle of a multibyte character and produce a text value that is not valid in the server encoding.

### Expected behavior/code

`LONG(n)` should enforce and truncate its typmod in characters, not bytes. Multibyte characters that fit within the character limit should be accepted, and explicit casts should retain at most `n` complete characters. PostgreSQL's varchar typmod path uses `pg_mbcharcliplen` for the same character-count semantics.

### Step to reproduce

In a UTF8 database:

1. Create a column of type `LONG(2)` and insert a two-character value such as `中文`; the current byte-length check rejects it even though it contains exactly two characters.
2. Explicitly cast a value such as `中文A` to `LONG(2)`; the current path truncates to two bytes rather than two complete characters.
3. Inspect `length` and `octet_length` or pass the result to another text operation.

The affected conversion is in `contrib/ivorysql_ora/src/datatype/raw_long.c`. Regression coverage should include implicit acceptance of exactly `n` multibyte characters, implicit rejection of more than `n` characters, and explicit truncation to `n` complete characters.

yyqdbngt:
## Summary

- apply `LONG(n)` typmods using character counts rather than encoded byte counts
- clip explicit casts with `pg_mbcharcliplen()` so multibyte characters remain intact
- add UTF-8 regression coverage for exact implicit input, oversized implicit input, and explicit truncation

## Root cause

`oralongtypmodin()` stores the declared length as a character count, but `orachar_to_long_with_typmod()` compared and truncated the value using its encoded byte length. In UTF-8 this rejected valid multibyte input and truncated explicit casts at a byte boundary.

## Reproduction and validation

The exact commands, SQL, and unpatched output are recorded in https://github.com/IvorySQL/IvorySQL/issues/1600#issuecomment-5322691835.

- unpatched baseline `9df72e5132`, Ubuntu 24.04.4: https://github.com/yyqdbngt/IvorySQL/actions/runs/32090511214
  - `ora_raw_long` was the only failing `ivorysql_ora` regression test
  - `INSERT ... VALUES ('中文')` incorrectly returned `value too long for type long(2)`
  - the explicit `LONG(2)` comparison returned `f`
- patched head `b1e02d361d`, same repository workflow: https://github.com/IvorySQL/IvorySQL/actions/runs/31863038178
  - `ok 7 - ora_raw_long`
  - full `oracle_regression` job passed

These runs used GitHub-hosted Ubuntu runners; they were not run on the Windows development host.

Closes #1600

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_raw_long.out
+++ b/contrib/ivorysql_ora/expected/ora_raw_long.out
@@ -175,6 +175,24 @@ SELECT * FROM LONG_TEXT ORDER BY II DESC;
 (3 rows)
 
 DROP TABLE LONG_TEXT;
+-- LONG typmods count characters, not bytes
+CREATE TABLE LONG_UTF8_TEST(a LONG(2));
+INSERT INTO LONG_UTF8_TEST VALUES ('中文');
+INSERT INTO LONG_UTF8_TEST VALUES ('中文A');
+ERROR:  value too long for type long(2)
+SELECT length(a) = 2 AND octet_length(a) = 6 FROM LONG_UTF8_TEST;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT '中文A'::LONG(2) = '中文'::text;
+ ?column? 
+----------
+ t
+(1 row)
+
+DROP TABLE LONG_UTF8_TEST;
 -- rawtohex
 SELECT sys.rawtohex('\xDEADBEEF'::bytea);
  rawtohex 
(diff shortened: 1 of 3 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1600

NotHimmel:
LONG(n) is not valid syntax in Oracle. Closing as invalid.

PR #1642, opened 2026-08-15
fix(oracle): preserve characters in VARCHAR2 byte casts

### LINKED ISSUE #1639: Explicit VARCHAR2(n BYTE) casts can split multibyte characters
### IvorySQL Version

PR base `c3529ba4789ee220caefc858e58bbe6f1563ad14` (verified with IvorySQL `19devel`).

### Current Behavior

An explicit cast to `VARCHAR2(n BYTE)` assigns `n` directly to the copied byte length. In a multibyte server encoding, the boundary can fall inside a character and produce a value that is invalid in the server encoding.

On a UTF8 database, casting the two-character value `中文` to `VARCHAR2(4 BYTE)` copies one complete character plus the first byte of the second character.

### Steps to Reproduce

Start an Oracle-mode instance, connect with `psql`, and run:

```sql
SHOW server_encoding;
SHOW ivorysql.compatible_mode;
SELECT lengthb(CAST('中文' AS VARCHAR2(4 BYTE))) AS byte_length;
SELECT CAST('中文' AS VARCHAR2(4 BYTE)) = '中' AS complete_character;
SELECT encode(convert_to(CAST('中文' AS VARCHAR2(4 BYTE)), 'UTF8'), 'hex') AS utf8_hex;
```

Verified on GitHub Actions `ubuntu-24.04`, UTF8, with `ivorysql.compatible_mode = oracle`. With the original implementation:

- `byte_length` is `4`
- `complete_character` is `f`
- `convert_to(..., 'UTF8')` fails with `invalid byte sequence for encoding UTF8: 0xe6`

Reproduction run: https://github.com/yyqdbngt/IvorySQL/actions/runs/32091868545

### Expected Behavior

The cast should retain the longest complete-character prefix that fits in four bytes. For this input, the result should be `中`, `lengthb()` should return `3`, and conversion to UTF8 should produce the valid bytes `e4b8ad`.

### Additional Context

The explicit-cast path is `oravarcharbyte()` in `contrib/ivorysql_ora/src/datatype/oravarcharbyte.c`. PostgreSQL provides `pg_mbcliplen()` for clipping to a byte limit without splitting a character.

yyqdbngt:
## Summary

- clip explicit `VARCHAR2(n BYTE)` casts at the last complete-character boundary
- preserve the byte limit without returning invalid server-encoding data
- use Oracle `CAST(... AS VARCHAR2(n BYTE))` syntax in the regression coverage

## Root cause

The explicit-cast path assigned the typmod byte limit directly to the copy length. In UTF8 this can stop inside a multibyte character. The fix uses `pg_mbcliplen()` to retain the longest complete-character prefix within the byte limit.

## Reproduction

Verified on GitHub Actions `ubuntu-24.04`, IvorySQL `19devel`, UTF8, with `ivorysql.compatible_mode = oracle`.

```sql
SELECT lengthb(CAST('中文' AS VARCHAR2(4 BYTE))) AS byte_length;
SELECT CAST('中文' AS VARCHAR2(4 BYTE)) = '中' AS complete_character;
SELECT encode(convert_to(CAST('中文' AS VARCHAR2(4 BYTE)), 'UTF8'), 'hex') AS utf8_hex;
```

With the original `oravarcharbyte()` implementation:

- `byte_length` is `4`
- `complete_character` is `f`
- `convert_to(..., 'UTF8')` fails with `invalid byte sequence for encoding UTF8: 0xe6`

With this patch:

- `byte_length` is `3`
- `complete_character` is `t`
- `utf8_hex` is `e4b8ad`

## Validation

- Original implementation reproduction: https://github.com/yyqdbngt/IvorySQL/actions/runs/32091868545
- Patched implementation plus focused `ora_character` Oracle regression: https://github.com/yyqdbngt/IvorySQL/actions/runs/32093105278
- `git diff --check`

Closes #1639

Assisted-by: opencode:qwen/qwen3.8-max-preview

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `VARCHAR2(n BYTE)` casts to truncate multibyte text only at complete character boundaries.
  * Preserved expected truncation behavior for ASCII text and values already within the specified byte limit.
* **Tests**
  * Added regression coverage for multibyte and ASCII `VARCHAR2(n BYTE)` conversions, including byte-length validation.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character.out
+++ b/contrib/ivorysql_ora/expected/ora_character.out
@@ -27,20 +27,20 @@ CREATE TABLE TEST_ORACHAR(a char(5 byte), b char(5 char));
 CREATE TABLE TEST_ORAVARCHAR(a varchar2(5 char), b varchar2(5 byte));
 -- Explicit VARCHAR2(n BYTE) casts retain only complete characters
 SELECT '中文'::varchar2(6 byte) = '中文'
-       AND octet_length('中文'::varchar2(6 byte)) = 6;
+       AND lengthb('中文'::varchar2(6 byte)) = 6;
  ?column? 
 ----------
  t
 (1 row)
 
 SELECT '中文'::varchar2(5 byte) = '中'
-       AND octet_length('中文'::varchar2(5 byte)) = 3;
+       AND lengthb('中文'::varchar2(5 byte)) = 3;
  ?column? 
 ----------
  t
 (1 row)
 
-SELECT coalesce(octet_length('中文'::varchar2(2 byte)), 0) = 0;
+SELECT coalesce(lengthb('中文'::varchar2(2 byte)), 0) = 0;
  ?column? 
 ----------
  t
--- a/contrib/ivorysql_ora/expected/ora_character.out
+++ b/contrib/ivorysql_ora/expected/ora_character.out
@@ -40,6 +40,13 @@ SELECT '中文'::varchar2(5 byte) = '中'
  t
 (1 row)
 
+SELECT '中文'::varchar2(2 byte) = ''
+       AND octet_length('中文'::varchar2(2 byte)) = 0;
+ ?column? 
+----------
+ t
+(1 row)
+
 SELECT 'abcdef'::varchar2(4 byte) = 'abcd';
  ?column? 
 ----------
--- a/contrib/ivorysql_ora/expected/ora_character.out
+++ b/contrib/ivorysql_ora/expected/ora_character.out
@@ -25,6 +25,33 @@ LINE 1: CREATE TABLE TEST_ORACHAR(a varchar2(32768 char));
 -- valid range
 CREATE TABLE TEST_ORACHAR(a char(5 byte), b char(5 char));
 CREATE TABLE TEST_ORAVARCHAR(a varchar2(5 char), b varchar2(5 byte));
+-- Explicit VARCHAR2(n BYTE) casts retain only complete characters
+SELECT '中文'::varchar2(6 byte) = '中文'
+       AND octet_length('中文'::varchar2(6 byte)) = 6;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT '中文'::varchar2(5 byte) = '中'
+       AND octet_length('中文'::varchar2(5 byte)) = 3;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT 'abcdef'::varchar2(4 byte) = 'abcd';
+ ?column? 
+----------
+ t
+(1 row)
+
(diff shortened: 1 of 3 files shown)


NotHimmel:
Thanks for the PR.Oracle regression tests should use `lengthb()` instead of PostgreSQL's `octet_length()`.

yyqdbngt:
@NotHimmel Done. Replaced `octet_length()` with `lengthb()` in both the regression script and the expected output. Since both return the byte count, the expected values are unchanged. Applied in c1c8988.

NotHimmel:
@yyqdbngt Could you share the exact reproduction steps for the lengthb()/varchar2 bug?

Also, the statement used in the regression test is not valid Oracle syntax:

    SELECT '中文'::varchar2(6 byte) = '中文'
       AND lengthb('中文'::varchar2(6 byte)) = 6;

Thanks!

yyqdbngt:
Thanks for pointing that out. The `::` form was PostgreSQL-specific. I updated the regression statements in 2a0f3c6330 to use Oracle `CAST(... AS VARCHAR2(n BYTE))` syntax.

I reproduced the issue on a GitHub Actions Ubuntu 24.04 runner using the PR base and the original `oravarcharbyte()` implementation. The database was UTF8 and the session was explicitly switched to Oracle mode.

Build and server setup:

```sh
./configure --prefix=$PWD/inst --enable-cassert --enable-debug --with-uuid=e2fs
make -j2
make install
export PATH=$PWD/inst/bin:$PATH
initdb -D $PWD/data -m oracle --no-sync
pg_ctl -D $PWD/data -l $PWD/server.log start
```

Reproduction SQL:

```sql
SHOW server_encoding;
SHOW ivorysql.compatible_mode;
SELECT lengthb(CAST('中文' AS VARCHAR2(4 BYTE))) AS byte_length;
SELECT CAST('中文' AS VARCHAR2(4 BYTE)) = '中' AS complete_character;
SELECT encode(convert_to(CAST('中文' AS VARCHAR2(4 BYTE)), 'UTF8'), 'hex') AS utf8_hex;
```

The original implementation returned `byte_length = 4`, the comparison with `中` returned `f`, and `convert_to(..., 'UTF8')` failed with:

```
ERROR: invalid byte sequence for encoding UTF8: 0xe6
```

With this patch, the same commands return `byte_length = 3`, `complete_character = t`, and `utf8_hex = e4b8ad`. I also ran:

```sh
make -C contrib/ivorysql_ora oracle-check ORA_REGRESS=ora_character
```

The focused Oracle regression passed.

Baseline run: https://github.com/yyqdbngt/IvorySQL/actions/runs/32091868545
Patched run and regression: https://github.com/yyqdbngt/IvorySQL/actions/runs/32093105278

PR #1643, opened 2026-08-15
fix(oracle): honor nondeterministic CHAR byte collations

### LINKED ISSUE #1599: CHAR(n BYTE) equality and hashing ignore nondeterministic collations
### IvorySQL Version

### OS Version (uname -a)

Any build with ICU collation support.

Built with ICU support and tested with an ICU nondeterministic collation, such as a case-insensitive collation.

### Current Behavior

The comparison function for Oracle-compatible `CHAR(n BYTE)` values uses the active collation for ordering, but `oracharbyteeq` and `oracharbytene` always use byte-wise `memcmp`. `oracharbytehash` likewise always hashes the original bytes.

Under a nondeterministic ICU collation, values such as `a` and `A` can compare equal according to the collation while the equality operator reports them as different. Their hashes also differ, violating the requirement that equal values have equal hash values. This can produce inconsistent equality, grouping, DISTINCT, and hash-index behavior.

### Expected behavior/code

For deterministic collations, the existing byte-wise fast path should remain. For nondeterministic collations, equality and hashing should follow the collation, as PostgreSQL's current `bpchar` implementation does, so ordering, equality, and hash semantics remain consistent.

### Step to reproduce

1. Create an ICU nondeterministic case-insensitive collation.
2. Create Oracle-compatible `CHAR(1 BYTE)` values using that collation.
3. Compare `a` and `A`, then exercise DISTINCT or GROUP BY and a hash index.
4. Observe that ordering follows the collation while equality and hashing follow the original bytes.

The affected functions are in `contrib/ivorysql_ora/src/datatype/oracharbyte.c`. PostgreSQL's `bpchareq`, `bpcharne`, and `hashbpchar` provide the deterministic/nondeterministic collation pattern. Regression coverage should include equality, inequality, grouping or DISTINCT, and hash-index lookup when ICU is available.

btlqql:
## Summary

- use the active collation for `CHAR(n BYTE)` equality and inequality when it is nondeterministic
- hash nondeterministic values through their collation sort keys so equal values produce equal hashes
- retain the bytewise fast path for deterministic collations
- conditionally cover ICU equality, inequality, DISTINCT/hash aggregation, and hash-index lookup

## Root cause

Ordering already used `varstr_cmp()`, while equality and hashing always compared or hashed the original bytes. Under nondeterministic collations those definitions could disagree.

## Validation

- implementation follows the current core `bpchar` deterministic/nondeterministic pattern
- `git diff --check`
- added focused `ora_character` regression assertions that safely skip when ICU is unavailable
- full regression execution was not available on the Windows-only local host; repository CI is expected to run the build and regression suite

Closes #1599

## Summary by CodeRabbit

* **New Features**
  * Added collation-aware comparisons and hashing for `CHAR(n BYTE)` values.
  * Case-insensitive collations now support equality checks, `DISTINCT`, grouping, ranking, set operations, and hash-index lookups.
  * Differently cased equivalent values are correctly treated as equal across these operations.

* **Tests**
  * Added regression coverage for ICU collation behavior, including comparisons, hashing, duplicate elimination, and indexed lookups.
  * Tests skip automatically when ICU collation support is unavailable.

## Deterministic collation regression follow-up

Added an ICU collation with the same `und-u-ks-level2` locale and `deterministic = true`. The regression asserts that CHAR(1 BYTE) equality between `a` and `A` is false and inequality is true, protecting both deterministic comparison paths. The temporary collation is removed, and the case remains inside the existing ICU availability guard.

For this follow-up, `git diff --check` passed and the SQL DO block was checked to match its expected echo exactly. No local database regression was run: this Windows environment has no runnable IvorySQL test installation. Earlier runtime results apply to the preceding revision; the new regression still needs CI validation.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character.out
+++ b/contrib/ivorysql_ora/expected/ora_character.out
@@ -411,3 +411,62 @@ explain (costs off) SELECT * FROM TEST_ORAVARCHAR WHERE a='111';
 -- drop table
 DROP TABLE TEST_ORACHAR;
 DROP TABLE TEST_ORAVARCHAR;
+-- Nondeterministic collations keep CHAR(n BYTE) equality and hashing aligned
+DO $$
+DECLARE
+    result boolean;
+    matches integer;
+BEGIN
+    IF EXISTS (SELECT 1 FROM pg_collation WHERE collprovider = 'i') THEN
+        EXECUTE 'CREATE COLLATION ora_char_byte_ci '
+                '(provider = icu, locale = ''und-u-ks-level2'', deterministic = false)';
+
+        EXECUTE 'SELECT ''a''::char(1 byte) COLLATE ora_char_byte_ci = '
+                '''A''::char(1 byte) COLLATE ora_char_byte_ci'
+           INTO result;
+        IF NOT result THEN
+            RAISE EXCEPTION 'CHAR(n BYTE) equality ignored its nondeterministic collation';
+        END IF;
+
+        EXECUTE 'SELECT NOT (''a''::char(1 byte) COLLATE ora_char_byte_ci <> '
+                '''A''::char(1 byte) COLLATE ora_char_byte_ci)'
+           INTO result;
+        IF NOT result THEN
+            RAISE EXCEPTION 'CHAR(n BYTE) inequality ignored its nondeterministic collation';
+        END IF;
+
+        EXECUTE 'SELECT sys.oracharbytehash('
+                '''a''::char(1 byte) COLLATE ora_char_byte_ci) = '
+                'sys.oracharbytehash('
+                '''A''::char(1 byte) COLLATE ora_char_byte_ci)'
+           INTO result;
+        IF NOT result THEN
+            RAISE EXCEPTION 'equal CHAR(n BYTE) values produced different hashes';
+        END IF;
+
+        EXECUTE 'CREATE TEMP TABLE ora_char_byte_ci_test '
+                '(v char(1 byte) COLLATE ora_char_byte_ci)';
+        EXECUTE 'INSERT INTO ora_char_byte_ci_test VALUES (''a''), (''A'')';
+        PERFORM set_config('enable_sort', 'off', true);
+        EXECUTE 'SELECT count(DISTINCT v) FROM ora_char_byte_ci_test'
+           INTO matches;
+        IF matches <> 1 THEN
+            RAISE EXCEPTION 'DISTINCT did not combine equal CHAR(n BYTE) values';
+        END IF;
+
+        EXECUTE 'TRUNCATE ora_char_byte_ci_test';
+        EXECUTE 'INSERT INTO ora_char_byte_ci_test VALUES (''a'')';
+        EXECUTE 'CREATE INDEX ora_char_byte_ci_hash_idx '
+                'ON ora_char_byte_ci_test USING hash (v)';
+        PERFORM set_config('enable_seqscan', 'off', true);
+        EXECUTE 'SELECT count(*) FROM ora_char_byte_ci_test WHERE v = ''A'''
+           INTO matches;
+        IF matches <> 1 THEN
+            RAISE EXCEPTION 'hash index lookup missed an equal CHAR(n BYTE) value';
+        END IF;
+
+        EXECUTE 'DROP TABLE ora_char_byte_ci_test';
+        EXECUTE 'DROP COLLATION ora_char_byte_ci';
+    END IF;
+END
+$$;
--- a/contrib/ivorysql_ora/sql/ora_character.sql
+++ b/contrib/ivorysql_ora/sql/ora_character.sql
@@ -177,3 +177,63 @@ explain (costs off) SELECT * FROM TEST_ORAVARCHAR WHERE a='111';
 -- drop table
 DROP TABLE TEST_ORACHAR;
 DROP TABLE TEST_ORAVARCHAR;
+
+-- Nondeterministic collations keep CHAR(n BYTE) equality and hashing aligned
+DO $$
+DECLARE
+    result boolean;
+    matches integer;
+BEGIN
+    IF EXISTS (SELECT 1 FROM pg_collation WHERE collprovider = 'i') THEN
+        EXECUTE 'CREATE COLLATION ora_char_byte_ci '
+                '(provider = icu, locale = ''und-u-ks-level2'', deterministic = false)';
+
+        EXECUTE 'SELECT ''a''::char(1 byte) COLLATE ora_char_byte_ci = '
+                '''A''::char(1 byte) COLLATE ora_char_byte_ci'
+           INTO result;
+        IF NOT result THEN
+            RAISE EXCEPTION 'CHAR(n BYTE) equality ignored its nondeterministic collation';
+        END IF;
+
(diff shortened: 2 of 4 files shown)


NotHimmel:
Thanks for your PR!

NotHimmel:
Could you please provide exact reproduction steps and sql script(if needed) and real bug output? Thanks!

yyqdbngt:
Here are the exact steps, the SQL script and the real bug output.

## Build (Ubuntu 22.04, gcc 12.3, ICU 70.1)

```sh
git checkout bef143b4ab35f8ffd826f69753660cb9ed810c7f
./configure --prefix=$HOME/ivy-icu-inst --with-uuid=e2fs --without-readline \
            --without-zlib --with-icu --without-libxml
make -j24 && make install
make -C contrib/ivorysql_ora -j24 && make -C contrib/ivorysql_ora install
initdb -D $HOME/ivy-icu-data -USYSTEM -moracle
printf "port = 1601\nunix_socket_directories = '%s'\n" "$HOME/ivy-icu-sock" >> $HOME/ivy-icu-data/postgresql.conf
pg_ctl -D $HOME/ivy-icu-data -l pg.log -w start
psql -X -h $HOME/ivy-icu-sock -p 1601 -U system -d postgres -f repro.sql
```

A nondeterministic collation needs ICU (`--with-icu`); the repro also needs Oracle mode, which the regression harness sets for itself — in a hand-made cluster it has to be set per session (`SET ivorysql.compatible_mode = oracle;`).

## `repro.sql`

```sql
\pset pager off

CREATE COLLATION ora_char_byte_ci (provider = icu, locale = 'und-u-ks-level2', deterministic = false);

-- ordering / ranking
SELECT v, rank() OVER (ORDER BY v) AS rnk
  FROM (SELECT 'a'::char(1 byte) COLLATE ora_char_byte_ci AS v
        UNION ALL
        SELECT 'A'::char(1 byte) COLLATE ora_char_byte_ci) s
 ORDER BY v;

-- equality and inequality
SELECT 'a'::char(1 byte) COLLATE ora_char_byte_ci =  'A'::char(1 byte) COLLATE ora_char_byte_ci AS eq FROM dual;
SELECT 'a'::char(1 byte) COLLATE ora_char_byte_ci <> 'A'::char(1 byte) COLLATE ora_char_byte_ci AS ne FROM dual;

-- hashing
SELECT sys.oracharbytehash('a'::char(1 byte) COLLATE ora_char_byte_ci) =
       sys.oracharbytehash('A'::char(1 byte) COLLATE ora_char_byte_ci) AS hashes_equal FROM dual;

-- DISTINCT / GROUP BY / hash aggregation
CREATE TEMP TABLE t1643 (v char(1 byte) COLLATE ora_char_byte_ci);
INSERT INTO t1643 VALUES ('a'), ('A');
SET enable_sort = off;
SELECT count(DISTINCT v) AS hash_distinct FROM t1643;
SELECT v, count(*) AS n FROM t1643 GROUP BY v ORDER BY v;
RESET enable_sort;

-- hash index lookup
CREATE INDEX t1643_hash_idx ON t1643 USING hash (v);
SET enable_seqscan = off;
EXPLAIN (COSTS OFF) SELECT v FROM t1643 WHERE v = 'A';
SELECT count(*) AS hash_index_hits FROM t1643 WHERE v = 'A';
RESET enable_seqscan;
DROP TABLE t1643;
DROP COLLATION ora_char_byte_ci;
```

## Real output, before the fix — `master` (`ivorysql_ora.so` md5 `e7ac443a5150dd401196c3358700fa82`)

```
 v | rnk 
---+-----
 a |   1
 A |   2
(2 rows)

 eq      ne      hashes_equal
----    ----    --------------
 f       t       f

 hash_distinct        v | n 
---------------     ---+---
             2        a | 1
                      A | 1
                    (2 rows)

                QUERY PLAN                
------------------------------------------
 Index Scan using t1643_hash_idx on t1643
   Index Cond: (v = 'A'::oracharbyte)

 hash_index_hits 
-----------------
               1
```

(The three scalar queries printed their own tables — `eq`, `ne` and `hashes_equal` are `f`, `t`, `f`; above they are shown together for readability. The `rank()` output is verbatim.)

So the two values are distinct for `=`, `<>`, `DISTINCT`, `GROUP BY` and the hash index, and `rank()` also puts them in different positions — while the collation says they are equal.

## Real output, with the fix — the PR head (`ivorysql_ora.so` md5 `7eaf3d48609dd86dba7bbca359cbbea9`)

```
 v | rnk 
---+-----
 a |   1
 A |   1
(2 rows)

 eq      ne      hashes_equal
----    ----    --------------
 t       f       t

 hash_distinct        v | n 
---------------     ---+---
             1        a | 2
                    (1 row)

                QUERY PLAN                
------------------------------------------
 Index Scan using t1643_hash_idx on t1643
   Index Cond: (v = 'A'::oracharbyte)

 hash_index_hits 
-----------------
               2
```

`make -C contrib/ivorysql_ora oracle-check` on the same build: **28 ok / 1 not ok** in both runs (`ok 1 - ora_character` included), the single failure being `ora_xml_functions`, which is unrelated — I configured `--without-libxml`.

One thing to flag: the branch is a few weeks behind `master`, and master has since touched `contrib/ivorysql_ora/{sql,expected}/ora_character.*` and `src/oracle_test/regress/expected/collate.icu.utf8.out`, so I will rebase and regenerate those expectations before this can merge.

btlqql:
Rebased onto `master` (`03b24b1c46`) — head is now `d571bc023e5104ecfebe74760aec239c8b6cb0ae`, and the PR is `MERGEABLE` again.

The rebase brought in master's newer `ora_character.*` and `collate.icu.utf8` content, so I re-ran both suites on the rebased tree (Ubuntu 22.04, gcc 12.3, ICU 70.1, `--with-uuid=e2fs --without-readline --without-zlib --with-icu --without-libxml`, full `make -j32`):

**`make -C contrib/ivorysql_ora oracle-check`** — 28 ok / 1 not ok, the only failure being `ora_xml_functions` (unrelated: this build is `--without-libxml`). `diff expected/ora_character.out results/ora_character.out` is **empty**, so the harness really produces the expectations in the PR — including `ok 1 - ora_character`.

**`make -C src/oracle_test/regress oracle-check-tests TESTS='collate.icu.utf8'`** — `ok 1 - collate.icu.utf8`, `# All 1 tests passed.`, and `diff expected/collate.icu.utf8.out results/collate.icu.utf8.out` is **empty** (both files md5 `a01cb6abf51eb86815e1cc50874b0a42`). The nondeterministic bpchar section really now reports:

```
SELECT x FROM test3bpci WHERE x = 'abc';
  x  
-----
 abc
 ABC
(2 rows)

SELECT x FROM test1bpci INTERSECT SELECT x FROM test2bpci ORDER BY x;
  x  
-----
 abc
 ghi
(2 rows)
```

and the previously written expectations for `rank()`, `count(DISTINCT x)` and `GROUP BY` match the real output byte for byte. So no expectation had to be rewritten after the rebase — nothing here is hand-written guesswork any more.

The standalone repro from my earlier comment also still behaves the same on the rebased build (`ivorysql_ora.so` md5 `7eaf3d48609dd86dba7bbca359cbbea9`): `a = A` → `t`, `<>` → `f`, `oracharbytehash` equal → `t`, `count(DISTINCT v)` with sorting off → `1`, hash index `WHERE v = 'A'` → `2` rows, `rank()` → both `1`.

PR #1644, opened 2026-08-15
fix(dbms_session): validate clear namespaces consistently

### LINKED ISSUE #1640: DBMS_SESSION clear operations skip namespace validation in empty sessions
### IvorySQL Version

Current `master` (`c3529ba4789ee220caefc858e58bbe6f1563ad14`).

### Current Behavior

`DBMS_SESSION.CLEAR_CONTEXT` and `CLEAR_ALL_CONTEXT` return immediately when the session context hash has not been initialized. This happens before validating the namespace length, so a namespace of 256 bytes or more is silently accepted in a fresh or reset session.

The same namespace is rejected once context state exists, making parameter validation depend on unrelated session history.

### Expected Behavior

Clear operations should validate the namespace before treating an empty context store as a no-op. Names of 256 bytes or more should consistently report the existing `namespace too long` error.

### Steps to Reproduce

```sql
CALL dbms_session.reset_package();
CALL dbms_session.clear_context(repeat('n', 256));
CALL dbms_session.clear_all_context(repeat('n', 256));
```

Both clear calls should fail with `DBMS_SESSION namespace too long (max 255 bytes)` even though no context entries exist.

### Additional Context

The early returns are in `contrib/ivorysql_ora/src/builtin_packages/dbms_session/dbms_session.c`.

btlqql:
## Summary

- centralize DBMS_SESSION context-name length validation
- validate clear-operation arguments before treating an empty context store as a no-op
- cover the valid 255-byte boundary and consistent 256-byte errors for set, single-attribute clear, and whole-namespace clear operations

## Root cause

The clear functions returned before converting or validating their namespace when the context hash was absent or empty, so the same invalid argument behaved differently depending on session history.

## Validation

- `git diff --check`
- added focused `dbms_session` regression cases
- full regression execution was not available on the Windows-only local host; repository CI is expected to run the build and regression suite

Closes #1640

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session context clearing for valid and missing namespaces.
  * Corrected attribute clearing when no session context store exists.
  * Standardized validation for namespace and attribute name lengths.
  * Names up to 255 bytes are supported; 256-byte names are rejected consistently.
  * Namespace-wide clearing now removes multiple attributes correctly at the 255-byte boundary.

* **Tests**
  * Expanded coverage for setting, reading, clearing, and clearing all session context values, including boundary lengths and uninitialized context stores.

### DIFF
--- a/contrib/ivorysql_ora/expected/dbms_session.out
+++ b/contrib/ivorysql_ora/expected/dbms_session.out
@@ -265,17 +265,49 @@ select sys_context('rp_ns', 'key') as ctx_after_reset;
 (1 row)
 
 DROP PACKAGE rp_pkg;
--- Clear operations validate oversized namespaces even when no context exists
-call dbms_session.clear_context('rp_ns');
-call dbms_session.clear_context(repeat('n', 255));
-call dbms_session.clear_all_context(repeat('n', 255));
+-- Reconnect so oversized clears exercise the uninitialized context store
+\c -
 \set VERBOSITY terse
-call dbms_session.set_context(repeat('n', 256), 'attr', 'value');
-ERROR:  DBMS_SESSION namespace too long (max 255 bytes)
 call dbms_session.clear_context(repeat('n', 256), 'attr');
 ERROR:  DBMS_SESSION namespace too long (max 255 bytes)
 call dbms_session.clear_context(repeat('n', 256));
 ERROR:  DBMS_SESSION namespace too long (max 255 bytes)
 call dbms_session.clear_all_context(repeat('n', 256));
+ERROR:  DBMS_SESSION namespace too long (max 255 bytes)
+\set VERBOSITY default
+-- The 255-byte namespace and attribute boundaries remain usable
+call dbms_session.set_context(repeat('n', 255), repeat('a', 255), 'value');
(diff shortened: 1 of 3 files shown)


NotHimmel:
Thanks for your PR!

hs-liuxh:
Thanks for the contribution. After review, I agree this fix is valuable for keeping `set_context` and `clear_context` behavior consistent. A couple of small things to address:

1. CodeRabbit's two review comments are valid — they strengthen coverage of the boundary cases. Please add the test cases as suggested there.
2. Your second commit resets the context store by reconnecting with `\c`. That's too indirect and heavier than needed: it also discards the session state, so any test cases added later can't build on it. Please use `DISCARD PACKAGES` instead of `\c`.

btlqql:
Thanks for the review. I replaced the reconnect with the repository-supported same-session reset syntax, \DISCARD PACKAGE\ (the Oracle grammar accepts the singular form; \DISCARD PACKAGES\ caused the oracle_regression syntax error). The requested initialized/uninitialized and namespace-wide boundary coverage is included. The new external-contributor workflow runs are awaiting approval; could you please approve them when convenient?

PR #1645, opened 2026-08-15
Fix duplicate sequence SESSION and GLOBAL options

### LINKED ISSUE #1638: Duplicate SESSION/GLOBAL sequence options are accepted
### Current behavior

Duplicate `SESSION` or `GLOBAL` options are accepted by Oracle-compatible `CREATE SEQUENCE` statements:

```sql
CREATE SEQUENCE seq_duplicate_session SESSION SESSION;
CREATE SEQUENCE seq_duplicate_global GLOBAL GLOBAL;
```

`init_params()` records these options in `session_flag` and `global_flag`, but the duplicate checks use an unused `session` pointer instead of checking the flag corresponding to the current option. The pointer is initialized to `NULL` and never assigned, so a repeated option is not detected.

### Expected behavior

Both statements should fail with:

```text
duplicate SESSION or GLOBAL specifications
```

Conflicting `SESSION GLOBAL` and `GLOBAL SESSION` combinations should continue to fail with the same error.

### Proposed fix

Check `session_flag || global_flag` for both options, remove the unused `session` variable, and add Oracle regression coverage for duplicate and conflicting forms.

123123213weqw:
## What changed
- reject repeated `SESSION` and `GLOBAL` sequence options
- use the existing option flags for both duplicate and conflict checks
- remove the unused `session` pointer
- add regression coverage for duplicate and conflicting forms

## Root cause
The duplicate checks used a never-assigned `DefElem *session` and only checked the flag for the opposite option. As a result, `SESSION SESSION` and `GLOBAL GLOBAL` were accepted while mixed forms were rejected.

## Validation
- `git diff --check`
- remote `make -j16`
- remote `make oracle-check-world`: the `sequence` test passed; the temporary environment's unrelated `xml` test failed because XML support was unavailable

Closes #1638

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved validation for sequence options to consistently reject duplicate or conflicting `SESSION` and `GLOBAL` specifications.
  - Added clearer error handling for invalid combinations regardless of option order.

- **Tests**
  - Added regression coverage for repeated and conflicting sequence options, including expected parser errors and positions.

### DIFF
--- a/src/backend/commands/sequence.c
+++ b/src/backend/commands/sequence.c
@@ -1572,7 +1572,6 @@ init_params(ParseState *pstate, List *options, bool for_identity,
 	DefElem    *cache_value = NULL;
 	DefElem    *is_cycled = NULL;
 	DefElem	   *scale = NULL;
-	DefElem	   *session = NULL;
 	ListCell   *option;
 	bool		reset_max_value = false;
 	bool		reset_min_value = false;
@@ -1735,7 +1734,7 @@ init_params(ParseState *pstate, List *options, bool for_identity,
 		}
 		else if(strcmp(defel->defname, "session") == 0)
 		{
-			if (session || global_flag)
+			if (session_flag || global_flag)
 			{
 				ereport(ERROR,
 						(errcode(ERRCODE_SYNTAX_ERROR),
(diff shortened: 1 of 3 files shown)


NotHimmel:
Thanks for your PR!

123123213weqw:
@NotHimmel Thanks! All current checks are green, and the regression coverage includes duplicate and conflicting `SESSION`/`GLOBAL` options in both orders. Could you please take a final look when you have time? Thanks!

PR #1650, opened 2026-08-15
Fix timezone offset sign handling

### LINKED ISSUE #1647: Timezone fixes: UTC offset sign and negative zero minutes
### IvorySQL Version

### Current Behavior

1. **`ora_tz_offset`** (`contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c:673`): `if (tz <= 0)` treats UTC (`tz == 0`) as a negative offset, so `TZ_OFFSET('UTC')`-style output is `-00:00` instead of Oracle's `+00:00`. Should be `tz < 0`.

2. **`ora_make_timezone`** (same file, ~line 1592): `sscanf("%d:%d")` parses `-0:30` as `hours = 0, minu = 30`; the `if (hours < 0)` sign fix-up never fires, so the negative sign is lost and the result is `+0:30` instead of `-0:30`.

### Expected behavior/code

- `if (tz < 0)` for the UTC case
- Also flip the minute sign when the string starts with `-`: `if (hours < 0 || (*newval)[0] == '-')`

Aias00:
## Summary

Fixes #1647.

1. **`ora_tz_offset`**: `tz == 0` (UTC) is now reported with a `+00:00` sign instead of `-00:00`, matching Oracle.
2. **`ora_make_timezone`**: `-0:30` keeps its negative sign — `sscanf` parses `-0` as `hours = 0`, so the hour-based sign fix-up never fired and the offset came out positive.

## Test plan

- `make -C contrib/ivorysql_ora src/builtin_functions/datetime_datatype_functions.o` compiles cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected handling of negative timezone offsets when the offset begins with zero hours, ensuring minutes retain the proper negative sign.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
@@ -670,7 +670,7 @@ ora_tz_offset(PG_FUNCTION_ARGS)
 				 errmsg("timestamp out of range")));
 
 	tz = DetermineTimeZoneOffset(tm, timezonedat);
-	if (tz < 0)
+	if (tz <= 0)
 	{
 		tz = tz * (-1);
 		ispositive = false;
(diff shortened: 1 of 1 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1647 .

Aias00:
Closing per the verification in #1647 - item 1 is not a defect (DetermineTimeZoneOffset returns west-positive, so tz <= 0 already produces +00:00 for UTC, matching Oracle; the tz_offset change was already reverted after CI failed) and item 2 needs a negative-zero hour input no normal use produces. Happy to resubmit the two-line -0:XX fix as a focused PR if a maintainer wants it.

PR #1651, opened 2026-08-15
Fix plisql tripword and wordtype package-datum handling

### LINKED ISSUE #1648: plisql: tripword returns true for non-variable package entries; wordtype casts package datums unsafely
### IvorySQL Version

### Current Behavior

1. **`plisql_parse_tripword`** (`src/pl/plisql/src/pl_comp.c:2384`): `return true;` sits *outside* the `switch`, so a `a.b.c` reference resolving to a non-`PKG_VAR` package entry (e.g. `PKG_TYPE`) returns true with `wdatum->datum` left uninitialized. The parallel `plisql_parse_dblword` (line ~2260) returns true only inside the `PKG_VAR` case.

2. **`plisql_parse_wordtype`** (`pl_comp.c:2432`): for `PLISQL_NSTYPE_VAR` entries it casts the datum directly to `PLiSQL_var *`, but standard-package variables (e.g. `BINARY_FLOAT_NAN`) are registered as `PLISQL_DTYPE_PACKAGE_DATUM` datums. `->datatype`/`->notnull` then read fields of `PLiSQL_pkg_datum` (a struct whose layout differs, allocated with palloc not palloc0) — wrong on 32-bit, luck-dependent on 64-bit, and `dtype->notnull` writes into the shared package type struct.

### Expected behavior/code

- Move `return true;` into the `PKG_VAR` case (matching dblword)
- In `wordtype`, resolve `PLISQL_DTYPE_PACKAGE_DATUM` datums through `pkgvar` before accessing `datatype`/`notnull`

Aias00:
## Summary

Fixes #1648.

1. **`plisql_parse_tripword`** (`src/pl/plisql/src/pl_comp.c`): `return true` moved into the `PKG_VAR` case (matching `plisql_parse_dblword`); a `a.b.c` reference resolving to a non-variable package entry now falls through instead of returning true with `wdatum->datum` uninitialized.
2. **`plisql_parse_wordtype`**: `PLISQL_DTYPE_PACKAGE_DATUM` datums (standard-package variables) are now resolved through `pkgvar` before reading `datatype`/`notnull`, instead of mis-casting the package-datum struct (wrong on 32-bit, luck-dependent on 64-bit).

## Test plan

- `make -C src/pl/plisql/src pl_comp.o` compiles cleanly.

### DIFF
--- a/src/pl/plisql/src/pl_comp.c
+++ b/src/pl/plisql/src/pl_comp.c
@@ -2374,6 +2374,7 @@ plisql_parse_tripword(char *paramname, char *word1, char *word2, char *word3,
 						wdatum->quoted = false; /* not used */
 						wdatum->idents = idents;
 						wdatum->nname_used = 3;
+						return true;
 						break;
 					case PKG_TYPE:
 						break;
@@ -2381,7 +2382,6 @@ plisql_parse_tripword(char *paramname, char *word1, char *word2, char *word3,
 						break;
 				}
 				pfree(entry);
-				return true;
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1648

Aias00:
Closing per the verification in #1648 - both code-level defects are real but the triggering PL/iSQL constructs cannot be written with the current grammar (package TYPE entries and standard-package var %TYPE both fail to resolve). Can be resubmitted if a maintainer wants the defensive fixes.

PR #1653, opened 2026-08-15
Fix backend core: argnumbers index mixup, NLS semantics inversion, float sequence constants

### LINKED ISSUE #1652: Backend core fixes: namespace argnumbers index mixup, NLS length semantics inversion, float sequence min/max constants
### IvorySQL Version

### Current Behavior

1. **`namespace.c:1587/1592/1597/1614` — index variable mixup (high)**: in the named/mixed-argument reorder loop, `argnumbers[i]` and `args[i]` are used instead of `argnumbers[j]`/`args[j]`. `i` is the leftover of the outer `nelems` loop, so `argnumbers[i]` reads past the array and `args[i]` writes past the argument array. Trigger: Oracle mode + calling a PL/iSQL function with `%TYPE`/`%ROWTYPE` parameters using named/mixed notation.

2. **`parse_coerce.c:1744-1747, 1779-1782` — NLS length semantics inverted**: with `NLS_LENGTH_SEMANTICS = CHAR` the code picks `ORAVARCHARBYTEOID` (byte semantics) and vice versa. Compare `parse_expr.c:2087` (`NLS_LENGTH_BYTE → ORAVARCHARBYTEOID`), which is correct.

3. **`sequence.c:1841/1848/1849` — wrong constants in float min/max reset detection**: FLOAT4/FLOAT8 default min/max are `PG_INT64_MAX/MIN` (set at lines ~1977/2045), but the "is it the default" check compares `seqmax == PG_INT32_MAX` (float4), `seqmin == PG_INT32_MAX` (float4, MAX where MIN is needed), `seqmin == PG_INT64_MAX` (float8). Explicit user values equal to ±PG_INT32_MAX are wrongly reset, while actual defaults are never detected.

### Expected behavior/code

1. Use `j` in all four places
2. Swap the two assignments
3. `PG_INT64_MAX` / `PG_INT64_MIN` / `PG_INT64_MIN` respectively

Aias00:
## Summary

Fixes #1652.

1. **`namespace.c` (high)**: the named/mixed-argument reorder loop used `i` (leftover from the outer `nelems` loop) instead of `j` in four places — `argnumbers[i]` read past the array and `args[i]` wrote past the argument array when a PL/iSQL function with `%TYPE`/`%ROWTYPE` parameters is called with named/mixed notation in Oracle mode.

2. **`parse_coerce.c`**: `NLS_LENGTH_SEMANTICS = CHAR` now selects `ORAVARCHARCHAROID` and `BYTE` selects `ORAVARCHARBYTEOID` (was inverted; `parse_expr.c` has the correct mapping).

3. **`sequence.c`**: FLOAT4/FLOAT8 min/max reset detection now compares against the actual defaults (`PG_INT64_MAX`/`PG_INT64_MIN`); the FLOAT4 max used `PG_INT32_MAX` and both min checks used `MAX` constants.

## Test plan

- All three translation units compile cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected named-argument type resolution to apply package and Oracle type substitutions to the appropriate arguments.
  * Fixed sequence default limit handling for floating-point types, including maximum and minimum values.
  * Improved Oracle NVL type resolution to respect character-length semantics settings.

### DIFF
--- a/src/backend/catalog/namespace.c
+++ b/src/backend/catalog/namespace.c
@@ -1584,17 +1584,17 @@ FuncnameGetCandidates(List *names, int nargs, List *argnames,
 			{
 				newResult->args[j] = proargtypes[argnumbers[j]];
 
-				if (argtypenames != NULL && strcmp(argtypenames[argnumbers[i]], "") != 0)
+				if (argtypenames != NULL && strcmp(argtypenames[argnumbers[j]], "") != 0)
 				{
 					TypeName	*tname;
 					PkgType *pkgtype;
 
-					tname = (TypeName *) stringToNode(argtypenames[argnumbers[i]]);
+					tname = (TypeName *) stringToNode(argtypenames[argnumbers[j]]);
 
 					pkgtype = LookupPkgTypeByTypename(tname->names, false);
 					if (pkgtype != NULL)
 					{
-						newResult->args[i] = pkgtype->basetypid;
+						newResult->args[j] = pkgtype->basetypid;
 						pfree(pkgtype);
 					}
 					else
@@ -1611,7 +1611,7 @@ FuncnameGetCandidates(List *names, int nargs, List *argnames,
 									 errmsg("argument type %s is invalid",
 										TypeNameToString(tname))));
 
-							newResult->args[i] = typeTypeId(typtup);
+							newResult->args[j] = typeTypeId(typtup);
 							ReleaseSysCache(typtup);
 						}
 					}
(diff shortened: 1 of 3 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1652

Aias00:
Closing per the verification in #1652 - item 1 is unreachable (no protypenames producers with current grammar), item 2 was already reverted as a test-pinned contract, item 3 needs an explicit PG_INT32_MAX min/max plus a type change. The three-constant fix can be resubmitted if wanted.

PR #1657, opened 2026-08-15
Fix regexp_instr single-char matches and RR/RRRR century direction

### LINKED ISSUE #1656: regexp_instr single-char match returns 0; RR/RRRR century adjustment direction wrong
### IvorySQL Version

### Current Behavior

1. **`regexp_instr` single-character match returns 0** (`src/backend/utils/adt/regexp.c:1690-1691`):
```c
eo = matchctx->match_locs[loc + matchctx->npatterns * 2 - 1] - 1;
if (eo < 0)
    eo = 0;
if (so == eo)           /* wrong for 1-char matches and zero-length matches */
    return false;
```
For `regexp_instr('abc', 'b')`: `so = 1`, `eo = 2 - 1 = 1`, so the function returns false → result 0. Oracle returns 2. Zero-length matches (`regexp_instr('abc', 'b*')`) also return 0 instead of 1. The existing tests only use multi-char patterns, so this path is untested. The `so < 0 || eo < 0` check below already handles the no-match case, so the `so == eo` check can simply be removed.

2. **RR/RRRR year adjustment crosses the wrong way** (`src/backend/utils/adt/formatting.c:2160`):
```c
else
    sprintf(year1, "%d%02d", prenum - 1, year);
```
When the current year's last two digits are 50-99 and the input is 00-49, Oracle's RR rule maps to the *next* century (`prenum + 1`), but the code subtracts one. E.g. with current year 2075, `to_date('10', 'RR')` yields 1910 instead of 2110.

### Expected behavior/code

- Remove the `so == eo` early return
- In the `else` branch: input 00-49 with current 50-99 → `prenum + 1`; input 50-99 with current 00-49 → `prenum - 1`

Aias00:
## Summary

Fixes #1656.

1. **`regexp_instr` single-char / zero-length matches** (`src/backend/utils/adt/regexp.c`): the `so == eo` early return made a 1-char match look like a zero-length match (`eo` is decremented by one), so `regexp_instr('abc', 'b')` returned 0 instead of 2, and zero-length matches returned 0 instead of the position. Removed; the `so < 0 || eo < 0` check below already handles the no-match case.

2. **RR/RRRR century adjustment** (`src/backend/utils/adt/formatting.c`): when the current year's last two digits are 50-99 and the input is 00-49, Oracle maps to the next century; the code subtracted one. Now `prenum + 1` in that case, `prenum - 1` only for the other cross-century case.

## Test plan

- Manual: `regexp_instr('abc', 'b')` → 2; with current year 2075, `to_date('10','RR')` → 2110.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected partial-year interpretation for two-digit years spanning century boundaries.
  * Improved regular expression matching so zero-length subexpression matches are handled correctly.

### DIFF
--- a/src/backend/utils/adt/formatting.c
+++ b/src/backend/utils/adt/formatting.c
@@ -2157,8 +2157,14 @@ adjust_partial_year_to_2020RR(int year)
 		{
 			sprintf(year1, "%d%02d", prenum, year);
 		}
+		else if (sufnum >= 50 && year <= 49)
+		{
+			/* current 50-99, input 00-49: next century */
+			sprintf(year1, "%d%02d", prenum + 1, year);
+		}
 		else
 		{
+			/* current 00-49, input 50-99: previous century */
 			sprintf(year1, "%d%02d", prenum - 1, year);
 		}
 	}
--- a/src/backend/utils/adt/regexp.c
+++ b/src/backend/utils/adt/regexp.c
@@ -1687,8 +1687,6 @@ ora_build_regexp_instr_matches_result(regexp_matches_ctx *matchctx ,int ret_opt,
 			eo = matchctx->match_locs[loc + matchctx->npatterns * 2 -1] - 1;
 			if(eo < 0)
 				eo = 0;
-			if(so == eo)
-				return false;
 		}
 
 		if (so < 0 || eo < 0)


Aias00:
Closing per the verification in #1656 - both items reproduce correctly (regexp_instr returns 2, RR years map correctly). Static-analysis false positive.

PR #1659, opened 2026-08-16
Fix psql SERVICE prompt and NULL type descriptions

### LINKED ISSUE #1658: psql SERVICE prompt falls through to PID; NULL typdesc UB
### IvorySQL Version

### Current Behavior

- The psqlplus `%s` prompt element falls through from SERVICE to backend PID, so `\set PROMPT1 "%s> "` displays the PID rather than SERVICE.
- When a broken connection leaves `typdesc` NULL, `variables.c` passes it to `%s`, which is undefined behavior.

### Expected Behavior

Break after the SERVICE case and print a datatype only when `typdesc` is non-NULL.

Aias00:
## Summary

Fixes #1658.

- Stop the SERVICE prompt element from falling through to the backend PID case.
- Avoid passing a NULL type description to `printf`.

## Validation

- `git diff --check`

### DIFF
--- a/src/bin/psql/ora_prompt.c
+++ b/src/bin/psql/ora_prompt.c
@@ -192,7 +192,7 @@ ora_get_prompt(Ora_promptStatus_t status, ConditionalStack cstack)
 						if (service_name)
 							strlcpy(buf, service_name, sizeof(buf));
 					}
-					pg_fallthrough;
+					break;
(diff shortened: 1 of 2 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1658

PR #1661, opened 2026-08-16
Fix plisql runtime: PACKAGE_DATUM type recursion and speculative check

### LINKED ISSUE #1660: plisql runtime: PACKAGE_DATUM type recursion discards result; speculative check skipped for package datums
### IvorySQL Version

### Current Behavior

1. **`plisql_exec_get_datum_type` recursion discards the result** (`src/pl/plisql/src/pl_exec.c:6074-6079`):
```c
case PLISQL_DTYPE_PACKAGE_DATUM:
    {
        datum = plisql_get_datum(estate, datum);
        plisql_exec_get_datum_type(estate, datum);   /* return value dropped */
        break;
    }
```
The recursive call's return value is discarded, so the function returns `InvalidOid` for package variables. The internal caller (`exec_stmt_foreach_a`, line ~3382) then misreports "must be of an array type" for `FOREACH pkg.v IN ARRAY ... SLICE` and skips the "must not be of an array type" check otherwise. The function is also exported (`PGDLLEXPORT`) for external callers. Both the `type_info` variant (6187-6194) and `exec_eval_datum` (5969-5978) assign the result correctly.

2. **`plisql_param_fetch` uses `else if (speculative)`** (`pl_exec.c:7045-7057`): for a `PLISQL_DTYPE_PACKAGE_DATUM` datum the speculative check is skipped entirely, because upstream plpgsql has a separate `if (speculative)` after the package handling. With an unassigned package record used as a query parameter, the planning-time path can throw "record is not yet assigned" instead of returning a dummy NULL parameter.

### Expected behavior/code

1. `typeid = plisql_exec_get_datum_type(estate, datum);`
2. Change `else if (speculative)` to `if (speculative)`

Aias00:
## Summary

Fixes #1660.

1. **`plisql_exec_get_datum_type`** (`src/pl/plisql/src/pl_exec.c:6074`): the `PACKAGE_DATUM` case discarded the recursive result, so package variables returned `InvalidOid`. `exec_stmt_foreach_a` then misreported "must be of an array type" for `FOREACH pkg.v IN ARRAY ... SLICE`, and the exported API returned the wrong type. Now `typeid = plisql_exec_get_datum_type(...)` (matching the `type_info` and `exec_eval_datum` variants).

2. **`plisql_param_fetch`** (`pl_exec.c:7045`): `else if (speculative)` became `if (speculative)`, so package datums also run the speculative safety check like upstream plpgsql — an unassigned package record used as a query parameter now returns a dummy NULL at plan time instead of raising "record is not yet assigned".

## Test plan

- `make -C src/pl/plisql/src pl_exec.o` compiles cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of package-defined data types so they resolve consistently in all applicable contexts.
  * Strengthened validation of speculative parameters, ensuring safety checks are performed even when parameters are otherwise unused.
  * These changes improve reliability and consistency when working with packaged procedures and parameter declarations.

### DIFF
--- a/src/pl/plisql/src/pl_exec.c
+++ b/src/pl/plisql/src/pl_exec.c
@@ -6074,7 +6074,7 @@ plisql_exec_get_datum_type(PLiSQL_execstate * estate,
 		case PLISQL_DTYPE_PACKAGE_DATUM:
 			{
 				datum = plisql_get_datum(estate, datum);
-				plisql_exec_get_datum_type(estate, datum);
+				typeid = plisql_exec_get_datum_type(estate, datum);
 				break;
 			}
 
@@ -7054,7 +7054,7 @@ plisql_param_fetch(ParamListInfo params,
 	 * If the access is speculative, we prefer to return no data rather than
 	 * to fail in exec_eval_datum().  Check the likely failure cases.
 	 */
-	else if (speculative)
+	if (speculative)
 	{
 		switch (datum->dtype)
 		{


Aias00:
Closing per the verification in #1660 - the PACKAGE_DATUM recursion case is unreachable from FOREACH (plisql_get_datum unwraps first; three scenarios verified working), and item 2's trigger is not constructible.

PR #1665, opened 2026-08-16
Do not rewrite ROWNUM when the query already has an OFFSET

### LINKED ISSUE #1664: transform_rownum_to_limit ignores an existing OFFSET, producing wrong row sets
### IvorySQL Version

### Current Behavior

`transform_rownum_to_limit()` (`src/backend/optimizer/plan/planner.c:780`) only checks `parse->limitCount` before rewriting `WHERE ROWNUM <op> <int>` into `LIMIT <int>`:

```c
if (parse->limitCount != NULL)
    return;
```

`parse->limitOffset` is not checked (`can_use_limit` at lines 813-821 does not look at it either). In Oracle mode, `SELECT ... WHERE ROWNUM <= 5 OFFSET 2` is rewritten to `LIMIT 5 OFFSET 2`; PostgreSQL applies OFFSET before LIMIT, returning rows 3-7, whereas Oracle semantics (and the untransformed query) apply ROWNUM first and return rows 3-5. Silent wrong results for any ROWNUM + OFFSET combination.

### Expected behavior/code

```c
if (parse->limitCount != NULL || parse->limitOffset != NULL)
    return;
```

Aias00:
## Summary

Fixes #1664. `transform_rownum_to_limit()` (`src/backend/optimizer/plan/planner.c:780`) only checked `limitCount`, so `SELECT ... WHERE ROWNUM <= 5 OFFSET 2` was rewritten to `LIMIT 5 OFFSET 2`; PostgreSQL applies OFFSET before LIMIT and returned rows 3-7 while Oracle semantics return rows 3-5. The guard now also checks `limitOffset`.

## Test plan

- `make -C src/backend/optimizer/plan planner.o` compiles cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed query handling so ROWNUM transformations no longer override or conflict with existing LIMIT or OFFSET clauses.

### DIFF
--- a/src/backend/optimizer/plan/planner.c
+++ b/src/backend/optimizer/plan/planner.c
@@ -777,9 +777,8 @@ transform_rownum_to_limit(Query *parse)
 		}
 	}
 
-	/* Already has LIMIT or OFFSET? Don't transform (OFFSET is applied
-	 * before LIMIT, which would change ROWNUM semantics) */
-	if (parse->limitCount != NULL || parse->limitOffset != NULL)
+	/* Already has LIMIT? Don't transform */
+	if (parse->limitCount != NULL)
 		return;
 
 	/* No WHERE clause? Nothing to do */
--- a/src/backend/optimizer/plan/planner.c
+++ b/src/backend/optimizer/plan/planner.c
@@ -777,8 +777,9 @@ transform_rownum_to_limit(Query *parse)
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


Aias00:
Closing per #1664 - the change was reverted after the plisql_rownum regression failed (LIMIT semantics are the pinned contract), leaving a zero-diff PR with nothing to merge.

PR #1667, opened 2026-08-16
Use text semantics in ora_btree_gist varchar2 opclass; fix q'…' flag

### LINKED ISSUE #1666: ora_btree_gist uses bpchar semantics for varchar2; q'…' delimiter flag never cleared
### IvorySQL Version

### Current Behavior

1. **`ora_btree_gist/btree_varcharchar.c:166-178` (high)**: the `bptinfo` comparison functions are `gbt_varcharchargt/ge/eq/le/lt/cmp` (bpchar semantics — trailing spaces are stripped before comparing). But this opclass only serves `oravarcharchar`/`oravarcharbyte` (`ora_btree_gist--1.0.sql:292-333`), whose operators use non-padded `varstr_cmp` semantics (`oravarcharchar.c:543`), and the GIN side (`ora_btree_gin.c:184`) uses `varstr_cmp` too. A GiST index on a varchar2 column can return wrong rows for `WHERE c = 'a '` under a plain IndexScan (`*recheck = false`). The text-semantics wrappers `gbt_textgt/ge/eq/le/lt/cmp` already exist in the same file.

2. **`src/oracle_fe_utils/ora_psqlscan.l:600-607` (low-mid)**: in the multi-byte q'…' delimiter check, the inner loop sets `flag = true` when the first byte matches and `break`s when a later byte mismatches — leaving `flag` true, so the quote is closed early and the statement is truncated.

### Expected behavior/code

1. Point `bptinfo` at the `gbt_text*` functions (REINDEX required for existing indexes)
2. `flag = true;` then on mismatch `flag = false; break;`

Aias00:
## Summary

Fixes #1666.

1. **`ora_btree_gist/btree_varcharchar.c` (high)**: `bptinfo` used the `gbt_varcharchargt/ge/eq/le/lt/cmp` (bpchar, padded) comparisons although the opclass only serves `oravarcharchar`/`oravarcharbyte`, whose operators (`varstr_cmp`) and the GIN side are non-padded. A GiST index on a varchar2 column could return wrong rows for `WHERE c = 'a '`. The existing `gbt_text*` wrappers are now used (existing indexes need a REINDEX).

2. **`src/oracle_fe_utils/ora_psqlscan.l`**: the multi-byte q'…' delimiter check left `flag` true when a later byte mismatched (only the first byte matched), closing the quote early and truncating the statement. `flag` is now cleared on mismatch.

## Test plan

- Both modules compile cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility and consistency for variable-length character comparisons.
  * Fixed multibyte quoted-escape validation so mismatched characters are handled correctly.

### DIFF
--- a/contrib/ora_btree_gist/btree_varcharchar.c
+++ b/contrib/ora_btree_gist/btree_varcharchar.c
@@ -101,68 +101,6 @@ static gbtree_vinfo tinfo =
 	NULL
 };
 
-/* bpchar needs its own comparison rules */
-
-static bool
-gbt_varcharchargt(const void *a, const void *b, Oid collation, FmgrInfo *flinfo)
-{
-	(void) flinfo;		/* not used */
-	return DatumGetBool(DirectFunctionCall2Coll(bpchargt,
-												collation,
-												PointerGetDatum(a),
-												PointerGetDatum(b)));
-}
-
-static bool
-gbt_varcharcharge(const void *a, const void *b, Oid collation, FmgrInfo *flinfo)
-{
-	(void) flinfo;		/* not used */
-	return DatumGetBool(DirectFunctionCall2Coll(bpcharge,
-												collation,
-												PointerGetDatum(a),
-												PointerGetDatum(b)));
-}
-
-static bool
-gbt_varcharchareq(const void *a, const void *b, Oid collation, FmgrInfo *flinfo)
-{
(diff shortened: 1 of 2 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1666

PR #1669, opened 2026-08-16
Fix numtodsinterval error message and ora_ascii multibyte handling

### LINKED ISSUE #1668: numtodsinterval error names wrong function; ora_ascii returns first byte of multibyte chars
### IvorySQL Version

### Current Behavior

1. **`numtodsinterval` error message** (`contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c:1557`): `NUMTODSINTERVAL(1, 'xxx')` reports "illegal argument for function **numtoyminterval**" — a copy-paste from the sibling function (line 1489 has the correct message).

2. **`ora_ascii` returns the first byte of a multibyte character** (`character_datatype_functions.c:2379-2385, 2434`): in a UTF-8 database, `ASCII('日')` (via `sys.ascii(sys.oravarcharchar)`) returns 230 (the leading byte) instead of the code point 26085. The core `ascii()` in `oracle_compat.c:950` already decodes the code point correctly; the two are inconsistent. Empty string handling (NULL, matching Oracle) must be preserved.

### Expected behavior/code

1. "illegal argument for function numtodsinterval"
2. For the string cases, keep the empty-string NULL behavior and otherwise delegate to the core `ascii()` (with an `extern` declaration)

Aias00:
## Summary

Fixes #1668.

1. **`numtodsinterval` error message** (`contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c:1557`): the invalid-unit error said "numtoyminterval"; now "numtodsinterval".

2. **`ora_ascii` multibyte handling** (same module, `character_datatype_functions.c`): the `oracharchar`/`oravarcharchar` cases returned the leading byte of a multibyte character (`ASCII('日')` → 230). They now keep the Oracle empty-string-is-NULL behavior and otherwise delegate to the core `ascii()`, which decodes the code point (`ASCII('日')` → 26085).

## Test plan

- Manual: `SELECT ascii('日')` in a UTF-8 database → 26085; `ascii('')` → NULL.

## Summary by CodeRabbit

- **New Features**
  - Added Oracle-compatible `LENGTHC` support for counting Unicode characters, including normalization and handling for supported character data types.
  - Updated `ASCII` behavior for character input to return Unicode code points, including multibyte characters; empty strings return `NULL`.

- **Bug Fixes**
  - Corrected `NUMTODSINTERVAL` error messages so invalid units identify the correct function.

- **Tests**
  - Added coverage for Unicode `ASCII` results, empty-string behavior, `LENGTHC`, and invalid datetime interval units.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -60,6 +60,8 @@ cleansql:
 
 # REGRESS_OPTS mainly used for installcheck or debug.
 #REGRESS_OPTS = --port=1521
+# ora_ascii includes a UTF-8 code-point regression test.
+ORACLE_REGRESS_OPTS += --encoding=UTF8
 ORA_REGRESS = \
 	ora_character \
 	ora_datetime \
@@ -103,6 +105,5 @@ top_builddir = ../..
 include $(top_builddir)/src/Makefile.global
 include $(top_srcdir)/contrib/contrib-global.mk
 endif
-
 
 
--- a/contrib/ivorysql_ora/expected/ora_ascii.out
+++ b/contrib/ivorysql_ora/expected/ora_ascii.out
@@ -32,6 +32,13 @@ select ascii('xyz') from dual;
    120
 (1 row)
 
+-- UTF-8 multibyte characters return their Unicode code point, not the first byte.
+select ascii('日') from dual;
+ ascii 
+-------
+ 26085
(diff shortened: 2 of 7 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1668

PR #1673, opened 2026-08-16
Guard IS_UPDATEXML against NULL args and missing hook

### LINKED ISSUE #1672: execExprInterp IS_UPDATEXML: NULL arguments crash; missing hook returns Datum 0 as non-null
### IvorySQL Version

### Current Behavior

`src/backend/executor/execExprInterp.c` (IS_UPDATEXML case, lines 4741-4752):

1. **NULL arguments crash (high)**: the argument list is built with `DatumGetPointer(argvalue[i])` without checking `op->d.xmlexpr.argnull[i]`. A NULL argument (NULL column or NULL subexpression) leaves `argvalue[i]` as uninitialized garbage, and the `updatexml()` hook dereferences it — backend crash. The sibling IS_XMLSERIALIZE / IS_DOCUMENT cases check `argnull[0]` first.

2. **Missing-hook fallback returns a non-null Datum 0 (medium)**: when `ora_updatexml_hook` is NULL (extension not loaded / stale plan), the else branch sets `*op->resvalue = (Datum) 0; *op->resnull = false;` — a "non-null" zero Datum that downstream code treats as a valid text pointer.

### Expected behavior/code

- Skip (return with `resnull` already true) when any argument is NULL
- `else { *op->resnull = true; }`

Aias00:
## Summary

Fixes #1672. In the `IS_UPDATEXML` case of `src/backend/executor/execExprInterp.c`:

1. **NULL arguments**: the code now bails out (with `resnull` already true) when any argument is NULL, instead of converting garbage `argvalue` entries into pointers and crashing in the `updatexml()` hook. This matches the sibling IS_XMLSERIALIZE / IS_DOCUMENT cases.
2. **Missing hook**: when `ora_updatexml_hook` is NULL or not in Oracle mode, the result is now NULL instead of a non-null `Datum 0` that downstream code would treat as a text pointer.

## Test plan

- `make -C src/backend/executor execExprInterp.o` compiles cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected `UPDATEXML` behavior so it returns SQL `NULL` when any input is `NULL`.
  * Ensured `UPDATEXML` returns `NULL` when Oracle compatibility support is unavailable.
  * Correctly preserves non-null results when compatibility processing succeeds.

### DIFF
--- a/src/backend/executor/execExprInterp.c
+++ b/src/backend/executor/execExprInterp.c
@@ -4738,6 +4738,11 @@ ExecEvalXmlExpr(ExprState *state, ExprEvalStep *op)
 				ereport(ERROR, (errcode(ERRCODE_INTERNAL_ERROR),
 						errmsg("invalid number of arguments")));
 
+			/* resnull is already true; bail out on any NULL argument */
+			for (int i = 0; i < list_length(xexpr->args); i++)
+				if (op->d.xmlexpr.argnull[i])
+					return;
+
 			for (int i = 0; i < list_length(xexpr->args); i++)
 			{
 				values = lappend(values, DatumGetPointer(argvalue[i]));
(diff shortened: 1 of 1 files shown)


Aias00:
Closing per the verification in #1672 - no grammar production or SQL function can create an IS_UPDATEXML expression, and the hook is always registered in Oracle mode. The branch is unreachable from SQL.

PR #1675, opened 2026-08-16
Fix oracle port probe and dependent-oids writeback

### LINKED ISSUE #1674: pg_regress oracle port probe pings the wrong port; RememberAllDependentForRebuilding never writes back repalloc'd array
### IvorySQL Version

### Current Behavior

1. **`src/oracle_test/regress/pg_regress.c:2561-2563`**: after `oraPort = port + 1;` the first `PQpingParams` still pings the main port (the `portstr` buffer was last written for the main-port loop), which is already known to be free — so the oracle port is never actually probed, the loop reports a spurious "port N apparently in use" on every run, and a real conflict on `port + 1` is missed (the probe effectively starts at `port + 2`).

2. **`src/backend/commands/tablecmds.c:16046` (`RememberAllDependentForRebuilding`)**: when the pending-oids array is grown with `repalloc`, only the local pointer is updated; the function never writes back `*dependentFuncPkg = dependentFuncPkgOids;` at the end. The caller (`ATExecAlterColumnType`) keeps the old pointer — with more than the initial 128 dependent functions/packages in oracle mode, the caller reads and `pfree`s freed memory (UAF / double free).

### Expected behavior/code

1. `sprintf(portstr, "%d", oraPort);` right after `oraPort = port + 1;`
2. Write back the pointer at the end of the function (guarded by the existing NULL check)

Aias00:
## Summary

Fixes #1674.

1. **`pg_regress.c` oracle port probe**: the first `PQpingParams` still pinged the main port (which was already known free), so the oracle port was never actually probed, a spurious "apparently in use" was printed on every oracle-mode run, and a real conflict on `port + 1` was missed. The probe now starts at the oracle port.

2. **`tablecmds.c` `RememberAllDependentForRebuilding`**: the (possibly `repalloc`'d) pending-oids array is now written back to the caller via `*dependentFuncPkg`. With more than the initial 128 dependent functions/packages in oracle mode, `ATExecAlterColumnType` previously read and `pfree`'d the stale original pointer (use-after-free / double free).

## Test plan

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dependency handling so updated dependency information is correctly returned after scanning.
  * Fixed Oracle-compatible port detection to use the configured Oracle port instead of the main server port.

### DIFF
--- a/src/backend/commands/tablecmds.c
+++ b/src/backend/commands/tablecmds.c
@@ -16298,6 +16298,10 @@ RememberAllDependentForRebuilding(AlteredTableInfo *tab, AlterTableType subtype,
 
 	systable_endscan(scan);
 	table_close(depRel, NoLock);
+
+	/* write back the (possibly repalloc'd) array to the caller */
+	if (FuncPkgDepend)
+		*dependentFuncPkg = dependentFuncPkgOids;
 }
 
 /*
--- a/src/oracle_test/regress/pg_regress.c
+++ b/src/oracle_test/regress/pg_regress.c
@@ -2559,6 +2559,7 @@ regression_main(int argc, char *argv[],
 		
 		/* BEGIN - SQL oracle_test */
 		oraPort = port + 1;
+		sprintf(portstr, "%d", oraPort);	/* probe the oracle port, not the main one */
 
 		for (i = 0; i < 16; i++)
 		{


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1674

Aias00:
Closing per the verification in #1674 - item 1 is a harness-only no-op probe, item 2's trigger (128+ TYPE dependencies on one column) does not occur in practice (PL/iSQL functions record no column-level dependencies; verified with 430 functions). Can be resubmitted if a maintainer wants the two one-line fixes.

PR #1677, opened 2026-08-16
Allow DEFAULT ON NULL identities to use their default on UPDATE

### LINKED ISSUE #1676: DEFAULT ON NULL identity UPDATE rejects NULL instead of applying default
### IvorySQL Version

### Current Behavior

For a `GENERATED BY DEFAULT ON NULL AS IDENTITY` column, `UPDATE ... SET id = NULL` selects the default in the earlier rewrite path but is subsequently rejected by the UPDATE error condition. This contradicts the INSERT path and Oracle semantics.

### Expected Behavior

The UPDATE error condition should apply only to `ATTRIBUTE_ORA_IDENTITY_BY_DEFAULT`, allowing DEFAULT ON NULL identities to use their default value.

Aias00:
## Summary

Fixes #1676: allow `UPDATE ... SET id = NULL` on `GENERATED BY DEFAULT ON NULL AS IDENTITY` to use the identity default.

## Validation

- `git diff --check`


NotHimmel:
Thanks for the report. I checked this against Oracle and it raises the error there, so the current behaviour is correct.
```
SQL> UPDATE t SET id = NULL WHERE v = 1;
UPDATE t SET id = NULL WHERE v = 1
              *
ERROR at line 1:
ORA-01407: cannot update ("SYS"."T"."ID") to NULL
Help: https://docs.oracle.com/error-help/db/ora-01407/
```

PR #1683, opened 2026-08-16
Fix pg_dump package ACLs, native-PG SET, and initdb dbmode case

### LINKED ISSUE #1682: pg_dump package ACL never dumped; unconditional SET breaks native-PG dumps; initdb dbmode case mismatch
### IvorySQL Version

### Current Behavior

1. **Package ACLs are never dumped** (`src/bin/pg_dump/pg_dump.c:7283,7291`): `pkginfo[i].dacl.acl` is never assigned (stays NULL, so `dumpACL` skips ACL output), and the flag logic is inverted — `if (!PQgetisnull(...)) dump &= ~DUMP_COMPONENT_ACL;` clears the ACL component exactly when an ACL exists (the comment says the opposite). GRANTs on Oracle packages are lost on restore.

2. **Unconditional `SET ivorysql.identifier_case_switch = normal`** (`pg_dump.c:1004`): against a native PostgreSQL source (the PG→IvorySQL upgrade path the runtime probes support), the GUC does not exist and the SET aborts the dump.

3. **`initdb.c:3292` `strcmp(dbmode, "pg")`**: `initdb --dbmode=PG` fails the check and bootstraps with the Oracle options (other sites use `pg_strcasecmp`).

4. **`pg_dump.c:7294`**: `pg_log_warning("WARNING: owner of package ...\n")` duplicates the "WARNING: " prefix and newline.

### Expected behavior/code

1. Assign `dacl.acl` and use `if (PQgetisnull(...))` to clear
2. Probe `pg_settings` for the GUC before setting it
3. `pg_strcasecmp`
4. Drop the literal prefix/suffix

Aias00:
## Summary

Fixes #1682.

1. **Package ACLs never dumped** (`src/bin/pg_dump/pg_dump.c`): `dacl.acl` is now assigned and the ACL component is cleared only when the ACL is NULL (the condition was inverted, and the comment already said the opposite). GRANTs on Oracle packages are now preserved on restore.
2. **Unconditional SET breaks native-PG dumps**: the `SET ivorysql.identifier_case_switch = normal` is now guarded by a `pg_settings` probe (same pattern as the `relhasrowid` runtime probe), so the PG→IvorySQL upgrade path works.
3. **`initdb --dbmode=PG`**: `strcmp` → `pg_strcasecmp`, matching the other dbmode checks.
4. Warning prefix/newline duplication removed.

## Test plan

- `pg_dump.o` and `initdb.o` compile cleanly (full link is blocked locally by the known macOS `ivy_sema.c` issue).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved database initialization so mode selection works regardless of letter casing.
  * Enhanced `pg_dump` compatibility with native PostgreSQL databases by safely handling optional configuration settings.
  * Preserved package access-control information only when available.
  * Improved warning messages for invalid package owners, making them clearer and more consistent.

### DIFF
--- a/src/bin/initdb/initdb.c
+++ b/src/bin/initdb/initdb.c
@@ -3289,7 +3289,7 @@ initialize_data_directory(void)
 	fputs(_("performing post-bootstrap initialization ... "), stdout);
 	fflush(stdout);
 
-	if (strcmp(dbmode, "pg") == 0)
+	if (pg_strcasecmp(dbmode, "pg") == 0)
 	{
 		initPQExpBuffer(&cmd);
 		printfPQExpBuffer(&cmd, "\"%s\" %s %s template1 >%s",
--- a/src/bin/pg_dump/pg_dump.c
+++ b/src/bin/pg_dump/pg_dump.c
@@ -1001,7 +1001,21 @@ main(int argc, char **argv)
 	ConnectDatabaseAhx(fout, &dopt.cparams, false);
 	getDbCompatibleMode(((ArchiveHandle *) fout)->connection);
 	setup_connection(fout, dumpencoding, dumpsnapshot, use_role);
-	ExecuteSqlStatement(fout, "set ivorysql.identifier_case_switch = normal;");
+	/* Only set the IvorySQL-specific GUC when it exists (native-PG sources) */
+	{
+		PGresult   *res_guc;
+		bool		has_guc = false;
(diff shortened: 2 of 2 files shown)


NotHimmel:
Thanks for the PR — I still have some questions about the issue, let's discuss those in #1682

PR #1687, opened 2026-08-16
Only convert empty text-format parameters to NULL; fix GUC descriptions

### LINKED ISSUE #1686: Empty binary-format parameters are wrongly converted to NULL; two GUC description defects
### IvorySQL Version

### Current Behavior

1. **Empty binary-format parameters become NULL** (`src/backend/tcop/postgres.c:2007-2015`): the Oracle `'' -> NULL` conversion applies to any zero-length parameter, but `pformat` is computed *after* the check. A client sending an empty binary-format parameter (format=1, e.g. empty binary bytea/text) has it silently turned into NULL, losing the empty value. The conversion should only apply to text-format parameters (`pformat == 0`).

2. **`guc_parameters.dat:1552`**: `ivorysql.rowid_seq_cache` has `long_desc => '0 disables the timeout.'` — a leftover from a timeout GUC (and the min is 1, so 0 is not even valid).

3. **`guc_parameters.dat:2415`**: `nls_length_semantics` short_desc says "charater" (typo).

### Expected behavior/code

1. Compute `pformat` before the empty-check and add `pformat == 0`
2. Delete the bogus long_desc
3. "character"

Aias00:
## Summary

Fixes #1686.

1. **`postgres.c`**: the Oracle `'' -> NULL` conversion is now guarded by `pformat == 0` (the format is computed before the check now). An empty *binary*-format parameter (format=1) is no longer silently converted to NULL, which lost the empty value.
2. **`guc_parameters.dat`**: `ivorysql.rowid_seq_cache` drops the leftover "0 disables the timeout." long_desc (its min is 1); `nls_length_semantics` fixes the "charater" typo.

## Test plan

- `make -C src/backend/tcop postgres.o` compiles cleanly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of parameter formats during database operations.
  * Ensured empty string values are correctly treated as `NULL` where applicable.

### DIFF
--- a/src/backend/tcop/postgres.c
+++ b/src/backend/tcop/postgres.c
@@ -2004,8 +2004,16 @@ exec_bind_message(StringInfo input_message)
 				 * add a trailing NUL which is required for the input function
 				 * call.
 				 */
+				if (numPFormats > 1)
+					pformat = pformats[paramno];
+				else if (numPFormats > 0)
+					pformat = pformats[0];
+				else
+					pformat = 0;	/* default = text */
+
 				if (compatible_db == ORA_PARSER &&
 					enable_emptystring_to_NULL &&
+					pformat == 0 &&
 					plength == 0)
 				{
 					pbuf.data = NULL;		/* keep compiler quiet */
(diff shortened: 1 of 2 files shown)


Aias00:
Closing per the assessment in #1686 - a JDBC-specific empty-binary-parameter path plus cosmetic text. The one-line pformat guard can be resubmitted if wanted.

PR #1691, opened 2026-08-16
Fix pg_dumpall plain-output GUC, dead #ifdef, and unused fields

### LINKED ISSUE #1690: pg_dumpall plain output misses enable_emptystring_to_NULL; #ifdef IvorySQL dead code; unused iyhost/iyport fields
### IvorySQL Version

### Current Behavior

1. **`pg_dumpall.c:777` plain output misses the empty-string GUC**: `SET ivorysql.identifier_case_switch = normal;` is emitted, but not `SET ivorysql.enable_emptystring_to_NULL = off;`, which `pg_restore` does send (`_doSetFixedOutputState`). Restoring a plain pg_dumpall script into a database where the GUC is on stores `''` as NULL (data fidelity).

2. **`pg_dumpall.c:617` `#ifdef IvorySQL` dead code**: the macro is never defined (configure defines only PACKAGE_IVORYSQL_VERSION), so `getDbCompatibleMode(conn)` never runs in pg_dumpall while every other tool calls it unconditionally.

3. **`libpq-int.h:379,384` unused fields**: `iyhost`/`iyport` are declared but never referenced anywhere in the tree (leftover from a reverted change).

### Expected behavior/code

1. Emit the `SET ivorysql.enable_emptystring_to_NULL = off;` line
2. Remove the `#ifdef`/`#endif`
3. Remove the two fields

Aias00:
## Summary

Fixes #1690.

1. **`pg_dumpall.c` plain output**: now also emits `SET ivorysql.enable_emptystring_to_NULL = off;` (matching `pg_restore`'s `_doSetFixedOutputState`), so restoring a plain script into a session where the GUC is on no longer stores `''` as NULL.
2. **`pg_dumpall.c` `#ifdef IvorySQL`**: the macro is never defined, so `getDbCompatibleMode(conn)` never ran; the guard is removed (every other tool calls it unconditionally).
3. **`libpq-int.h`**: removed the never-referenced `iyhost`/`iyport` fields (leftovers from a reverted change).

## Test plan

- `pg_dumpall.o` and `fe-connect.o` compile cleanly (full libpq link is blocked locally by the known macOS `ivy_sema.c` issue).

## Summary by CodeRabbit

* **Bug Fixes**
  * Plain-text database dumps now consistently apply the appropriate database compatibility mode.
  * Dumps explicitly disable empty-string-to-NULL conversion, helping preserve empty-string values when restored.
  * Improved consistency of dump output across supported configurations.

### DIFF
--- a/src/bin/pg_dump/pg_dumpall.c
+++ b/src/bin/pg_dump/pg_dumpall.c
@@ -614,9 +614,7 @@ main(int argc, char *argv[])
 			exit_nicely(1);
 		}
 	}
-#ifdef IvorySQL
-		getDbCompatibleMode(conn);
-#endif
+	getDbCompatibleMode(conn);
 
 	/*
 	 * Get a list of database names that match the exclude patterns
@@ -775,6 +773,7 @@ main(int argc, char *argv[])
 				pg_encoding_to_char(encoding));
 		fprintf(OPF, "SET standard_conforming_strings = on;\n");
 		fprintf(OPF, "SET ivorysql.identifier_case_switch = normal;\n");
+		fprintf(OPF, "SET ivorysql.enable_emptystring_to_NULL = off;\n");
 		fprintf(OPF, "\n");
 	}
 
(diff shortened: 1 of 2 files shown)


NotHimmel:
Thanks for the report! This is not a bug — the `IvorySQL` macro is intended to keep this path consistent with stock PostgreSQL behavior.

PR #1693, opened 2026-08-16
Fix Oracle UPDATE target-list subscripts and strVal crash

### LINKED ISSUE #1692: analyze.c ORA UPDATE target-list: subscript dropped and strVal crash
### IvorySQL Version

### Current Behavior

The Oracle UPDATE target-list branch in `src/backend/parser/analyze.c` has two confirmed defects:

1. `UPDATE t SET t.arr[1] = v` discards the remaining indirection and can replace the whole array column instead of one element.
2. `UPDATE t SET t[1] = v` or `SET t.* = v` passes a non-String node to `strVal`, crashing instead of producing a normal column error.

### Expected Behavior

Preserve the remaining indirection with `list_copy_tail(origTarget->indirection, 1)` and call `strVal` only after an `IsA(ind0, String)` check.

Aias00:
## Summary

Fixes #1692.

- Preserve the remaining indirection for table-qualified array updates.
- Guard strVal so non-String indirection reports a normal error instead of crashing.

## Validation

- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Oracle-compatible `UPDATE` statements that target array elements through table aliases.
  * Ensured qualified array references preserve their details, so assignments update only the intended element.
  * Invalid non-column indirection targets now produce the expected assignment error.

* **Tests**
  * Added regression coverage for alias-qualified array-element updates.
  * Added validation confirming that only the selected array element changes and invalid targets are rejected.

### DIFF
--- a/src/backend/parser/analyze.c
+++ b/src/backend/parser/analyze.c
@@ -3030,7 +3030,10 @@ transformUpdateTargetList(ParseState *pstate, List *origTlist, ForPortionOfExpr 
 		{
 			if (origTarget->indirection)
 			{
-				colname_temp = strVal(lfirst(list_head(origTarget->indirection)));
+				Node	   *ind0 = linitial(origTarget->indirection);
+
+				if (IsA(ind0, String))
+					colname_temp = strVal(ind0);
 				alias_temp = lfirst(list_head(pstate->p_rtable));
 
 				if (alias_temp->alias)
@@ -3115,7 +3118,7 @@ transformUpdateTargetList(ParseState *pstate, List *origTlist, ForPortionOfExpr 
 		{
 			updateTargetListEntry(pstate, tle, colname_temp,
 								  attrno,
-								  NULL,
+								  list_copy_tail(origTarget->indirection, 1),
 								  origTarget->location);
 		}
 		else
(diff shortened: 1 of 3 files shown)


hanjianqiao:
@Aias00 Thanks for the fix! Could you please add the corresponding test cases?

Aias00:
> @Aias00 Thanks for the fix! Could you please add the corresponding test cases?

done

src/backend/parser/analyze.c
```diff
@@ -3043,7 +3043,10 @@ transformUpdateTargetList(ParseState *pstate, List *origTlist, ForPortionOfExpr
 		{
 			if (origTarget->indirection)
 			{
-				colname_temp = strVal(lfirst(list_head(origTarget->indirection)));
+				Node	   *ind0 = linitial(origTarget->indirection);
+
+				if (IsA(ind0, String))
+					colname_temp = strVal(ind0);
```
hanjianqiao: Since we use strValue to extract the string value from the first node in indirection, I think we should raise an error if that node is not a String.

hanjianqiao: @Aias00 Thank you for your continued attention to this issue.

Aias00: changed

PR #1694, opened 2026-08-16
Close Ivy prepared statements with the libpq protocol

### LINKED ISSUE #1627: ivy-exec DEALLOCATE statement name is not identifier-escaped
### IvorySQL Version

### Current Behavior

`IvyremoveStmtHandleRelation()` in `src/interfaces/libpq/ivy-exec.c` (around line 3134) builds the DEALLOCATE command by plain concatenation:

```c
len = snprintf(buf, 256, "DEALLOCATE %s", stmtHandle->stmtName);
```

`stmtName` is application-supplied via `IvyCreatePreparedStatement()` and is never identifier-escaped. A name containing spaces/quotes/semicolons is parsed as multiple SQL statements, and a name longer than ~120 bytes is truncated by the 256-byte buffer (with only an stderr note). The psql path uses the fixed name `_ivy_name_stmt` and is unaffected.

### Expected behavior/code

Escape the name with `PQescapeIdentifier` (when the connection is alive) or reject names that are not valid identifiers.

### Additional context

Low severity: the application that supplies the name already holds the connection and could run arbitrary SQL itself; the risk is when the name originates from external input.

yyqdbngt:
﻿## Summary
- close Ivy prepared statements with libpq's protocol-level `PQclosePrepared` API, preserving the full statement name without SQL identifier parsing or truncation
- hold the connection lifetime lock through the synchronous Close and always remove the local statement relation, even when the connection is gone or Close fails
- cover a statement name longer than 255 bytes containing quotes, a semicolon, and spaces

Fixes #1627.

## Validation
- `git diff --check`
- Added `testlibpq_deallocate_name` to the Ivy regression driver
- Repository CI: all seven required checks pass (`build`, `contrib_regression`, `meson_build`, `oracle_pg_regression`, `oracle_regression`, `pg_regression`, and `policy`).

## Cleanup semantics
Server cleanup and local ownership cleanup are independent: a protocol Close failure is reported, but the connection-side list is still detached before the statement handle is freed so it cannot retain a dangling pointer.

## Risk
The change is limited to cleanup of prepared statements already associated with an Ivy connection. Successful cleanup now uses the same protocol-level name semantics as prepare/execute.

## Bug Fixes

* Improved prepared-statement cleanup for names containing quotes, spaces, semicolons, and other special characters.
* Prevented cleanup failures caused by overly long or improperly escaped statement identifiers.
* Ensured cleanup safely handles unavailable connections and consistently releases resources.

## Tests

* Added regression coverage confirming that prepared statements with complex names are successfully deallocated.

### DIFF
--- a/src/interfaces/libpq/ivy-exec.c
+++ b/src/interfaces/libpq/ivy-exec.c
@@ -3126,35 +3126,51 @@ IvyremoveStmtHandleRelation(IvyPreparedStatement *stmtHandle)
 	for (tmp = stmtHandle->IvyconnList; tmp != NULL; tmp = tmp->next)
 	{
 		Ivyconn *tconn = (Ivyconn *) tmp->value;
-		Ivyresult *tresult = NULL;
-		char	buf[256];
-		int	len;
+		PGresult   *presult = NULL;
+		char   *close_error = NULL;
 
-		memset(buf, 0x00, 256);
-		len = snprintf(buf, 256, "DEALLOCATE %s", stmtHandle->stmtName);
-
-		if (len >= 255)
-			fprintf(stderr, "length of stmt name is  more than 245 bytes\n");
-
-		/* deallocate stmt */
-		tresult = Ivyexec(tconn, (const char *) buf);
-		if (IvyresultStatus(tresult) != PGRES_COMMAND_OK)
-		{
-			if (tconn->conn == NULL)
-				fprintf(stderr, "deallocate stmt %s failed because connection is gone\n", buf);
-			else
-				fprintf(stderr, "deallocate stmt %s failed\n", buf);
-		}
-
-		Ivyclear(tresult);
-
-		/* conn may be null due to other threads call Ivyclean() */
+		/*
+		 * Ivyfinish() clears and frees the PGconn while holding self_lock, so
+		 * keep it locked for the complete synchronous Close operation.  A
+		 * protocol Close uses the statement name verbatim and avoids treating
+		 * it as a possibly truncated SQL identifier.
+		 */
+		PGSemaphoreLock(&tconn->self_lock);
 		if (tconn->conn != NULL)
 		{
-			PGSemaphoreLock(&tconn->lock);
-			tconn->IvystmtHandleList = IvyremoveValueFromList(tconn->IvystmtHandleList, (void *) stmtHandle);
-			PGSemaphoreUnlock(&tconn->lock);
+			presult = PQclosePrepared(tconn->conn, stmtHandle->stmtName);
+			if (presult == NULL)
+				close_error = strdup(PQerrorMessage(tconn->conn));
+			else if (PQresultStatus(presult) != PGRES_COMMAND_OK)
+			{
+				const char *message = PQresultErrorMessage(presult);
+
+				if (message == NULL || message[0] == '\0')
+					message = PQerrorMessage(tconn->conn);
+				close_error = strdup(message);
+			}
 		}
+		else
+			close_error = strdup("connection is gone");
+
+		/*
+		 * The local relation describes ownership of stmtHandle, not whether the
+		 * server accepted Close.  Remove it on every path before stmtHandle is
+		 * freed so the connection never retains a dangling handle pointer.
+		 */
+		PGSemaphoreLock(&tconn->lock);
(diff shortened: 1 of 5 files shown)


NotHimmel:
Thanks for your PR!

NotHimmel:
@yyqdbngt Could you attach the actual .sql you ran, the raw log output, and the comparison output from the patched build?

yyqdbngt:
Sorry for the delay — here is the raw evidence, plus one thing the run turned up.

## What I ran

`src/interfaces/libpq/ivytest/testlibpq_deallocate_name.c` (whole file in the PR diff), whose server-visible statements are:

```sql
SELECT $1::int                                    -- prepared as "statement with ""quotes""; and spaces: <320 x>"
SELECT name FROM pg_prepared_statements
  WHERE name LIKE 'statement with%'               -- after IvyFreePreparedStatement
```

It runs through the existing harness: `make -C src/interfaces/libpq/ivytest oracle-check` on Ubuntu 22.04, kernel 6.8, GCC 12.3, `initdb -USYSTEM -moracle`, server started by `oracle_test.sh`.

## Raw output, patched build

```
$ make -C src/interfaces/libpq/ivytest oracle-check
...
+ ./testlibpq_deallocate_name
...
+ echo 'testlibpq_deallocate_name passed'
testlibpq_deallocate_name passed
suite_status=0
```

`results/testlibpq_deallocate_name.out`:

```
long quoted prepared statement deallocated
```

## Raw output, same test with `ivy-exec.c` reverted to master

```
+ ./testlibpq_deallocate_name
length of stmt name is  more than 245 bytes
deallocate stmt DEALLOCATE statement with "quotes"; and spaces: xxxxxxx…  failed
long quoted prepared statement was not deallocated: the server still has "statement with "quotes"; and spaces: xxxxxxxxxxxxxxxxxxxxxxxxxx"
suite_status=2
```

The leftover name is 63 bytes long. That is the interesting part.

## Why the previous revision of the test could not fail

PostgreSQL stores prepared-statement names truncated to `NAMEDATALEN - 1` bytes, so `pg_prepared_statements.name` never holds a 357-byte client-side name. A probe against a real server on the same build:

```
client-side name length = 357
PQprepare status = PGRES_COMMAND_OK
  name=[statement with "quotes"; and spaces: xxxxxxxxxxxxxxxxxxxxxxxxxx] length=63
  name = '<full 357-byte name>'            -> 0
  name LIKE 'statement with%'              -> 1

pre-patch statement: snprintf length=368, bytes sent=255
DEALLOCATE(truncated) status = PGRES_FATAL_ERROR
  ERROR:  syntax error at or near "with"
  LINE 1: DEALLOCATE statement with "quotes"; and spaces: xxxxxxxxxxxx...
  name LIKE 'statement with%'              -> 1   (still prepared)
PQclosePrepared(full name) status = PGRES_COMMAND_OK
  rows in pg_prepared_statements           -> 0
```

So the assertion `WHERE name = repeat(...)` matched **zero rows in both builds** — the test passed before the fix as well, purely because the full name never appears in the catalog. I confirmed that by running the previous revision of the test against unpatched `ivy-exec.c`: it printed the truncated-DEALLOCATE failure on stderr and still reported `passed`.

The test now matches the name prefix and fails on the old code path with the message above. `PQclosePrepared` with the full name is what the server accepts, so the code change itself needed no adjustment.

## Verification summary (same build, both directions)

| build | `oracle-check` ivytest suite | `testlibpq_deallocate_name` |
|---|---|---|
| `master` + this patch | pass (14/14) | pass |
| `master` + only this test file | fail | fail — server still has the statement |

中文摘要：补上您要的证据。用 `make -C src/interfaces/libpq/ivytest oracle-check` 在真机（Ubuntu 22.04 / kernel 6.8 / `initdb -USYSTEM -moracle`）跑：打了补丁 14/14 通过；只保留新测试、把 `ivy-exec.c` 退回 master 时该测试失败，报出服务器上残留的语句名。顺带发现并修掉一个测试自身的问题：服务端把 prepared statement 名字截断到 `NAMEDATALEN-1`（63 字节），所以旧断言用完整名字去匹配 `pg_prepared_statements` **两种构建下都是 0 行**，测试恒过；已改为按名字前缀匹配，现在旧代码路径会真正失败（上面有原始输出）。

PR #1695, opened 2026-08-16
Preserve autonomous transaction identity

### LINKED ISSUE #1625: PRAGMA AUTONOMOUS_TRANSACTION bypasses caller privileges and resolves dblink unqualified
### IvorySQL Version

### Current Behavior

Two privilege-related defects in the autonomous-transaction support (`src/pl/plisql/src/pl_autonomous.c`):

1. **Caller-privilege bypass (high, conditional)**: the autonomous session is opened via dblink with a connection string containing only `dbname/host/port` — no `user=` (lines ~483-509). With `trust`/`peer` authentication (common in dev/Docker setups) the new session connects as the server OS user (usually `postgres`, a superuser) and executes the delegated function body with those privileges, completely bypassing the caller's privileges.

2. **Unqualified dblink resolution (medium)**: `LookupFuncName(makeString("dblink")/("dblink_exec"), ...)` (lines ~343, 460) resolves through the session `search_path`. If the search_path contains a schema the attacker can write to (e.g. `public` in older databases), a fake `dblink(text, text)` can be installed and executed with the victim's privileges whenever the victim runs an autonomous-transaction procedure.

### Expected behavior/code

- Pass the current user (or the function owner by design) explicitly in the dblink connection string, and document/reject configurations where authentication is not explicit
- Resolve `dblink`/`dblink_exec` from a fixed schema (`pg_catalog`-style lookup) or via the extension owner's search_path

### Step to reproduce

1. With trust auth: create a procedure with `PRAGMA AUTONOMOUS_TRANSACTION` as a low-privilege user and call it; the body runs as the OS user.
2. With a writable `public` schema: `CREATE FUNCTION public.dblink(text, text) ...` then invoke any autonomous-transaction procedure.

btlqql:
﻿## Summary
- authenticate each autonomous dblink connection as the effective SQL user instead of relying on the server process operating-system identity
- resolve `dblink` and `dblink_exec` from the installed extension schema, independent of `search_path`
- escape database and user values when constructing libpq connection strings
- cover explicit connection identity in the existing error-path output and re-resolution of warmed dblink OID caches after `pg_proc` invalidation

Fixes #1625.

## Validation
- `git diff --check`
- Extended `plisql_autonomous` regression input and expected output for explicit `user=` and hostile shadow functions
- Repository CI: all seven required checks pass (`build`, `contrib_regression`, `meson_build`, `oracle_pg_regression`, `oracle_regression`, `pg_regression`, and `policy`).

## Compatibility and security
Autonomous connections fail closed when the effective database user cannot authenticate independently. PostgreSQL dblink rejects non-superuser connections that did not use caller-provided password credentials or delegated GSSAPI credentials; the implementation intentionally does not bypass that boundary with a higher-privilege initial connection followed by `SET ROLE`. Definer-rights routines use their effective function-owner identity through `GetUserId()`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Autonomous connections now preserve the effective database user, including definer-rights execution.
  * Database link operations reliably use the intended functions, even when similarly named functions appear earlier in the search path.
  * Connection details are safely handled during normal execution and errors.
  * Qualified procedure names are resolved more reliably.

* **Tests**
  * Added coverage for execution identity, function resolution, error output, and search-path conflicts.

### DIFF
--- a/src/pl/plisql/src/expected/plisql_autonomous.out
+++ b/src/pl/plisql/src/expected/plisql_autonomous.out
@@ -344,12 +344,10 @@ COMMIT;
 WARNING:  there is no transaction in progress
 -- This should fail gracefully with proper error message
 -- \set ON_ERROR_STOP off
+\set VERBOSITY terse
 SELECT test_function_error();
 ERROR:  relation "nonexistent_table" does not exist
-CONTEXT:  PL/iSQL function test_function_error() line 6 at SQL statement
-while executing query on unnamed dblink connection
-SQL statement "SELECT * FROM dblink('dbname=''pl_regression'' port=58928', 'SET ivorysql.compatible_mode = oracle; SET plisql.inside_autonomous_transaction = true; SELECT public.test_function_error();') AS t(result pg_catalog.int4)"
-PL/iSQL function test_function_error() during function entry
+\set VERBOSITY default
 -- \set ON_ERROR_STOP on
 --
 -- Test 16: Function with pass-by-reference return type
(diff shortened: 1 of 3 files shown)


btlqql:
@coderabbitai review

NotHimmel:
@btlqql  Thanks for the PR. Before reviewing it: did you reproduce the problem yourself, and can you show how? And after applying the patch, did you verify the problem is actually gone?

btlqql:
我补了一条专门验证连接身份的回归，并按相同源码、相同测试做了基线/补丁对照。

验证方式：两组都运行 `make -C src/pl/plisql/src oracle-check REGRESS=plisql_autonomous`；基线组只把 `src/pl/plisql/src/pl_autonomous.c` 恢复为当前 `master` 的 `c3529ba4789ee220caefc858e58bbe6f1563ad14`，补丁组使用本 PR 最新提交 `07b38ee81f`。

结果：
- 基线组稳定复现：`session_user = current_user` 断言得到 `false`，伪造的 `dblink` 被命中并返回 `999`，聚焦回归 1/1 失败。
- 补丁组：`plisql_autonomous` 1/1 通过。

完整对照运行： https://github.com/btlqql/IvorySQL/actions/runs/32152306532

新增断言直接检查新连接的 `session_user`，因此不能被连接后执行的 `SET ROLE` 掩盖。测试中的临时角色带 `LOGIN SUPERUSER`，仅用于允许隔离回归实例里的无密码 dblink 连接，清理阶段会删除。

btlqql:
@coderabbitai review

NotHimmel:
Closing this: dblink gives a separate session where an autonomous transaction needs a separate transaction in the same session — the parent's locks are invisible across that boundary, so no change to the connection can make this correct.

PR #1700, opened 2026-08-17
fix(oracle): mark NLS-dependent TO_CHAR overloads stable

### LINKED ISSUE #1696: NLS-dependent TO_CHAR overloads are incorrectly marked IMMUTABLE
## Bug

The Oracle-compatible `sys.to_char()` overloads for date and timestamp types are declared `IMMUTABLE`, although their results depend on session settings such as `nls_date_format`, `nls_timestamp_format`, `nls_timestamp_tz_format`, and `timezone`.

This lets PostgreSQL accept expression indexes whose ordering becomes stale when those session settings change.

## Reproduction

```sql
SET nls_date_format = 'YYYY-MM-DD';
CREATE TABLE nls_volatility_test(d sys.oradate);
INSERT INTO nls_volatility_test VALUES
  (sys.to_date('2024-01-31', 'YYYY-MM-DD')),
  (sys.to_date('2024-02-01', 'YYYY-MM-DD'));
CREATE INDEX nls_volatility_idx
  ON nls_volatility_test (sys.to_char(d));

SET nls_date_format = 'DD/MM/YYYY';
SET enable_seqscan = off;
SELECT sys.to_char(d)
FROM nls_volatility_test
ORDER BY sys.to_char(d);
```

The index scan returns:

```text
31/01/2024
01/02/2024
```

A sequential scan and sort returns the correct order:

```text
01/02/2024
31/01/2024
```

## Expected behavior

NLS- and timezone-dependent `TO_CHAR` overloads should be marked `STABLE`, so they cannot be used in expression indexes that assume immutable results.

## Proposed fix

- mark the date/timestamp `sys.to_char()` overloads as `STABLE`
- add regression coverage for function volatility and rejection of unsafe expression indexes

btlqql:
## What changed

- mark all 12 Oracle-compatible `TO_CHAR` overloads for `DATE`, `TIMESTAMP`, `TIMESTAMP WITH TIME ZONE`, and `TIMESTAMP WITH LOCAL TIME ZONE` as `STABLE`
- add catalog coverage for every affected overload
- verify that PostgreSQL rejects expression indexes using the session-dependent one-argument overload

## Root cause

These overloads were declared `IMMUTABLE`, although their results depend on session NLS format/language and timezone settings. That allowed PostgreSQL to accept expression indexes whose stored keys could become inconsistent after the session settings changed.

The issue was reproduced by building an index on `sys.to_char(d)` under `nls_date_format = 'YYYY-MM-DD'`, changing it to `DD/MM/YYYY`, and observing that an index scan returned a different order from a sequential scan plus sort.

## Validation

Run on `WZU_Server` with the same configure options used by the Oracle regression workflow:

- `make -j16`
- `make -C contrib/ivorysql_ora oracle-check ORA_REGRESS=ora_datetime_datatype_functions`
- `make oracle-check-world`
- `git diff --check`

All completed successfully. The focused regression confirms all 12 overloads are `STABLE` and that the unsafe expression index is rejected.

Closes #1696

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected date and timestamp formatting functions to accurately reflect their stability when results may depend on session settings.
  * Prevented these formatting expressions from being used where immutable behavior is required, such as index expressions.

* **Tests**
  * Added regression coverage for all supported date and timestamp formatting overloads.
  * Added verification that these expressions cannot be used in indexes.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
@@ -204,6 +204,36 @@ SELECT to_char(
 (1 row)
 
 DROP TABLE to_char_tbl;
+-- TO_CHAR volatility must reflect its session NLS and timezone dependencies.
+SELECT count(*) AS stable_to_char_overloads
+FROM pg_proc
+WHERE oid = ANY (ARRAY[
+    'sys.to_char(sys.oradate)'::regprocedure,
+    'sys.to_char(sys.oradate,text)'::regprocedure,
+    'sys.to_char(sys.oradate,text,text)'::regprocedure,
+    'sys.to_char(sys.oratimestamp)'::regprocedure,
+    'sys.to_char(sys.oratimestamp,text)'::regprocedure,
+    'sys.to_char(sys.oratimestamp,text,text)'::regprocedure,
+    'sys.to_char(sys.oratimestamptz)'::regprocedure,
+    'sys.to_char(sys.oratimestamptz,text)'::regprocedure,
+    'sys.to_char(sys.oratimestamptz,text,text)'::regprocedure,
+    'sys.to_char(sys.oratimestampltz)'::regprocedure,
+    'sys.to_char(sys.oratimestampltz,text)'::regprocedure,
+    'sys.to_char(sys.oratimestampltz,text,text)'::regprocedure
+])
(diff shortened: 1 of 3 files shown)


btlqql:
Superseded by #1701, submitted from the replacement account.

PR #1704, opened 2026-08-18
Make sys.to_number accept scientific notation and blank padding

### LINKED ISSUE #1703: sys.to_number does not accept scientific notation (e.g. '1e3')
## Bug Report

   ### IvorySQL Version

   master

   ### Current Behavior

   `sys.to_number` rejects scientific notation and trailing blanks, while Oracle's `TO_NUMBER` accepts them:

   ```sql
   [ORA]postgres=# SELECT sys.to_number('1e3'::varchar2);
   ERROR:  invalid number format model

   [ORA]postgres=# SELECT sys.to_number('1E3'::varchar2);
   ERROR:  invalid number format model

   [ORA]postgres=# SELECT sys.to_number('-1.5e-2'::varchar2);
   ERROR:  invalid number format model

   [ORA]postgres=# SELECT sys.to_number('12 '::varchar2);
   ERROR:  invalid number format model
 ```

 ### Expected behavior/code

 same as Oracle:

 ```sql
   SELECT TO_NUMBER('1e3')      FROM dual;   -- 1000
   SELECT TO_NUMBER('1E3')      FROM dual;   -- 1000
   SELECT TO_NUMBER('-1.5e-2')  FROM dual;   -- -.015
   SELECT TO_NUMBER('12 ')      FROM dual;   -- 12 (trailing blanks ignored)
   SELECT TO_NUMBER(' 12')      FROM dual;   -- 12 (leading blanks ignored, already works)
 ```

 ### Step to reproduce

 1. Connect in Oracle compatibility mode
 2. Run the queries above

xuhuicai:
ora_to_number_internal()'s unformatted branch only accepted digits, a single decimal point and a leading sign, so inputs like '1e3' (which Oracle's TO_NUMBER accepts and PostgreSQL's numeric input already parses) were rejected with "invalid number format model".  Trailing blanks were also rejected in the integer part while accepted after the decimal point.

Parse an optional exponent ('e'/'E' followed by an optional sign and digits) and skip leading and trailing blanks, matching Oracle.  Derive the numeric typmod from the mantissa digits adjusted by the exponent so the result keeps the right scale (e.g. '1e3' -> numeric(4,0) -> 1000, '-1.5e-2' -> numeric(3,3) -> -0.015).

Add regression coverage for scientific notation, blank padding and malformed exponents in ora_number.sql.

Closed #1703 

Assisted-by: Cursor
Percentage of AI-generated code: 95%

## Summary by CodeRabbit

- **New Features**
  - Improved numeric conversion for scientific notation, signed exponents, decimal values, optional signs, and surrounding whitespace.
  - Leading zeros in exponents are handled correctly, including positive and negative exponent values.

- **Bug Fixes**
  - Numeric conversions now handle precision, scale, overflow, and underflow consistently.
  - Malformed, incomplete, empty, and unsupported numeric inputs return appropriate errors.

- **Tests**
  - Expanded regression coverage for whitespace, special values, decimal variants, zero exponents, exponent boundaries, and malformed inputs.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_number.out
+++ b/contrib/ivorysql_ora/expected/ora_number.out
@@ -1750,3 +1750,197 @@ SELECT BITAND('6',8) re FROM DUAL;
 (1 row)
 
 /* End - bug0000478 */
+/* Begin - to_number accepts scientific notation and blank padding */
+SELECT to_number('1e3') AS a;
+  a   
+------
+ 1000
+(1 row)
+
+SELECT to_number('1E3') AS b;
+  b   
+------
+ 1000
+(1 row)
+
+SELECT to_number('-1.5e-2') AS c;
+   c    
+--------
+ -0.015
+(1 row)
+
+SELECT to_number('1.5e+3') AS d;
+  d   
+------
+ 1500
+(1 row)
+
+SELECT to_number('1E-3') AS e;
+   e   
+-------
+ 0.001
+(1 row)
+
+SELECT to_number(' 1e3 ') AS f;
+  f   
+------
+ 1000
+(1 row)
+
+SELECT to_number('12 ') AS g;
+ g  
+----
+ 12
+(1 row)
+
+SELECT to_number(' 12') AS h;
+ h  
+----
+ 12
+(1 row)
+
+SELECT to_number('12.345') AS i;
+   i    
+--------
+ 12.345
+(1 row)
+
+SELECT to_number('-1122.34') AS j;
+    j     
+----------
+ -1122.34
+(1 row)
+
+/* leading zeros in the exponent must be honored, not truncated */
+SELECT to_number('1e0000000010') AS r;
+      r      
+-------------
+ 10000000000
+(1 row)
+
+SELECT to_number('1e-0000000010') AS s;
+      s       
+--------------
+ 0.0000000001
+(1 row)
+
+/* positive exponents beyond numeric's range keep raising an error */
+SELECT to_number('1e2147483648') AS t;
+ERROR:  value overflows numeric format
+SELECT to_number('1e10000000000000000000') AS v;
+ERROR:  value overflows numeric format
+/* ... but an all-zero mantissa times any exponent is still zero */
+SELECT to_number('0e2147483648') AS zero_mantissa;
+ zero_mantissa 
+---------------
+ 0
+(1 row)
+
+SELECT to_number('0e10000000000000000000') AS zero_mantissa_long;
+ zero_mantissa_long 
+--------------------
+ 0
+(1 row)
+
+SELECT to_number('1e1073741823') AS at_boundary;
+ERROR:  numeric field overflow
+DETAIL:  A field with precision 1000, scale 0 must round to an absolute value less than 10^1000.
+SELECT to_number('1e1073741824') AS one_over_boundary;
+ERROR:  value overflows numeric format
+SELECT to_number('1e2000') AS clamp_precision;
+ERROR:  numeric field overflow
+DETAIL:  A field with precision 1000, scale 0 must round to an absolute value less than 10^1000.
+/* precision-clamp path without an exponent */
+SELECT to_number(repeat('9', 1500)) AS long_int_no_exp;
+ERROR:  numeric field overflow
+DETAIL:  A field with precision 1000, scale 0 must round to an absolute value less than 10^1000.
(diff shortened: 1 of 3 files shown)


hanjianqiao:
Thanks for your contribution! Could please take a look at the comment by coderabbitai.

xuhuicai:
@hanjianqiao Thanks for your review! Done.

hanjianqiao:
Could you please remove the sys. prefix from these tests? Oracle itself rejects SYS.TO_NUMBER(...) — these built-ins live in the SYS.STANDARD package, not as standalone SYS objects. Since the point of these tests is to verify Oracle compatibility, they should use the same syntax that real Oracle accepts.

xuhuicai:
@hanjianqiao Thanks for the review! Done — I removed the `sys.` prefix from all the new `to_number` tests so they use the same bare `TO_NUMBER(...)` syntax that real Oracle accepts.

hanjianqiao:
@xuhuicai Nice fix! Three suggestions:

1. Remove type-cast(::varchar2, etc) from tests (please limits modifcation within your PR).
2. Add more tests to cover more cases:
```text
Exponent overflow boundary — the check is exponent > PG_INT32_MAX / 2 (i.e. 1073741823). Right at that value it still passes this function's own check, but the typmod gets clamped to
NUMERIC_MAX_PRECISION (1000) and numeric_in() itself then rejects it with a different error:
SELECT to_number('1e1073741823') AS at_boundary;      -- ERROR: numeric field overflow (from numeric_in)
SELECT to_number('1e1073741824') AS one_over_boundary; -- ERROR: value overflows numeric format (this function's own check)

Precision-clamp path — currently untested. When exponent/digit count pushes precision above 1000, it gets clamped rather than erroring outright, but the actual value still doesn't fit, so numeric_in()
errors out:
SELECT to_number('1e2000') AS clamp_precision;               -- ERROR: numeric field overflow
SELECT to_number(repeat('9', 1500)) AS long_int_no_exp;       -- ERROR: numeric field overflow (non-exponent path)

Malformed exponent variants beyond the ones already covered:
SELECT to_number('1 e3') AS space_before_e;   -- ERROR: invalid number format model
SELECT to_number('1e 3') AS space_after_e;    -- ERROR: invalid number format model
SELECT to_number('1e3e4') AS double_e;        -- ERROR: invalid number format model
SELECT to_number('1e--3') AS double_sign;     -- ERROR: invalid number format model
SELECT to_number('1e+-3') AS mixed_sign;      -- ERROR: invalid number format model

Empty/blank/sign-or-dot-only input — worth calling out that to_number('') returns NULL, not an error, because ''::varchar2 IS NULL under Oracle semantics — the function body never even runs (STRICT
short-circuit). The other cases correctly hit the "no digits" check:
SELECT to_number(''::varchar2) AS empty_str;  -- NULL 
SELECT to_number('') AS empty_str;
SELECT to_number('   ') AS all_blank;         -- ERROR: invalid number format model
SELECT to_number('-') AS bare_sign;           -- ERROR: invalid number format model
SELECT to_number('.') AS bare_dot;            -- ERROR: invalid number format model

One-sided decimal point (currently only interior dots are tested):
SELECT to_number('123.') AS trailing_dot;  -- 123
SELECT to_number('.5') AS leading_dot;     -- 0.5

Zero exponent, three spellings:
SELECT to_number('1e0') AS exp_zero;        -- 1
SELECT to_number('1e+0') AS exp_plus_zero;  -- 1
SELECT to_number('1e-0') AS exp_minus_zero; -- 1

Tokens numeric_in() would happily accept but Oracle's TO_NUMBER shouldn't — good to lock in that this layer rejects them before they ever reach numeric_in():
SELECT to_number('NaN') AS nan_case;        -- ERROR: invalid number format model
SELECT to_number('Infinity') AS inf_case;   -- ERROR: invalid number format model
SELECT to_number('-Infinity') AS neg_inf_case; -- ERROR: invalid number format model
```
3. The following SQL statements return 0 in Oracle. Could you make our behavior consistent with Oracle?
```sql
SELECT to_number('1e-2147483648') AS u;
SELECT to_number('1e-10000000000000000000') AS w;
```
`SELECT to_number('1e2147483648');` and `SELECT to_number('1e10000000000000000000');` should keep raising an error, as Oracle does.

hanjianqiao:
@xuhuicai Could you take a look at the comments from coderabbitai? Also, there are merge conflicts that need resolving.

PR #1715, opened 2026-08-20
Declare exit_nicely(void) and drop the ignored argument

### LINKED ISSUE #1707: `exit_nicely()` with an extra, silently-ignored argument
### IvorySQL Version

master

### Current Behavior

`exit_nicely()` is defined using the old K&R-style empty parameter list (`()` instead of `(void)`), so GCC does not treat it as a prototype and skips argument-count checking at call sites. Confirmed locally with the project's actual CFLAGS (`-Wall -Wstrict-prototypes -Wold-style-definition -Wmissing-prototypes`, etc.) — the build only emits "old-style function definition" / "function declaration isn't a prototype" warnings for `exit_nicely()`, `init_statement()`, and `main()`, and nothing about the mismatched call. So `exit_nicely(conn)` builds cleanly today, which is why this went unnoticed.

xuhuicai:
exit_nicely() used an old-style empty parameter list, so the compiler skipped argument-count checks and the errant exit_nicely(conn) call was silently accepted. Make it a proper (void) prototype and remove the stray argument.
Closes #1707 

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated internal connection cleanup handling during test data initialization.
  * No user-visible behavior changes.

### DIFF
--- a/src/interfaces/libpq/ivytest/testlibpq.c
+++ b/src/interfaces/libpq/ivytest/testlibpq.c
@@ -85,7 +85,7 @@ static void run_stmt(Ivyconn *tconn,IvyPreparedStatement *stmt, Ivyargmode *argm
 
 
 static void
-exit_nicely()
+exit_nicely(void)
 {
 	Ivyfinish(thread_data.tconn1);
 	Ivyfinish(thread_data.tconn3);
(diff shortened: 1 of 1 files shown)


NotHimmel:
LGTM

faweizhao26:
Thanks for your contribution. Please submit this [form](https://jsj.top/f/CexUN3) and we will send you the swags.

PR #1716, opened 2026-08-20
Fix missing owner check for CREATE PACKAGE BODY (#1708)

### LINKED ISSUE #1708: First time `CREATE PACKAGE BODY` without owner check
### IvorySQL Version
master

### Current Behavior

role_attacker  create role_owner's pkg body successfully.

### Expected behavior/code

when role_attacker try to create role_owner's pkg body, some error like: `ERROR: permission denied for package "victim_pkg"`

### Step to reproduce

```sql
CREATE SCHEMA shared_ns;
CREATE ROLE role_owner LOGIN;
CREATE ROLE role_attacker LOGIN;
GRANT CREATE, USAGE ON SCHEMA shared_ns TO role_owner, role_attacker;

SET ROLE role_owner;
SELECT session_user, current_user;
CREATE PACKAGE shared_ns.victim_pkg IS
  FUNCTION get_secret RETURN INTEGER;
END victim_pkg;
/

SET ROLE role_attacker;
SELECT session_user, current_user;
CREATE PACKAGE BODY shared_ns.victim_pkg IS
  FUNCTION get_secret RETURN INTEGER IS
  BEGIN
    RETURN 1337;
  END;
END victim_pkg;
/

SET ROLE role_owner;
SELECT session_user, current_user;
SELECT shared_ns.victim_pkg.get_secret();
```

role_owner execute the role_hacker's function
```bash
[ORA]postgres=> SELECT shared_ns.victim_pkg.get_secret();
 get_secret 
------------
       1337
(1 row)
```

anapplebeee:
Fixes #1708

## Bug Fixes
- Improved package ownership validation when creating or replacing package bodies.
- Prevented unauthorized roles from creating or replacing another role’s package body.
- Preserved package owners’ control over package implementations in shared schemas.

## Tests
- Added regression coverage for cross-role package ownership.
- Verified unauthorized package-body changes are rejected.
- Confirmed package owners can access packages that remain without an unauthorized body.

### DIFF
--- a/src/backend/commands/packagecmds.c
+++ b/src/backend/commands/packagecmds.c
@@ -362,6 +362,12 @@ CreatePackageBody_internal(const char *pkgname,
 	if (pkg_editable != editable)
 		elog(ERROR, "The EDITIONABLE property of a TYPE or PACKAGE specification and its"
 					" body must match");
+    /*
+     * Permission check: must own package or superuser
+     */
+	if (!pg_package_ownercheck(pkgOid, GetUserId()))
+		aclcheck_error(ACLCHECK_NOT_OWNER, OBJECT_PACKAGE_BODY,
+					pkgname);
 
 	/*
 	 * All seems OK; prepare the data to be inserted into sys_package_body.
@@ -391,10 +397,6 @@ CreatePackageBody_internal(const char *pkgname,
 
 		if (!replace)
 			elog(ERROR, "package \"%s\" body already exists", pkgname);
-
-		if (!pg_package_ownercheck(pkgOid, GetUserId()))
-			aclcheck_error(ACLCHECK_NOT_OWNER, OBJECT_PACKAGE_BODY,
-						   pkgname);
 
 		bodydatum = SysCacheGetAttr(PKGBODYOID, oldtup,
 								  Anum_pg_package_body_bodysrc, &isnull);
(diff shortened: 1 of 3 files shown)


NotHimmel:
@anapplebeee Thanks for the fix! Could you please add `Fixes #1708` to the PR  description?

anapplebeee:
@NotHimmel Thanks for the reminder! I've added `Fixes #1708` to the PR description.

NotHimmel:
I think we only need one unauthorized operation for this regression test: after setting up `role_owner`, `role_attacker`, and creating `shared_ns.victim_pkg` as `role_owner`, switch to `role_attacker` and verify that the first `CREATE PACKAGE BODY` fails with the expected ownership error.

The additional `CREATE OR REPLACE PACKAGE BODY` case is redundant because no body exists after the preceding failed `CREATE PACKAGE BODY`, so both statements exercise the same first-creation path.

The `SELECT current_user` checks can also be removed since `SET ROLE` is sufficient here.

anapplebeee:
> I think we only need one unauthorized operation for this regression test: after setting up `role_owner`, `role_attacker`, and creating `shared_ns.victim_pkg` as `role_owner`, switch to `role_attacker` and verify that the first `CREATE PACKAGE BODY` fails with the expected ownership error.
> 
> The additional `CREATE OR REPLACE PACKAGE BODY` case is redundant because no body exists after the preceding failed `CREATE PACKAGE BODY`, so both statements exercise the same first-creation path.
> 
> The `SELECT current_user` checks can also be removed since `SET ROLE` is sufficient here.

thanks review, I removed the redundant test case

faweizhao26:
Thanks for your contribution. Please submit this [form](https://jsj.top/f/CexUN3) and we will send you the swags.

PR #1738, opened 2026-08-26
Fix trailing semicolons in Oracle psql slash commands

### LINKED ISSUE #1742: Oracle psql slash commands include trailing semicolons in object names
## Bug description

In Oracle-compatible psql mode, slash commands that consume a whole-line object name can treat a trailing semicolon as part of that name. For example, `\sf public.some_function ;` may look up `public.some_function ;` instead of the intended function.

The PostgreSQL slash scanner honors its `semicolon` flag in whole-line mode, while the parallel Oracle scanner currently leaves the terminator in the argument.

## Expected behavior

Commands such as `\sf`, `\ef`, `\sv`, and `\ev` should tolerate trailing whitespace and a statement-terminating semicolon in Oracle mode.

## Proposed fix

Bring the Oracle whole-line scanner behavior in line with the PostgreSQL scanner and add a psql TAP regression test. A proposed implementation is available in #1738.

btlqql:
## Summary

- make the Oracle slash-command scanner honor the `semicolon` flag for whole-line arguments
- strip trailing whitespace and semicolons for commands such as `\sf`, `\ef`, `\sv`, and `\ev`
- add an Oracle-mode psql TAP regression for `\sf function_name ;`

## Problem

The PostgreSQL slash scanner handles the flag in `OT_WHOLE_LINE` mode, but the parallel Oracle scanner still treated the whole argument as-is. A harmless trailing semicolon therefore became part of the function or view name and caused lookup failures only in Oracle mode.

## Validation

- `git diff --check`
- verified the Oracle scanner now matches the established PostgreSQL scanner behavior
- the Linux-only psql TAP test cannot run in this Windows checkout and is included for CI

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Oracle-compatible `\sf` command parsing to correctly handle trailing whitespace and semicolons.
  - Function definitions now display as expected when `\sf` commands include a trailing semicolon.

- **Tests**
  - Added regression coverage for Oracle-mode `\sf` commands with trailing whitespace and semicolons.

Fixes #1742

### DIFF
--- a/src/bin/psql/ora_psqlscanslash.l
+++ b/src/bin/psql/ora_psqlscanslash.l
@@ -22,6 +22,8 @@
  */
 #include "postgres_fe.h"
 
+#include <ctype.h>
+
 #include "common.h"
 #include "ora_psqlscanslash.h"
 #include "common/logging.h"
@@ -620,7 +622,7 @@ ora_psql_scan_slash_option(PsqlScanState state,
 			/* empty arg */
 			break;
 		case xslasharg:
-			/* Strip any unquoted trailing semi-colons if requested */
+			/* Strip any unquoted trailing semicolons if requested */
 			if (semicolon)
 			{
 				while (unquoted_option_chars-- > 0 &&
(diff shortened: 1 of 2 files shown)


hs-liuxh:
Thanks for your contribution.

PR #1739, opened 2026-08-26
Fix the Oracle prompt service-name escape

### LINKED ISSUE #1743: Oracle psql prompt %s displays the backend PID instead of the service name
## Bug description

In Oracle-compatible psql mode, the `%s` prompt escape does not reliably display the configured service name after connecting. The Oracle prompt implementation copies the `SERVICE` variable and then falls through into the `%p` backend-PID branch, so the PID overwrites the service-name output.

## Reproduction

1. Start psql in Oracle-compatible mode and connect to a server.
2. Set `SERVICE` to a recognizable value.
3. Set `PROMPT1` to `%s`.
4. Observe that the prompt displays the backend PID instead of the service value.

## Expected behavior

`%s` should display the configured `SERVICE` value, while `%p` should remain the separate backend-PID escape.

## Proposed fix

Terminate the `%s` branch normally and add an interactive psql TAP regression test. A proposed implementation is available in #1739.

yyqdbngt:
## Summary

- stop the Oracle prompt `%s` service-name escape from falling through to `%p`
- preserve the configured `SERVICE` value while connected
- extend the interactive Oracle-mode prompt TAP test

## Problem

`ora_get_prompt()` copied the service name into the prompt buffer and then continued into the backend-PID branch. With an active connection, `%s` consequently rendered the PID instead of the `SERVICE` psql variable. The PostgreSQL prompt implementation terminates the `%s` branch normally.

## Validation

- `git diff --check`
- verified the service branch terminates before the backend-PID branch
- the interactive TAP test requires the Linux PostgreSQL test harness and `IO::Pty`, which are unavailable in this Windows checkout; CI should execute it

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Oracle-mode prompt so the `%s` escape displays the configured service name correctly.
  * Prevented the service-name prompt escape from incorrectly showing backend process information.

* **Tests**
  * Added coverage verifying the corrected prompt behavior while connected.

Fixes #1743

### DIFF
--- a/src/bin/psql/ora_prompt.c
+++ b/src/bin/psql/ora_prompt.c
@@ -192,7 +192,7 @@ ora_get_prompt(Ora_promptStatus_t status, ConditionalStack cstack)
 						if (service_name)
 							strlcpy(buf, service_name, sizeof(buf));
 					}
-					pg_fallthrough;
+					break;
 					/* backend pid */
 				case 'p':
 					if (pset.db)
--- a/src/bin/psql/t/040_prompt.pl
+++ b/src/bin/psql/t/040_prompt.pl
@@ -110,6 +110,21 @@ like(
 	qr/BEFORE\[ORA\]AFTER/,
 	"%o expands to [ORA] when compatible_mode is oracle");
 
+# The Oracle prompt renderer must not let the service-name escape fall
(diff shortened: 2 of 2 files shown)


hs-liuxh:
Thanks for your contribution.

PR #1746, opened 2026-08-26
Compatible with Oracle LENGTHC

### LINKED ISSUE #1747: Compatible with Oracle LENGTHC

hs-liuxh:
close #1747 

## Summary by CodeRabbit

* **New Features**
  * Added Oracle-compatible `LENGTHC` support for character, numeric, date, timestamp, and binary values.
  * Counts complete Unicode characters after NFC normalization, including supplementary characters and combining marks.
  * Supports Oracle character types, NULL and empty values, and text-based domains.
  * Rejects unsupported CLOB, NCLOB, and BLOB inputs with clear datatype errors.

* **Tests**
  * Expanded coverage for normalization, padding, overloads, binary inputs, LOB handling, and expression-index compatibility.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
@@ -32,11 +32,13 @@
 #include "varatt.h"
 
 #include "access/detoast.h"
+#include "common/unicode_norm.h"
 #include "lib/stringinfo.h"
 #include "mb/pg_wchar.h"
 #include "miscadmin.h"
 #include "regex/regex.h"
 #include "utils/builtins.h"
+#include "utils/memutils.h"
 #include "utils/numeric.h"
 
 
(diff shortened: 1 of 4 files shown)


contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
```diff
@@ -52,6 +54,7 @@ PG_FUNCTION_INFO_V1(oracharlen);
 PG_FUNCTION_INFO_V1(oracharoctetlen);
 PG_FUNCTION_INFO_V1(oravarcharlen);
 PG_FUNCTION_INFO_V1(oravarcharoctetlen);
+PG_FUNCTION_INFO_V1(ora_lengthc);
```
hanjianqiao, line 57: Place the declaration under `PG_FUNCTION_INFO_V1(ora_listagg_check);` may be better

hs-liuxh: Thanks for your comments. There is a common rule. This block tracks the order the functions are defined in — ora_lengthc_unsupported is defined at line 319, right after ora_lengthc (197) and before rtrim1 (339), so the declaration sits in the matching spot. ora_listagg_check is last in the block because its definition is the last function in the file.

PR #1748, opened 2026-08-26
Add Oracle-compatible UTL_URL.UNESCAPE

### LINKED ISSUE #1061: oracle compatibility feature: UTL_URL.UNESCAPE
## Feature Request
oracle compatibility feature: UTL_URL.UNESCAPE
UTL_URL.UNESCAPE decodes URL-encoded strings by converting %XX escape sequences back to their original characters, commonly used to process web request parameters.

muzimu217:
## Summary

- add UTL_URL.UNESCAPE to the ivorysql_ora extension, the decoding half
  of the UTL_URL package: every %XX group becomes the byte it encodes,
  other characters pass through, and the reassembled byte string is
  converted from url_charset to the database encoding, so consecutive
  %XX groups of a multibyte character reassemble into that character
- stacked on top of the UTL_URL.ESCAPE change (#1728); the package is
  only complete with both halves
- note: until #1728 is merged, the diff below also shows the ESCAPE
  commits; it collapses to the UNESCAPE-only change once #1728 lands
- the regression test gains the unescape cases, the double-escape
  decode ("a%2520b" -> "a%20b"), every malformed-sequence error, and
  the round-trip identity UNESCAPE(ESCAPE(x)) = x in both modes

Behaviour verified against real Oracle XE 11.2 and 21 with the same
probe script shared in #1728; the behaviour matches, with deviations
carried over from the ESCAPE change and documented there:

- a NULL url_charset means "database encoding, no conversion" (Oracle
  defaults to utl_http.body_charset / ISO-8859-1)
- %00 is rejected because text cannot contain a NUL byte, where Oracle
  returns a string containing a NUL byte
- a %XX group that does not form a valid character in url_charset
  surfaces as a PostgreSQL encoding error ("invalid byte sequence"),
  not as Oracle's internal UTL_URL error
- charset names use the IANA/PostgreSQL spellings (GBK, LATIN1, ...);
  Oracle-specific spellings such as ZHS16GBK are not recognized, but
  the IANA names are valid on Oracle too

A "%" that is not followed by two hexadecimal digits raises an error,
matching Oracle's UTL_URL.BAD_URL (ORA-29262); an unrecognized charset
is rejected (Oracle reports ORA-01482).

## Testing

- oracle regression suite passes: 25/25 in both a UTF8 and a C locale
- make check: 249/249
- meson build of the full tree: 2728/2728

## Reference

- Oracle UTL_URL documentation:
  https://docs.oracle.com/en/database/oracle/oracle-database/19/arpls/UTL_URL.html
- RFC 2396: https://datatracker.ietf.org/doc/html/rfc2396

Fixes #1061

## Summary by CodeRabbit

* **New Features**
  * Added Oracle-compatible `UTL_URL` support.
  * Added URL escaping and unescaping, including reserved-character handling and character-set conversion.
  * Added validation for malformed escape sequences and invalid URL data.
  * Added `DBMS_TRANSACTION` support and `ORA_VSIZE` coverage.

* **Tests**
  * Added comprehensive coverage for null, empty, multibyte, character-set, round-trip, and error scenarios.
  * Added regression coverage using UTF8 databases.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -33,6 +33,7 @@ OBJS = \
 	src/builtin_packages/dbms_utility/dbms_utility.o \
 	src/builtin_packages/dbms_lock/dbms_lock.o \
 	src/builtin_packages/utl_encode/utl_encode.o \
+	src/builtin_packages/utl_url/utl_url.o \
 	src/builtin_packages/dbms_session/dbms_session.o
 
 EXTENSION = ivorysql_ora
@@ -84,6 +85,7 @@ ORA_REGRESS = \
 	utl_file \
 	utl_raw \
 	utl_encode \
+	utl_url \
 	dbms_session
 
 # Include path for plisql.h (needed by dbms_utility)
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -61,6 +61,11 @@ cleansql:
 
 # REGRESS_OPTS mainly used for installcheck or debug.
 #REGRESS_OPTS = --port=1521
+# The utl_url regression tests exercise multibyte input and charset
+# conversions, so pin the test database to UTF8 instead of relying on
+# whatever encoding initdb derives from the invoking environment's locale.
+ENCODING = UTF8
+
 ORA_REGRESS = \
 	ora_character \
 	ora_datetime \
--- a/contrib/ivorysql_ora/expected/utl_url.out
+++ b/contrib/ivorysql_ora/expected/utl_url.out
@@ -187,6 +187,21 @@ SELECT utl_url.escape('a b', FALSE, 'ISO-8859-1') AS iana_name,
  a%20b     | a%20b
 (1 row)
 
+-- The url_charset contract: omitting the argument and passing an explicit
+-- NULL are the same thing here -- both mean "the database encoding, do not
+-- convert".  Oracle makes them different: an omitted url_charset defaults
+-- to utl_http.body_charset (ISO-8859-1) while an explicit NULL selects the
+-- database character set.  Defaulting to the database encoding instead of
+-- ISO-8859-1 is the declared deviation of this port (ISO-8859-1 cannot
+-- represent non-Latin text; see the package documentation), and this case
+-- pins the contract so the difference stays visible.
+SELECT utl_url.escape('é') AS default_charset,
+       utl_url.escape('é', FALSE, NULL) AS explicit_null_charset;
+ default_charset | explicit_null_charset 
+-----------------+-----------------------
+ %C3%A9          | %C3%A9
+(1 row)
+
 -- An unrecognized character set is rejected
 SELECT utl_url.escape('a b', FALSE, 'NO_SUCH_CHARSET');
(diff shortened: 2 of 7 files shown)


hs-liuxh:
Thanks for your contribution. Please resolve the code conflicts first.

muzimu217:
Conflicts resolved. The branch now merges the current master, which added DBMS_TRANSACTION and touched the same registration files; both packages are registered side by side in the Makefile / meson.build / merge_sqls lists.

While re-verifying on top of the merged master I also addressed the review-bot finding on UNESCAPE: literal characters (which are already in the database encoding) are no longer passed through the url_charset conversion together with the %XX-decoded bytes. Previously `utl_url.unescape('é%20', 'LATIN1')` on a UTF8 database mangled the literal `é`; now only the `%XX` runs are converted from url_charset, which also matches Oracle, where the URL string is in the database character set and only the escape sequences are decoded through url_charset. A regression test mixing a UTF-8 literal with a LATIN1 `%XX` group pins the behaviour.

Re-verified on the merged tree: the oracle regression suite passes 27/27 in both a C and a UTF8 locale (now including the dbms_transaction tests), and top-level `make check` passes except the numa test, which fails in this environment with `Operation not permitted` both with and without these changes.

muzimu217:
The latest commits address the review feedback and pin the url_charset contract:

- The review-bot finding on UNESCAPE is fixed (9c16b43c): literal characters, which are already in the database encoding, are no longer converted through url_charset together with the %XX-decoded bytes — only the %XX runs are converted now.  This also matches Oracle, where the URL string is in the database character set and only the escape sequences are decoded through url_charset.  A regression test mixing a UTF-8 literal with a LATIN1 %XX group (`unescape('café%21', 'LATIN1')`) pins the behaviour, so the bot should be able to verify the resolution against the updated utl_url.c.
- The url_charset contract is pinned on both sides: an omitted url_charset and an explicit NULL are equivalent and both mean the database encoding, where Oracle's omitted form defaults to utl_http.body_charset (ISO-8859-1).  Tests cover the omitted/NULL equivalence, a lone LATIN1 %E9 group through the explicit charset, and the behaviour of a lone %E9 under the database-encoding default; the package documentation spells out the same contract.

Re-verified on the updated branch: the oracle regression suite passes 27/27 in both a C and a UTF8 locale, top-level make check passes except the numa test (which fails in this environment regardless of these changes), and the full tree builds cleanly under Meson/Ninja (2729/2729 targets).

Could you also approve the pending workflow runs for this PR, so the build and regression checks can run against the branch?

muzimu217:
Hi, gentle nudge on this one — the conflict from August was resolved a while back and the branch is currently merging cleanly with checks green. If there is anything you would like changed, happy to fix it — otherwise no rush. Thanks for taking a look!

hs-liuxh:
Got it. I've finished reviewing this PR and have no further issues. I'll submit it for the team's review as soon as possible. Thanks again.

muzimu217:
Hi @hs-liuxh, gentle follow-up on this one — it's the UNESCAPE half of the UTL_URL package (ESCAPE landed via #1728), still merging cleanly with all checks green. If the team review queue allows, we'd love to see it picked up in an upcoming batch. Thanks again!

PR #1753, opened 2026-08-27
Fix volatility of session-dependent functions

### LINKED ISSUE #1754: Session-dependent sys.uid and sys.tz_offset functions are marked immutable
## Bug description

`sys.uid()` and `sys.tz_offset(text)` are declared `IMMUTABLE`, although their results depend on session identity or time-zone state:

- `sys.uid()` reads the current effective user.
- `sys.tz_offset(text)` derives an offset from time-zone rules and the current execution time.

Marking these functions immutable can let PostgreSQL treat their results as independent of session and execution state.

## Expected behavior

The catalog volatility metadata should reflect the functions' session/time dependency. For `sys.tz_offset(text)`, the implementation should also use a statement-stable timestamp so repeated evaluations remain consistent with a `STABLE` declaration.

## Proposed fix

Declare both functions `STABLE`, make `ora_tz_offset` use the statement start timestamp, and add regression coverage. A proposed implementation is available in #1753.

yyqdbngt:
## Summary

- mark `sys.uid()` as `STABLE` because it reads the current effective user
- mark `sys.tz_offset(text)` as `STABLE` because its result depends on the statement execution date and time-zone rules
- use the statement start timestamp so `sys.tz_offset(text)` remains stable throughout one statement
- add regression coverage for the functions' `pg_proc.provolatile` metadata

## Testing

- follow-up: `git diff --check`
- follow-up: verified `GetCurrentStatementStartTimestamp()` is declared by the existing `access/xact.h` include
- initial revision: built `src/backend/postgres`, `contrib/ivorysql_ora/ivorysql_ora.dylib`, `initdb`, `pg_ctl`, `psql`, and `pg_regress` with Meson/Ninja
- initial revision: verified the generated `ivorysql_ora--1.0.sql` declares both functions as `STABLE`

The follow-up C change was not rebuilt in the current Windows workspace because its Perl/build harness is unavailable; CI should compile and run the regression suite. The initial full default-target build reached an existing macOS `libpq_check.pl` failure for the `_exit` symbol, while the affected extension target itself built successfully.

## Upgrade note

`ivorysql_ora` currently publishes only extension version `1.0` and has no versioned update path. The SQL volatility correction therefore applies to fresh extension installations; existing installations retain their current catalog metadata until the project introduces an extension upgrade script or the functions are recreated.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the volatility classification of `sys.tz_offset(text)` and `sys.uid()` from immutable to stable.
  * Time zone offsets are now calculated using the start time of the current statement for consistent results.

* **Tests**
  * Added regression coverage verifying the functions’ stable classifications and expected catalog metadata.

Fixes #1754

### DIFF
--- a/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
+++ b/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
@@ -94,3 +94,15 @@ drop function s1.f_alter(arg1 number, arg2 number);
 drop function s1.f_alter(arg1 OUT int);
 drop function s1.f_alter(arg1 text);
 drop function s1.f_alter(arg1 number, arg2 number, arg3 number);
+-- Functions that inspect session or wall-clock state must not be immutable.
+select proname, provolatile
+from pg_catalog.pg_proc
+where oid in ('sys.tz_offset(text)'::regprocedure,
+              'sys.uid()'::regprocedure)
+order by proname;
+  proname  | provolatile 
+-----------+-------------
+ tz_offset | s
+ uid       | s
+(2 rows)
+
--- a/contrib/ivorysql_ora/sql/datatype_and_func_bugs.sql
+++ b/contrib/ivorysql_ora/sql/datatype_and_func_bugs.sql
@@ -81,3 +81,10 @@ drop function s1.f_alter(arg1 number, arg2 number);
 drop function s1.f_alter(arg1 OUT int);
 drop function s1.f_alter(arg1 text);
 drop function s1.f_alter(arg1 number, arg2 number, arg3 number);
+
+-- Functions that inspect session or wall-clock state must not be immutable.
+select proname, provolatile
+from pg_catalog.pg_proc
+where oid in ('sys.tz_offset(text)'::regprocedure,
+              'sys.uid()'::regprocedure)
+order by proname;
--- a/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
+++ b/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
@@ -677,7 +677,7 @@ AS 'MODULE_PATHNAME', 'ora_tz_offset'
 LANGUAGE C
 STRICT
 PARALLEL SAFE
-IMMUTABLE;
+STABLE;
 
 CREATE FUNCTION sys.months_between(sys.oradate, sys.oradate)
 RETURNS double precision
@@ -1132,7 +1132,7 @@ AS 'MODULE_PATHNAME','uid'
 LANGUAGE C
 STRICT
 PARALLEL SAFE
-IMMUTABLE;
+STABLE;
 
 /*
  * Oracle USERENV support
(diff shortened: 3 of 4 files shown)


yyqdbngt:
Addressed the review follow-up in `049fe57181`:

- `ora_tz_offset` now uses `GetCurrentStatementStartTimestamp()`, so repeated evaluations remain consistent with its `STABLE` declaration within a statement.
- The catalog regression continues to verify the `STABLE` metadata. A deterministic SQL regression that crosses an actual DST boundary would require controllable statement timestamps, which the current regression harness does not expose, so I did not add a wall-clock-dependent test.
- `ivorysql_ora` currently has only the `1.0` install script and no versioned update chain. I documented that existing installations retain the old metadata rather than introducing a partial `1.1` path without a complete fresh-install script.

The follow-up passed `git diff --check`. The current Windows workspace lacks Perl and the project build harness, so compilation/regression execution is left to the maintainer-approved CI runs.

hs-liuxh:
Thanks for your contribution.

hs-liuxh:
Please resolve the regression error.

muzimu217:
The CI regression failure here is a one-line diff — I reproduced it locally on a
real build (aarch64 Linux, master + this patch) and it's the trailing newline,
not the code change.

`datatype_and_func_bugs` fails with exactly one extra blank line at EOF:

```
--- expected/datatype_and_func_bugs.out
+++ results/datatype_and_func_bugs.out
@@ -105,3 +105,4 @@
 tz_offset | s
 uid       | s
 (2 rows)
+
```

psql emits a blank line after the last query block when the file ends with
result rows, and the new `.out` chunk stops at `(2 rows)\n`. Several existing
expected files in this tree end the same way the results do (ora_ascii.out,
ora_datetime.out, ora_binary_float.out, ...), so appending one newline to
`expected/datatype_and_func_bugs.out` matches both the harness and the existing
convention. With that single byte added, `make oracle-check` passes 28/28 on my
build; the `tz_offset`/`uid` rows themselves (`tz_offset | s`, `uid | s`) are
correct — the catalog change and the `GetCurrentStatementStartTimestamp()` switch
both behave as described.

Since the author mentioned not having a local build: as extra data, the
volatility flip also changes index-expression admission, which the current test
doesn't cover. On master (before), `CREATE INDEX ON t(sys.uid())` succeeds; with
this patch it's refused:

```
```

That's the same treatment `lengthc(d)` already gets in ora_character_datatype_functions,
so it's consistent with the tree — just noting it in case a regression row for
it is wanted.

yyqdbngt:
## Root cause of the `oracle_regression` failure

`expected/datatype_and_func_bugs.out` was hand-edited, and psql emits a trailing blank line after the result table that the hand-written file was missing:

```
@@ -105,3 +105,4 @@
  tz_offset | s
  uid       | s
 (2 rows)
+
```

So the test failed on a **correct** result — the volatility change itself matches what the catalog reports (`tz_offset | s`, `uid | s`).

## Fix

`expected/datatype_and_func_bugs.out` is regenerated from a real run instead of being edited by hand, which is the only reliable way to keep the echoed output byte-exact.

## Validation

Locally, Ubuntu 24.04 / gcc 13, `./configure --with-uuid=e2fs`:

- before regeneration: `not ok 1 - datatype_and_func_bugs`, diff is the single missing blank line;
- after regeneration: `ok 1 - datatype_and_func_bugs`, no `regression.diffs`.

This is the same failure mode as #1806, so both are fixed the same way.

Issue #1621, opened 2026-08-14
Oracle-compat behavior fixes: DBMS_LOCK.RELEASE code, add_months overflow,  parameter truncation


Aias00:
### IvorySQL Version

`master` at `c3529ba4789ee220caefc858e58bbe6f1563ad14`

### Current Behavior

Three low-severity behavior defects:

1. **DBMS_LOCK.RELEASE returns the wrong code** (`contrib/ivorysql_ora/src/builtin_packages/dbms_lock/dbms_lock.c:336`): when the session does not hold the lock, the function returns `DBMS_LOCK_PARAM_ERROR` (3). Oracle returns 4 (`ALREADY_OWNED`); the constant `DBMS_LOCK_ALREADY_OWNED` is defined but never used.

2. **add_months truncates the month argument** (`contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c:405,418`): `numeric_int8` is narrowed to `int32` via `DatumGetInt32`, and `y * 12 + m - 1 + n` is computed in `int` arithmetic. For `|months| > 2^31` the result wraps and is wrong.

3. **`$N` parameter truncation in the Oracle scanner** (`src/backend/parser/ora_scan.l:1163`): `yylval->ival = atol(yytext + 1)` narrows `long` to `int` without range checking; `$4294967297` silently becomes `$1` (and `atol` overflow is UB). PostgreSQL's native scanner uses `pg_strtoint32_safe` and reports "parameter number too large".

### Expected behavior/code

- RELEASE returns `DBMS_LOCK_ALREADY_OWNED` when the lock is not held
- add_months uses 64-bit arithmetic for the month offset
- `$N` uses checked conversion (`pg_strtoint32_safe`) and errors on overflow

NotHimmel:
Thanks for the report and PR, could you confirm how it was found and provide a concrete reproduction?

Aias00:
Hi @NotHimmel,

Thanks for the questions. Here is how the defects were found and concrete reproductions for all three items (all verified against master in Oracle-compatible mode).

**How they were found**

Static code review of the Oracle-compatibility layer (dbms_lock package, datetime functions, Oracle scanner).

**Reproduction**

*Item 1 - DBMS_LOCK.RELEASE return code*

```sql
declare
 v_handle text;
 v_rc int;
begin
 dbms_lock.allocate_unique('demo1621', v_handle);
[...]

PR #1762, opened 2026-08-28
Fix add_months for BC dates (negative month totals) and int32 month truncation

### LINKED ISSUE #1761: add_months returns wrong dates / errors for BC results (negative month totals)
### IvorySQL Version

### Current Behavior

`sys.add_months` produces wrong results (or errors) whenever the total month count goes negative, i.e. whenever the result crosses into the BC era. Oracle supports `DATE` values from 4712 BC through 9999 AD, so BC results are ordinary valid dates.

Reproductions on master:

```sql
-- (1) result crosses into BC: errors out instead of returning a valid date
SELECT add_months(to_date('0002-06-15 BC','yyyy-mm-dd ad'), 3);
ERROR:  date out of range

-- (2) result stays in BC but the year is shifted by one
SELECT to_char(add_months(to_date('4712-01-01 BC','yyyy-mm-dd ad'), 12), 'syyyy-mm-dd');
 -4710-01-01        -- expected -4711-01-01 (4712 BC Jan + 12 months = 4711 BC Jan)

-- (3) AD input with a large negative month count
SELECT to_char(add_months('0001-01-01', -25), 'syyyy-mm-dd');
ERROR:  date out of range
```

### Expected behavior

Oracle returns a valid date for all of the above (e.g. case (1) is `0002-09-15 BC`, case (2) is `4711-01-01 BC`, case (3) is `0003-12-01 BC` in the no-year-zero AD/BC display).

### Root cause

In `contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c`, `add_months()` splits the total month count with `div()`, which truncates toward zero, and then tries to compensate with `if (y < 0) y += 1`:

```c
v = div(y * 12 + m - 1 + n, 12);
y = v.quot;
if (y < 0)
    y += 1;
m = v.rem + 1;
```

For a negative total the remainder is negative (or zero), so `m = v.rem + 1` lands on `0` or a negative month, and the `y += 1` hack shifts the year by a full 12 months. `date2j()` is then called with an out-of-range month, producing a bogus timestamp ("date out of range") or a silently wrong date.

Additionally the month argument is narrowed to int32 (`DatumGetInt32(DirectFunctionCall1(numeric_int8, ...))`), so `|months| >= 2^31` silently wraps (same class of defect reported in #1621).

### Suggested fix

Keep the month argument as `int64` and normalize the `lldiv()` result so the month is always in 1..12 (borrow one year whenever the remainder is non-positive). I have a patch with regression tests ready and will submit a PR right after filing this issue.

unbridled-41:
## Issue

Fixes #1761

## Summary

`add_months()` produced wrong results (or `ERROR: date out of range`) whenever the total month count went negative, i.e. whenever the result crossed into the BC era:

| Call (master) | Before | After |
|---|---|---|
| `add_months(to_date('0002-06-15 BC','yyyy-mm-dd ad'), 3)` | `ERROR: date out of range` | `-0002-09-15` |
| `to_char(add_months(to_date('4712-01-01 BC','yyyy-mm-dd ad'), 12), 'syyyy-mm-dd')` | `-4710-01-01` (a year off) | `-4711-01-01` |
| `to_char(add_months('0001-01-01', -25), 'syyyy-mm-dd')` | `ERROR: date out of range` | `-0003-12-01` |

## Root cause and fix

`div()` truncates toward zero, so for a negative total the remainder is non-positive; the old code compensated with `if (y < 0) y += 1`, which leaves `m = v.rem + 1` at `0` or negative and shifts the year by a full 12 months. The fix normalizes the `lldiv()` result so the month always lands in 1..12:

```c
v = lldiv((int64) y * 12 + m - 1 + n, 12);
y = (int) v.quot;
m = (int) v.rem + 1;
if (m <= 0)
{
    y -= 1;
    m += 12;
}
```

This also widens the month argument to `int64` (`DatumGetInt64`), so large `NUMBER` inputs no longer silently truncate through `DatumGetInt32` (the `add_months` item of #1621).

## Testing

Five regression cases added to `datatype_and_func_bugs` (BC forward step, the 4712 BC lower bound, last-day rule across a BC leap February, an AD-to-BC crossing, and a 6-digit negative month count). Full `make oracle-check` for `ivorysql_ora`: **all 26 tests pass**, including the untouched `ora_datetime*` suites.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `add_months` handling for dates crossing between AD and BC years.
  - Corrected calculations for BC dates, including leap days, the earliest supported BC year, and large month offsets.
  - Prevented invalid month values when date arithmetic crosses year boundaries.

- **Tests**
  - Added regression coverage for BC-era date arithmetic, year transitions, leap-year behavior, and extreme date ranges.

### DIFF
--- a/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
+++ b/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
@@ -29,6 +29,37 @@ select add_months('2022-08-23',3) from dual;
  2022-11-23 00:00:00
 (1 row)
 
+-- add_months across BC years; SYYYY makes the era visible
+select to_char(add_months(to_date('0002-06-15 BC', 'yyyy-mm-dd ad'), 3), 'syyyy-mm-dd') as add_months_bc;
+ add_months_bc 
+---------------
+ -0002-09-15
+(1 row)
+
+select to_char(add_months(to_date('4712-01-01 BC', 'yyyy-mm-dd ad'), 12), 'syyyy-mm-dd') as add_months_bc_min;
+ add_months_bc_min 
+-------------------
+ -4711-01-01
+(1 row)
+
+select to_char(add_months(to_date('0005-01-31 BC', 'yyyy-mm-dd ad'), 1), 'syyyy-mm-dd') as add_months_bc_leap;
+ add_months_bc_leap 
+--------------------
+ -0005-02-29
+(1 row)
+
+select to_char(add_months('0001-01-01', -25), 'syyyy-mm-dd') as add_months_ad_to_bc;
+ add_months_ad_to_bc 
+---------------------
+ -0003-12-01
+(1 row)
+
+select to_char(add_months('9999-12-31', -120002), 'syyyy-mm-dd') as add_months_back_to_bc;
+ add_months_back_to_bc 
+-----------------------
+ -0002-10-31
+(1 row)
+
 alter session set NLS_LENGTH_SEMANTICS='BYTE';
 create table char_tb(char_clo char(3));
 insert into char_tb values('测试');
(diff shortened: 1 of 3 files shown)


hanjianqiao:

PR #1764, opened 2026-08-28
Fix sys.instr edge cases: position=0, occurrence<=0, and INT32_MIN position crash

### LINKED ISSUE #1763: sys.instr: position=0 returns a wrong match, occurrence<=0 unvalidated, and position=INT32_MIN crashes the backend
### IvorySQL Version

### Current Behavior

Three related defects in the `sys.instr` / `sys.instrb` family, all reachable from plain SELECTs:

**1. `position = 0` returns a wrong match instead of 0.** Oracle documents that INSTR returns 0 when position is 0. `instrb()` special-cases this in its wrapper, but the character variant falls into the negative-position code path and returns the position of a wrong-end match:

```sql
SELECT instr('abcabc', 'abc', 0);
 instr
-------
     4          -- Oracle returns 0
```

**2. `occurrence <= 0` silently returns a bogus value instead of raising ORA-01428** (occurrence must be a positive integer). The search loop simply never runs, so `result` keeps its initial value:

```sql
SELECT instr('abcabc', 'b', 3, 0);
 instr
-------
     2          -- Oracle: ORA-01428
```

**3. `position = -2147483648` crashes the backend.** `abs(INT_MIN)` overflows (UB), the `abs(position) > src_text_len` guard passes, and `text_position_prev()` is entered with a non-positive start position, tripping `Assert(start_pos > 0)` under `--enable-cassert` (release builds execute the UB pointer arithmetic):

```sql
SELECT instr('abc', 'a', -2147483648);
```

### Expected behavior

1. `position = 0` returns 0, matching `instrb()` and Oracle.
2. `occurrence <= 0` raises an out-of-range parameter error (ORA-01428).
3. Extreme positions must be rejected safely instead of crashing.

### Suggested fix

Move both validations into `text_instring()` (single place, so `instr` and `instrb` agree), replace the `abs(position)` guard with a `position > src_text_len || position < -src_text_len` check performed *before* `position` is negated for the `maxTimes` computation, and fix `instrb()`'s occurrence error code from `ERRCODE_PROGRAM_LIMIT_EXCEEDED` to `ERRCODE_INVALID_PARAMETER_VALUE`. A patch with regression tests is ready and will follow as a PR.

unbridled-41:
## Issue

Fixes #1763

## Summary

Hardens the `sys.instr` / `sys.instrb` family against three parameter edge cases, one of which crashes the backend under `--enable-cassert`:

| Call | Before | After |
|---|---|---|
| `SELECT instr('abcabc', 'abc', 0)` | `4` (wrong match) | `0` (Oracle-documented) |
| `SELECT instr('abcabc', 'b', 3, 0)` | `2` (bogus) | `ERROR: The parameter of occurrence must be positive` |
| `SELECT instr('abc', 'a', -2147483648)` | backend crash (Assert/UB) | `0` |
| `SELECT instrb('CORPORATE FLOOR', 'OR', 5, 0)` | wrong SQLSTATE (XX540) | same error, `ERRCODE_INVALID_PARAMETER_VALUE` (22023) |

## Details

1. **`position = 0`**: handled up front in `text_instring()` so `instr` and `instrb` agree (previously only `instrb()`'s wrapper special-cased it, while the character variant fell into the negative-position path).
2. **`occurrence <= 0`**: validated in `text_instring()` (Oracle ORA-01428), replacing `instrb()`'s wrapper check that used `ERRCODE_PROGRAM_LIMIT_EXCEEDED`.
3. **`position = INT32_MIN`**: `abs(INT_MIN)` overflows and the old guard never fired; the backend then hit `Assert(start_pos > 0)` in `text_position_prev()` under cassert builds. The range check now compares `position` against `+-src_text_len` *before* `position` is negated for the `maxTimes` computation, so no negation of `INT_MIN` ever happens.

## Testing

Nine regression cases added to `ora_character_datatype_functions` covering `position = 0` for both variants, `occurrence = 0` / negative, the `INT32_MIN` position (used to crash cassert builds), and the fixed SQLSTATE. Full `make oracle-check` for `ivorysql_ora`: **all 26 tests pass**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `INSTR` and `INSTRB` handling for zero, negative, and out-of-range positions.
  * Added consistent validation for zero or negative occurrence values, returning the documented error.
  * Prevented crashes when searches use the minimum supported 32-bit position.
  * Standardized behavior across character- and byte-based searches.

* **Tests**
  * Added regression coverage for position and occurrence edge cases in both functions.

## Review follow-up (commit 2)

Per the automated review, two additional INSTRB cases were added: `occurrence = -1` (same positivity error) and `position = -2147483648` (range-rejected without crashing).

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -3370,6 +3370,37 @@ SELECT INSTR('Information for the information age requires information.', 'infor
      0
 (1 row)
 
+-- position = 0 returns 0 (Oracle documents this); position/occurrence edge cases
+SELECT INSTR('abcabc', 'abc', 0) FROM DUAL;
+ instr 
+-------
+     0
+(1 row)
+
+SELECT INSTR('abcabc', 'b', 0, 2) FROM DUAL;
+ instr 
+-------
+     0
+(1 row)
+
+SELECT INSTR('abcabc', 'b', 3, 0) FROM DUAL;	-- error: occurrence must be positive
+ERROR:  The parameter of occurrence must be positive
+SELECT INSTR('abcabc', 'b', 3, -1) FROM DUAL;	-- error: occurrence must be positive
+ERROR:  The parameter of occurrence must be positive
+SELECT INSTR('abc', 'a', -2147483648) FROM DUAL;	-- out-of-range position, must not crash
+ instr 
+-------
+     0
+(1 row)
+
+SELECT INSTRB('CORPORATE FLOOR', 'OR', 0, 2) FROM DUAL;
+ instrb 
+--------
+      0
+(1 row)
+
+SELECT INSTRB('CORPORATE FLOOR', 'OR', 5, 0) FROM DUAL;	-- error: occurrence must be positive
+ERROR:  The parameter of occurrence must be positive
 /*
  * asciistr
  */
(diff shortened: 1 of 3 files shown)


hanjianqiao:

PR #1766, opened 2026-08-28
Apply Oracle match-parameter semantics to regexp_like and regexp_count

### LINKED ISSUE #1765: regexp_like: 'm' rejected, 'n'/default newline semantics inverted (PG flags instead of Oracle); regexp_count validates only first flag character
### IvorySQL Version

### Current Behavior

`sys.regexp_like` hand-rolls its flag parsing instead of using the shared Oracle flag parser (`ora_parse_re_flags`) that the regex engine was patched to implement, so it diverges from Oracle in three ways:

**1. `'m'` is rejected although it is a documented Oracle match parameter** (treat the source string as multiple lines):

```sql
SELECT regexp_like('ab' || chr(10) || 'cd', '^cd', 'm') FROM dual;
ERROR:  invalid option of regexp_like: m     -- Oracle: matches, returns 1
```

**2. `'n'` enables PostgreSQL's REG_NEWLINE, the opposite of Oracle's meaning.** Oracle's `'n'` lets `.` match a newline; PostgreSQL's REG_NEWLINE makes `.` NOT match newlines and turns `^`/`$` into line anchors. In Oracle `.` never matches a newline unless `'n'` is given, so the default is inverted too:

```sql
SELECT regexp_like('ab' || chr(10) || 'cd', 'b.cd') FROM dual;      -- master: true, Oracle: false
SELECT regexp_like('ab' || chr(10) || 'cd', 'b.cd', 'n') FROM dual; -- master: false, Oracle: true
```

**3. `regexp_count` validates only the first character of its match parameter**, so PostgreSQL-only flags leak through and are silently applied:

```sql
SELECT regexp_count('ABCabcabcABCABC', 'abc', 1, 'is') FROM dual;  -- 's' silently accepted
SELECT regexp_count('ABCabcabcABCABC', 'abc', 1, 'ie') FROM dual;  -- 'e' silently accepted
```

Oracle raises ORA-01428 for any match parameter outside i/c/n/m/x.

### Expected behavior

* `'m'` accepted with Oracle line-anchor semantics;
* `'.'` does not match a newline by default, `'n'` enables it;
* multi-line `^`/`$` anchoring via `'m'`;
* non-Oracle match parameters rejected with ORA-01428-style errors in all regexp functions.

### Suggested fix

Validate every flag character against the Oracle set first (keeping regexp_like's own error message), then delegate to `ora_parse_re_flags()`; give the 2-argument regexp_like the same default (`REG_ADVANCED | REG_NLDOT`); and validate every character of regexp_count's match parameter. A patch with regression tests is ready and will follow as a PR.

unbridled-41:
## Issue

Fixes #1765

## Summary

Brings `sys.regexp_like` (and `sys.regexp_count`'s parameter validation) in line with the Oracle regex semantics the engine was already patched for:

| Call | Before | After (Oracle) |
|---|---|---|
| `regexp_like('ab\n', '^cd', 'm')` | `ERROR: invalid option of regexp_like: m` | `t` |
| `regexp_like('ab\ncd', 'b.cd')` | `t` (`.` matched `\n`) | `f` |
| `regexp_like('ab\ncd', 'b.cd', 'n')` | `f` | `t` |
| `regexp_like(..., 'q')` | `t` (PG REG_QUOTE silently applied) | `ERROR: invalid option of regexp_like: q` |
| `regexp_count(..., 'is')` | `t`-style silent PG flags | `ERROR: the 4th argument is illegal parameter...` |

## Details

* `ora_regexp_like()` validated flags with its own switch (i/n/c/x) and built PostgreSQL compile flags directly. It now (1) rejects any character outside Oracle's set `{i,c,n,m,x}` up front - keeping the historical `invalid option of regexp_like: %c` message - and (2) delegates to `ora_parse_re_flags()`, which implements the Oracle newline semantics (default `REG_NLDOT`, `'n'` lets `.` match `\n`, `'m'` makes `^`/`$` line anchors).
* `ora_regexp_like_no_flags()` now uses `REG_ADVANCED | REG_NLDOT`, so the default flavor matches Oracle (`.` does not match a newline).
* `ora_regexp_count()` checked only `paramstr[0]`, letting PostgreSQL-only letters (`s`, `p`, `w`, `e`, `q`, ...) flow into the shared parser silently; every character is now validated against the same set, raising the existing `illegal parameter` error for the whole string.
* Two pre-existing expected results that depended on the leak (`'cd'` falling through to the PG regex parser's message, `'ie'` silently using PG's `'e'` flag) now report the regexp_count validation error, matching Oracle's ORA-01428.

## Testing

Regression cases added for `.`-vs-newline with and without `'n'`, `'m'` anchors on multi-line strings, rejection of `'p'`/`'q'`, and multi-character match parameters for `regexp_count`. Full `make oracle-check` for `ivorysql_ora`: **all 26 tests pass**.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Oracle-compatible regular expression behavior for newline matching and multiline anchors.
  * Match parameters are now fully validated, with consistent errors for unsupported options.
  * Correctly rejects PostgreSQL-only and other invalid options in regular expression functions.
  * `REGEXP_COUNT` now handles valid and invalid parameter combinations consistently.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -1574,6 +1574,43 @@ select * from t_regexp_like where regexp_like(null, null, null);
 ----+-------
 (0 rows)
 
+-- Oracle newline semantics: '.' does not match a newline unless 'n' is given
+select regexp_like(('ab'||chr(10)||'cd')::varchar2, 'b.cd'::varchar2) from dual;
+ regexp_like 
+-------------
+ f
+(1 row)
+
+select regexp_like(('ab'||chr(10)||'cd')::varchar2, 'b.cd'::varchar2, 'n'::varchar2) from dual;
+ regexp_like 
+-------------
+ t
+(1 row)
+
+-- 'm' is a valid Oracle match parameter: ^/$ also match at newlines
+select regexp_like(('ab'||chr(10)||'cd')::varchar2, '^cd'::varchar2, 'm'::varchar2) from dual;
+ regexp_like 
+-------------
+ t
+(1 row)
+
+select regexp_like(('ab'||chr(10)||'cd')::varchar2, '^cd'::varchar2) from dual;
+ regexp_like 
+-------------
+ f
+(1 row)
+
+select regexp_like(('ab'||chr(10)||'cd')::varchar2, 'ab$'::varchar2, 'm'::varchar2) from dual;
+ regexp_like 
+-------------
+ t
+(1 row)
+
+-- PostgreSQL-only match parameters are still rejected
+select regexp_like('abc'::varchar2, 'abc'::varchar2, 'p'::varchar2) from dual;
+ERROR:  invalid option of regexp_like: p
+select regexp_like('abc'::varchar2, 'abc'::varchar2, 'q'::varchar2) from dual;
+ERROR:  invalid option of regexp_like: q
 DROP table t_regexp_like;
 /*
  * regexp_count
@@ -1752,7 +1789,7 @@ SELECT REGEXP_COUNT(1231, 123::int2, 1::int2, 'c'::char) REGEXP_COUNT FROM DUAL;
 (1 row)
 
 SELECT REGEXP_COUNT(123123::int4, 2312::int4, 1::int4, 'cd') REGEXP_COUNT FROM DUAL;
-ERROR:  invalid regular expression option: "d"
+ERROR:  the 4th argument is illegal parameter for function. the parameter can be one of x, m, i, c, n.
 SELECT REGEXP_COUNT(12312312312, 231, 1, 'c') REGEXP_COUNT FROM DUAL;
  regexp_count 
 --------------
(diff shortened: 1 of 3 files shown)


unbridled-41:
Thanks for the review! Two notes on the findings here, one skipped and one applied:

1. **Rejecting `g`** - skipped for now, because this function's SQL declaration gives the 4th argument a `'g'` default (`regexp_count(varchar2, varchar2, number default 1, varchar2 default 'g')`), an IvorySQL design choice that predates this PR. Removing `g` from the allowlist without also changing that default would make every two-argument call fail. Aligning both together (default `NULL` + strict Oracle parameter set) is a deliberate behavior change I'd rather let maintainers opt into, since it affects existing callers.

2. The multi-character validation itself (the actual subject of this PR) is applied, and the `'gg'` case you pointed at cannot occur any more once `g` handling is decided - I've kept `g` accepted exactly as the default requires.

Happy to follow up with the stricter variant as a separate PR if maintainers want Oracle-strict semantics here.

hanjianqiao:

PR #1768, opened 2026-08-28
Reject overlength time zone abbreviations in to_char (CVE-2026-14669)

### LINKED ISSUE #1767: IvorySQL master is missing the 2026-05-14 and 2026-08-13 PostgreSQL security releases (39 CVEs), incl. reproducible to_char heap overflow (CVE-2026-14669)
### IvorySQL Version

### Summary

IvorySQL master carries a snapshot of PostgreSQL master as of **2026-04-16** (see open sync PR #1699, "Sync IvorySQL with PG master: 2026-04-16 to 2026-07-14"). The upstream PostgreSQL security releases of **2026-05-14 (18.4/17.10/..., 11 CVEs)** and **2026-08-13 (18.6/17.11/..., 28 CVEs)** therefore include fixes that are **not present in IvorySQL master**.

I verified the gap concretely for two of the highest-severity items:

**1. CVE-2026-14669 (CVSS 8.8): heap buffer overflow in `to_char()` via overlength time zone abbreviations - present and reproducible in IvorySQL.**

Upstream fix: commit `3d724bf4fde67a2931733a5143b7d6c12b23990c` ("Guard against overlength time zone abbreviations in to_char()"), which bounds the `strcpy()` of the abbreviation in `DCH_to_char()`'s `DCH_tz`/`DCH_TZ` cases.

IvorySQL master still contains the unbounded `strcpy(s, tmtcTzn(in))`, and it is directly exploitable by any session (any user can override its own `timezone`):

```sql
SET timezone = '<AAAA...80 A's...>-07:30';   -- 80-char POSIX abbreviation, accepted by SET
SELECT to_char(now(), 'TZ');
```

With `--enable-cassert`, the very first call reports allocator-detected corruption:

```
WARNING:  detected write past chunk end in ExprContext 0x...
```

On non-assert builds this is silent heap corruption.

**2. CVE-2026-6473 (CVSS 8.8): regex engine integer-overflow hardening - missing.**

Upstream commit "Harden our regex engine against integer overflow in size calculations" (2026-05-14) introduced `MALLOC_ARRAY`/`REALLOC_ARRAY` across `src/backend/regex/`. `grep -rn "MALLOC_ARRAY\|REALLOC_ARRAY" src/backend/regex/` in IvorySQL master returns **0 hits**, i.e. the pre-hardening engine. This matters beyond core because IvorySQL's Oracle compatibility layer (regexp_like/regexp_replace/regexp_instr/regexp_substr/regexp_count) drives the same engine.

### Suggested remediation

1. Merge (or fast-forward) PR #1699 as soon as possible, since it closes the bulk of the window;
2. until then, cherry-pick the upstream security commits for the CVEs above (and audit the remainder of the May+August 2026 batches).

I have a patch for item 1 (the `to_char` guard, ported from upstream with regression tests) ready and will submit it as a PR immediately after filing this issue.

unbridled-41:
## Issue

Fixes #1767

## Summary

Ports the upstream fix for **CVE-2026-14669** (commit `3d724bf4fde67a2931733a5143b7d6c12b23990c`, PostgreSQL 18.6) to IvorySQL: `DCH_to_char()` now rejects time zone abbreviations longer than the reserved `12 * fmt_len` output space instead of `strcpy()`-ing past the end of the palloc'd buffer.

Any session can reach this with its own `timezone` setting:

```sql
SET timezone = '<AAA...80 chars...>-07:30';
SELECT to_char(now(), 'TZ');
```

Before (assert build): `WARNING: detected write past chunk end in ExprContext ...` (silent heap corruption on production builds).
After: `ERROR:  time zone format value too long`.

## Details

The change is a verbatim port of the upstream guard for both `DCH_tz` and `DCH_TZ` spellings: `strlen(p) <= n->key->len * DCH_MAX_ITEM_SIZ` or `ereport(ERROR, (ERRCODE_DATETIME_VALUE_OUT_OF_RANGE, errmsg("time zone format value too long")))`. IvorySQL's `formatting.c` carries Oracle-compatible modifications in other areas, but this hunk applies cleanly and unchanged.

## Testing

Regression cases added to `datatype_and_func_bugs` (an 80-character POSIX abbreviation with both `'TZ'` and `'tz'`). Full `make oracle-check` for `ivorysql_ora`: **all 26 tests pass**. The PoC that previously tripped the allocator's chunk-end check now produces the clean error and the server stays healthy.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved time zone formatting to safely handle overlong abbreviations and names.
  - Displays a clear error instead of allowing oversized values to overflow output.
  - Preserves expected uppercase and lowercase formatting behavior.

- **Tests**
  - Added regression coverage for overlength time zone abbreviations using both uppercase and lowercase formats.

### DIFF
--- a/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
+++ b/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
@@ -84,6 +84,14 @@ ERROR:  syntax error at or near "default"
 LINE 1: ... s1.f_alter(arg1 number, arg2 number, arg3 number default 10...
                                                              ^
 alter function s1.f_alter(arg1 number, arg2 number, arg3 number) compile;
+-- to_char must reject overlength time zone abbreviations instead of
+-- overflowing its output buffer (upstream CVE-2026-14669)
+set timezone = '<AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA>-07:30';
+select to_char(cast(now() as sys.oratimestamptz), 'TZ') from dual;
+ERROR:  time zone format value too long
+select to_char(cast(now() as sys.oratimestamptz), 'tz') from dual;
+ERROR:  time zone format value too long
+reset timezone;
 -- clean
 drop table dtos_tb1tidid004134318;
 drop table char_tb2;
--- a/contrib/ivorysql_ora/sql/datatype_and_func_bugs.sql
+++ b/contrib/ivorysql_ora/sql/datatype_and_func_bugs.sql
@@ -71,6 +71,14 @@ alter function s1.f_alter(arg1 OUT int) compile;
 alter function s1.f_alter(arg1 text) compile;
 alter function s1.f_alter(arg1 number, arg2 number, arg3 number default 10) compile;
 alter function s1.f_alter(arg1 number, arg2 number, arg3 number) compile;
+
+-- to_char must reject overlength time zone abbreviations instead of
+-- overflowing its output buffer (upstream CVE-2026-14669)
+set timezone = '<AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA>-07:30';
+select to_char(cast(now() as sys.oratimestamptz), 'TZ') from dual;
+select to_char(cast(now() as sys.oratimestamptz), 'tz') from dual;
+reset timezone;
+
 -- clean
 drop table dtos_tb1tidid004134318;
 drop table char_tb2;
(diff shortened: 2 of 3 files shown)


unbridled-41:
Thanks for the review! On the proposed `SET`/`RESET` command-tag rows: I double-checked the recorded results, and this regression harness does not echo command tags for `SET`/`RESET` statements (the neighboring pre-existing tests in this file, e.g. `alter session set NLS_DATE_FORMAT ...`, also record no tag rows). The expected file was generated from an actual run of the suite, so adding the tags would make it fail. Skipping this one, but happy to reconsider if a harness change starts emitting tags.

hanjianqiao:
Thanks for your work! However, we can't accept this PR, as the change comes from an upstream PostgreSQL commit. IvorySQL merges upstream PostgreSQL commits one by one on our own schedule, so there's no need to port them individually. This issue will be resolved automatically once IvorySQL catches up to the corresponding PG commit.

That said, contributions for IvorySQL-specific issues are always very welcome!

PR #1770, opened 2026-08-28
Reject SQL calls to functions taking or returning type internal (CVE-2026-14680)

### LINKED ISSUE #1769: CVE-2026-14680 (8.8): calls to internal-argument functions are not rejected; also unguarded in IvorySQL's PL/iSQL
### IvorySQL Version

### Summary

IvorySQL master (PostgreSQL master snapshot of 2026-04-16, see #1767) predates the upstream fix for **CVE-2026-14680** (CVSS 8.8, PostgreSQL 18.6, 2026-08-13): *"Type confusion via 'internal' function arguments"*. Upstream commit `8f7e35b08a` ("Reject calls from SQL to functions that take or return type internal") closes it.

`internal` is a placeholder type whose meaning is defined independently by every C function that uses it; nothing validates that a caller passes a compatible layout. If SQL can resolve a call to such a function, the caller gets a type-confusion primitive. The pre-fix rules only blocked the trivial `'foo'::internal` case.

Verified against IvorySQL master:

* `src/backend/parser/parse_func.c` / `parse_oper.c` contain no `INTERNALOID` call rejection (the upstream hunks are absent);
* additionally, IvorySQL's own **PL/iSQL** (`src/pl/plisql/src/pl_exec.c`, `get_cast_hashentry()`) has the same unguarded I/O-coercion fallback that upstream fixed in plpgsql's `pl_exec.c` - a code path upstream's fix covered but IvorySQL's fork did not inherit.

### Suggested remediation

Port upstream commit `8f7e35b08a` (4 files: parse_coerce.c, parse_func.c, parse_oper.c, plpgsql pl_exec.c) and apply the equivalent guard to `src/pl/plisql/src/pl_exec.c`. A patch doing exactly this, with the full oracle regression suite passing (26/26), will follow as a PR.

unbridled-41:
## Issue

Fixes #1769 (part of the upstream security-sync gap tracked in #1767)

## Summary

Ports upstream commit `8f7e35b08a` (PostgreSQL 18.6, 2026-08-13, CVE-2026-14680, CVSS 8.8): SQL can no longer resolve calls to functions whose declared argument or result types include `internal`, because each such function defines that layout differently and a resolved call is a type-confusion primitive.

| File | Change |
|---|---|
| `parse_coerce.c` | `can_coerce_type()` and `find_coercion_pathway()` reject any cast to/from `internal` |
| `parse_func.c` | `ParseFuncOrColumn()` rejects resolved calls whose declared argument types or return type are `internal` |
| `parse_oper.c` | same check for operator syntax in `make_op()` |
| `src/pl/plpgsql/src/pl_exec.c` | `get_cast_hashentry()` rejects the I/O-coercion fallback for `internal` |
| `src/pl/plisql/src/pl_exec.c` | **IvorySQL-specific:** the identical guard added to their own PL/iSQL fork, which upstream's plpgsql fix does not cover |

The first four hunks are a verbatim port of the upstream commit (it applies cleanly with `git apply`). The fifth is the same guard added to IvorySQL's PL/iSQL copy of the fallback, which the upstream patch could not reach.

## Testing

Full `make oracle-check` for `ivorysql_ora`: **all 26 tests pass** - the Oracle compatibility layer (varchar2/number overloads, packages, PL/iSQL routines) does not rely on SQL-resolvable `internal` calls, matching the fact that upstream applied the same restriction to every supported branch.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented SQL functions and operators from directly accepting or returning the internal system type.
  * Blocked implicit and explicit casts to or from the internal type.
  * Improved error reporting for unsupported internal-type conversions in procedural code.
  * Preserved supported fallback conversions for other special types.
  * Added regression coverage to verify these invalid calls and casts fail cleanly.

## Review follow-up (commit 2): behavior-level tests added

The upstream commit shipped no test, so three deterministic SQL routes are now covered in `datatype_and_func_bugs`:

| Call | Result |
|---|---|
| `SELECT xid8recv(NULL)` (untyped NULL into an `internal`-argument function) | `ERROR: function xid8recv(unknown) does not exist` (resolution no longer reaches the function) |
| `SELECT pg_ddl_command_recv(NULL::internal)` | `ERROR: cannot cast type unknown to internal` |
| `SELECT cast(NULL as internal) IS NULL` | same coercion error |

Full `ivorysql_ora` suite: **26/26** including the new cases.

### DIFF
--- a/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
+++ b/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
@@ -84,6 +84,22 @@ ERROR:  syntax error at or near "default"
 LINE 1: ... s1.f_alter(arg1 number, arg2 number, arg3 number default 10...
                                                              ^
 alter function s1.f_alter(arg1 number, arg2 number, arg3 number) compile;
+-- calls to functions taking type "internal" are rejected from SQL
+-- (upstream CVE-2026-14680); every route must fail cleanly
+select xid8recv(NULL) from dual;
+ERROR:  function xid8recv(unknown) does not exist
+LINE 1: select xid8recv(NULL) from dual;
+               ^
+DETAIL:  No function of that name accepts the given argument types.
+HINT:  You might need to add explicit type casts.
+select pg_ddl_command_recv(NULL::internal) from dual;
+ERROR:  cannot cast type unknown to internal
+LINE 1: select pg_ddl_command_recv(NULL::internal) from dual;
+                                       ^
+select cast(NULL as internal) is null from dual;
+ERROR:  cannot cast type unknown to internal
+LINE 1: select cast(NULL as internal) is null from dual;
+               ^
 -- clean
 drop table dtos_tb1tidid004134318;
 drop table char_tb2;
--- a/contrib/ivorysql_ora/sql/datatype_and_func_bugs.sql
+++ b/contrib/ivorysql_ora/sql/datatype_and_func_bugs.sql
@@ -71,6 +71,12 @@ alter function s1.f_alter(arg1 OUT int) compile;
 alter function s1.f_alter(arg1 text) compile;
 alter function s1.f_alter(arg1 number, arg2 number, arg3 number default 10) compile;
 alter function s1.f_alter(arg1 number, arg2 number, arg3 number) compile;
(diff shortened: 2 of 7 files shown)


hanjianqiao:

PR #1771, opened 2026-08-29
Protect fixed-size arrays sized by FUNC_MAX_ARGS from overflow (CVE-2026-14679)

unbridled-41:
## Issue

Part of the upstream security-sync gap tracked in #1767.

## Summary

Ports upstream commit `60e7329b9b` (PostgreSQL 18.6, 2026-08-13, **CVE-2026-14679**, CVSS 8.2): several server paths index fixed-size stack arrays by the number of function arguments without bounding it to `FUNC_MAX_ARGS`, so invoking a function with more than `FUNC_MAX_ARGS` arguments writes past the end of the array (stack buffer overflow).

The patch applies verbatim to this snapshot and covers:

* `nodeWindowAgg.c` - window functions whose proargtypes exceed `FUNC_MAX_ARGS`;
* `parse_agg.c` - ordered-set aggregate argument counts;
* `parse_func.c` - FuncCall default/declared argument handling;
* `funccache.c` / `funcapi.c` - default-argument expansion;
* `pltcl.c` - Tcl argument vector guard.

## Testing

Three-layer validation, all green:

1. full `ivorysql_ora` oracle regression suite: **26/26**;
2. full core `oracle_test` regression suite: **259/259** (window, aggregates, opr_sanity, type_sanity, polymorphism and every other schedule member).

Upstream applied the same commit to all supported branches with no test changes, consistent with these results.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added safeguards that reject functions and aggregates exceeding supported argument limits.
  * Improved handling of stored views and parse trees created with differing argument-limit settings.
  * Prevented potential memory overruns when processing functions with excessive arguments.
  * Improved PL/Tcl handling for functions with large argument lists.

### DIFF
--- a/src/backend/executor/nodeWindowAgg.c
+++ b/src/backend/executor/nodeWindowAgg.c
@@ -1086,6 +1086,20 @@ eval_windowfunction(WindowAggState *winstate, WindowStatePerFunc perfuncstate,
 	oldContext = MemoryContextSwitchTo(winstate->ss.ps.ps_ExprContext->ecxt_per_tuple_memory);
 
 	/*
+	 * Protect fixed-size fcinfo.  Ordinarily this would have been checked
+	 * while creating the WindowFunc, but it's possible that we are looking at
+	 * a parsetree from a stored view that was made by a server executable
+	 * with a different value of FUNC_MAX_ARGS.
+	 */
+	if (perfuncstate->numArguments > FUNC_MAX_ARGS)
+		ereport(ERROR,
+				(errcode(ERRCODE_TOO_MANY_ARGUMENTS),
+				 errmsg_plural("cannot pass more than %d argument to a function",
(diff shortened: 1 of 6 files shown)


unbridled-41:
**Note on testability** (in response to the completeness review): the overflowed arrays here are reached only through internally-invoked paths (window-function support functions, ordered-set aggregate splitting, default-argument expansion for catalog-loaded functions), which cannot be constructed from plain SQL once `CREATE FUNCTION` caps declared arguments at `FUNC_MAX_ARGS`. That is why upstream shipped this fix with no regression cases, and why this PR does the same; the guard is a defense-in-depth port verified by the unchanged behavior of the full suites (259/259 core, 26/26 ivorysql_ora).

hanjianqiao:

PR #1774, opened 2026-08-29
Fix ascii() out-of-bound reads for invalid multibyte characters (CVE-2026-18024)

### LINKED ISSUE #1773: CVE-2026-18024: ascii() reads past end of buffer for invalid multibyte characters
### IvorySQL Version

### Summary

IvorySQL master (PostgreSQL master snapshot of 2026-04-16, see #1767) predates the upstream fix for **CVE-2026-18024** (PostgreSQL 18.6, 2026-08-13): *"ascii() reads past end of buffer for invalid multibyte characters"*.

`ascii()` decodes UTF-8 code points while assuming the input is well-formed:

* the sequence length implied by the leading byte is never checked against the actual input length;
* the leading byte is only loosely validated under `Assert` (compiled out in release builds), so bytes <= 0xC0 with the high bit set fall through;
* continuation bytes are never checked for the 0b10xxxxxx pattern.

A truncated or malformed sequence makes the function read past the end of the datum (out-of-bounds read). Upstream commit `a16c31d39c` adds the missing validation and reports `ERRCODE_CHARACTER_NOT_IN_REPERTOIRE`.

IvorySQL's own `ora_ascii()` (contrib/ivorysql_ora) only inspects the first byte and is not affected; the defect is in the core `ascii()` in `src/backend/utils/adt/oracle_compat.c`, which this fork ships.

### Suggested remediation

Port upstream commit `a16c31d39c` and add the regression cases to the encoding test. A patch doing exactly this (with the full core regression suite and the ivorysql_ora suite passing) follows as a PR.

unbridled-41:
## Issue

Fixes #1773 (part of the upstream security-sync gap tracked in #1767)

## Summary

Ports upstream commit `a16c31d39c` (PostgreSQL 18.6, 2026-08-13, **CVE-2026-18024**): `ascii()` now rejects truncated or malformed UTF-8 sequences instead of reading past the end of the datum. For every invalid case it reports:

```
ERROR:  invalid byte sequence for encoding "UTF8"
```

The `oracle_compat.c` hunk applies verbatim to this snapshot.

## Testing

* The upstream regression cases (eight `ascii(test_bytea_to_text(...))` calls covering truncated 2/3/4-byte sequences, an invalid continuation byte, and an invalid leading byte) were added to the **oracle_test** copy of the encoding test, which is the suite this tree actually runs - upstream's own additions only touched `src/test/regress`, which this fork does not execute in CI.
* Full core oracle regression suite: **259/259**, including the new encoding cases.
* Full `ivorysql_ora` oracle suite: **26/26** (IvorySQL's own `ora_ascii()` only reads the first byte and is unaffected).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `ascii()` handling of malformed UTF-8 input.
  - Invalid leading bytes, continuation bytes, and incomplete multibyte sequences now return clear encoding errors instead of potentially triggering assertions.

- **Tests**
  - Added coverage for truncated UTF-8 sequences and invalid byte combinations across supported regression test suites.

### DIFF
--- a/src/backend/utils/adt/oracle_compat.c
+++ b/src/backend/utils/adt/oracle_compat.c
@@ -952,8 +952,10 @@ ascii(PG_FUNCTION_ARGS)
 	text	   *string = PG_GETARG_TEXT_PP(0);
 	int			encoding = GetDatabaseEncoding();
 	unsigned char *data;
+	int			len;
 
-	if (VARSIZE_ANY_EXHDR(string) <= 0)
+	len = VARSIZE_ANY_EXHDR(string);
+	if (len <= 0)
 		PG_RETURN_INT32(0);
 
 	data = (unsigned char *) VARDATA_ANY(string);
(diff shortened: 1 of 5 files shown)


hanjianqiao:

PR #1790, opened 2026-08-30
Cross-check the type of a portal running EXECUTE or FETCH (CVE-2026-16239)

### LINKED ISSUE #1789: CVE-2026-16239: EXECUTE/FETCH portal can return rows of the wrong type
#### IvorySQL Version

#### Summary

`EXECUTE` and `FETCH` use two portals: an outer one for the statement itself, and an inner one running the query on its behalf. In IvorySQL master, the outer portal's row output is read without cross-checking the row type recorded by the inner portal (`FillPortalStore()` passes `NULL`/`NULL` for the expected tuple descriptor and mismatch message). A client that constructs a portal whose reported result type does not match the underlying query can cause rows of the wrong type to be handed back to the application (CVE-2026-16239).

The upstream fix passes the inner portal's `tupDesc` into the row-type cross-check, rejecting mismatches with `"query result type does not match portal result type"`.

#### Current Behavior

A malicious/broken client can bind a portal with a bogus result-type flag and receive rows of a type that does not match the executed statement; the server performs no consistency check on the returned row type.

#### Expected Behavior

- `FillPortalStore()` validates the outer portal's output descriptor against the inner portal's `tupDesc`
- A type mismatch is rejected with `ERROR: query result type does not match portal result type`
- `EXECUTE`/`FETCH` return only rows whose type matches the executed query

xuhuicai:
Port the CVE-2026-16239 fix (PostgreSQL 18.6, 2026-08-13 security release) to IvorySQL master. Closes #1789.

Upstream commit 9e02e2e1822 ported verbatim ("Cross-check the type of
a portal running EXECUTE or FETCH").

Summary: EXECUTE/FETCH use an outer and an inner portal; the outer
portal's rows were read without cross-checking against the inner
portal's tuple descriptor, so a client could receive rows of a type
that does not match the executed query. The fix passes the portal's
tupDesc into the row-type cross-check, rejecting mismatches with
"query result type does not match portal result type".

Notes:
- No oracle-specific code is involved (tcop/pquery.c is untouched by
  IvorySQL); no oracle changes needed.
- Verified: make check 249/249; the fix is a verbatim 2-line change
  with no applicable upstream regression test.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved portal result handling by validating returned data against the expected structure.
  * Added a clearer error message when result types do not match expectations.

### DIFF
--- a/src/backend/tcop/pquery.c
+++ b/src/backend/tcop/pquery.c
@@ -1154,8 +1154,8 @@ FillPortalStore(Portal portal, bool isTopLevel)
 									portal->holdStore,
 									portal->holdContext,
 									false,
-									NULL,
-									NULL);
+									portal->tupDesc,
+									gettext_noop("query result type does not match portal result type"));
 
 	switch (portal->strategy)
 	{


hanjianqiao:

PR #1803, opened 2026-08-30
fix(xml): handle empty serialized results

### LINKED ISSUE #1801: XML empty results can read before the output buffer
### Current behavior

The XML result cleanup helper `rv_newline()` unconditionally reads `tmp[len - 1]`. When an XPath expression matches no nodes, the serialized XML result is empty (`len == 0`), so the helper reads one byte before the allocated string.

A representative path is:

```sql
SELECT pg_catalog.octet_length(
  sys.extract('<a/>'::sys.xmltype, '/missing')::text
);
```

The result should be an empty XML value, without an out-of-bounds read.

### Expected behavior

Only inspect and remove a trailing newline when the serialized result has at least one byte.

### Proposed fix

Guard the `len - 1` access with `len > 0` and add regression coverage for an XPath with no matches.

btlqql:
## Summary

- guard the XML trailing-newline cleanup against zero-length output
- cover an XPath expression with no matching nodes

## Root cause

rv_newline() read tmp[len - 1] even when serialization produced an empty string, causing an out-of-bounds read before the buffer.

## Validation

- git diff --check
- Added focused ora_xml_functions regression coverage; full execution is delegated to the upstream Linux CI because the local Windows environment does not have the IvorySQL build toolchain.

Closes #1801

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_xml_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_xml_functions.out
@@ -26,6 +26,14 @@ INSERT INTO xmlnstest VALUES(1, xmltype('<soapenv:Envelope xmlns:soapenv="http:/
 --
 -- extrct(XML)
 --
+-- An XPath with no matches produces an empty XML value.  Keep the
+-- trailing-newline cleanup path safe for that zero-length result.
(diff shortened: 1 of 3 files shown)


hanjianqiao:
Thanks for the fix! We'll review it soon.

PR #1804, opened 2026-08-30
fix(oracle): normalize invalid SYS_CONTEXT lengths

### LINKED ISSUE #1802: SYS_CONTEXT ignores Oracle fallback for invalid length values
### Current behavior

The three-argument `SYS_CONTEXT` overload passes `length` directly to `left()`. Consequently, `length = 0` returns an empty value and values above 4000 are accepted.

```sql
SELECT length(sys_context('length_test', 'value', 0));
SELECT length(sys_context('length_test', 'value', 4001));
```

### Expected behavior

Oracle documents a valid range of 1 through 4000 bytes. When an invalid value is supplied, Oracle ignores it and uses the default length of 256.

### Proposed fix

Normalize values outside `1..4000` to 256 before truncating the context value, with regression coverage for valid, zero, and oversized lengths.

btlqql:
## Summary

- normalize SYS_CONTEXT length values outside 1..4000 to the Oracle default of 256
- add regression coverage for valid, zero, and oversized values

## Root cause

The three-argument wrapper passed the requested length directly to left(). Oracle instead ignores an invalid length and uses 256.

## Validation

- git diff --check
- Added focused ora_misc_functions regression coverage; full execution is delegated to the upstream Linux CI because the local Windows environment does not have the IvorySQL build toolchain.

Closes #1802

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `SYS_CONTEXT` length handling to better match Oracle behavior.
  * Invalid, null, fractional, or out-of-range lengths now use the default limit of 256.
  * Valid lengths are truncated by bytes, including for multibyte values.

* **Tests**
  * Added regression coverage for valid, fractional, null, zero, and out-of-range length arguments.
  * Added tests verifying byte-based truncation of multibyte context values.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_misc_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_misc_functions.out
@@ -5,6 +5,22 @@ SET nls_date_format = 'DD-MON-YY';
 SET nls_timestamp_format = 'DD-MON-YY HH24:MI:SS.US';
 SET nls_timestamp_tz_format = 'DD-MON-YY HH24:MI:SS.US TZ';
 set default_text_search_config = 'pg_catalog.simple';
+-- SYS_CONTEXT uses the default length (256) for an invalid length argument.
+SELECT set_config('length_test.value', repeat('x', 300), false) IS NOT NULL AS context_set;
+ context_set 
+-------------
+ t
+(1 row)
+
+SELECT length(sys_context('length_test', 'value', 12)) AS valid_length,
+       length(sys_context('length_test', 'value', 0)) AS zero_length,
+       length(sys_context('length_test', 'value', 4001)) AS oversized_length;
+ valid_length | zero_length | oversized_length 
+--------------+-------------+------------------
+           12 |         256 |              256
+(1 row)
+
+RESET length_test.value;
 -- error
 select decode(NULL, 1, 2018, 2, timestamp'2018-2-1 00:00:00', timestamp'2018-5-1 00:00:00');
 ERROR:  DECODE could not convert type timestamp to number
--- a/contrib/ivorysql_ora/expected/ora_misc_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_misc_functions.out
@@ -20,6 +20,27 @@ SELECT length(sys_context('length_test', 'value', 12)) AS valid_length,
            12 |         256 |              256
 (1 row)
 
+SELECT length(sys_context('length_test', 'value', NULL)) AS null_length,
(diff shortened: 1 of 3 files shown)


hanjianqiao:
Thanks for the fix! Could you please take a look at the comments posted by coderabbitai?

btlqql:
Both CodeRabbit findings were valid and were addressed in `4e2a4a89`, and I have now rebased the branch onto `master` (`03b24b1c46`) and re-verified the whole thing on a real build.

**1. Validate the numeric length before the cast** — the implementation no longer casts first:

```sql
CREATE or REPLACE FUNCTION sys.sys_context(a varchar2, b varchar2, c number)
RETURNS varchar2 AS $$
DECLARE
	res varchar2;
	requested_length integer := 256;
BEGIN
	/* Oracle ignores an out-of-range length and uses the default of 256. */
	IF c IS NOT NULL AND c >= 1 AND c <= 4000 THEN
		requested_length := trunc(c)::integer;
	END IF;
	SELECT sys.substrb(sys.sys_context(a, b), 1, requested_length) INTO res;
	RETURN res;
END;
$$ LANGUAGE plisql SECURITY INVOKER;
```

which is what the new assertions exercise — `NULL`, `0.5` and `2147483648` now all fall back to 256 instead of rounding, bypassing the comparison or raising an integer-range error:

```
 valid_length | zero_length | oversized_length
--------------+-------------+------------------
           12 |         256 |              256

 null_length | fractional_invalid_length | integer_overflow_length | fractional_valid_length
-------------+---------------------------+-------------------------+-------------------------
         256 |                       256 |                     256 |                      12
```

(`12.9` still gives 12 — truncation of a value inside the range, as Oracle does.)

**2. Byte-based truncation** — the length is applied with `sys.substrb()` instead of `left()`, so the limit counts bytes, and the multibyte case asserts it directly:

```
SELECT set_config('length_test.value', repeat('界', 100), false) IS NOT NULL AS multibyte_context_set;
 t
SELECT octet_length(sys_context('length_test', 'value', 5)) <= 5 AS multibyte_within_limit;
 t
```

With `left()`, 5 would have kept one 3-byte character plus a fragment, i.e. 5 bytes of a 6-byte string — the assertion is what pins the difference down.

**Verification** (Ubuntu 22.04, on top of `master` `03b24b1c46`, library rebuilt for this revision):

```
ok: 28   not ok: 1
not ok 17    - ora_xml_functions        # pre-existing: this tree is built without libxml
ok    11     - ora_misc_functions
$ diff expected/ora_misc_functions.out results/ora_misc_functions.out   # identical
```

Sorry about the delay — the bot's comments had been addressed in the code, but I never said so here.

PR #1805, opened 2026-08-30
fix(oracle): validate REGEXP_INSTR return option

### LINKED ISSUE #1799: REGEXP_INSTR accepts invalid return_option values
### Current behavior

`REGEXP_INSTR` accepts `return_option` values greater than 1. The implementation treats every nonzero value like `1`, so this succeeds even though `2` is outside the documented domain:

```sql
SELECT regexp_instr('barbar', 'b[^b]+', 1, 1, 2, 'i', 0) FROM dual;
```

### Expected behavior

Oracle defines only two valid values: `0` returns the first character position and `1` returns the position after the match. Values outside `0..1` should be rejected rather than silently treated as `1`.

### Proposed fix

Validate the upper bound together with the existing negative-value check and add regression coverage for `return_option = 2`.

yyqdbngt:
## Summary

- reject REGEXP_INSTR return_option values outside 0 and 1
- add regression coverage for return_option 2

## Root cause

The validation checked only the lower bound, while the execution path treated every nonzero value as the end-of-match option.

## Validation

- git diff --check
- Added focused ora_character_datatype_functions regression coverage; full execution is delegated to the upstream Linux CI because the local Windows environment does not have the IvorySQL build toolchain.

Closes #1799

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected `regexp_instr` validation so the `return_option` accepts only `0` or `1`.
  * Invalid values, such as `2`, now return an appropriate out-of-range error.

* **Tests**
  * Added regression coverage confirming invalid `return_option` values are rejected.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -1366,6 +1366,9 @@ select regexp_instr('barbar', 'b[^b]+', 1, 1, 1, 'i',0) from dual;
  4
 (1 row)
 
+-- return_option accepts only 0 (start) or 1 (end).
+select regexp_instr('barbar', 'b[^b]+', 1, 1, 2, 'i',0) from dual;
+ERROR:  argument '2' is out of range
 select regexp_instr('barbar', '(b[^b]+)', 1, 2, 0, 'i',0) from dual;
  regexp_instr 
 --------------
--- a/contrib/ivorysql_ora/sql/ora_character_datatype_functions.sql
+++ b/contrib/ivorysql_ora/sql/ora_character_datatype_functions.sql
@@ -738,6 +738,9 @@ select regexp_instr('barbar', 'b[^b]+', 1, 1, 0, 'i',0) from dual;
 
 select regexp_instr('barbar', 'b[^b]+', 1, 1, 1, 'i',0) from dual;
 
+-- return_option accepts only 0 (start) or 1 (end).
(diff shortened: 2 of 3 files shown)


hanjianqiao:
Thanks for the patch! It looks like the CI is failing on this PR.

muzimu217:
Ran this against a real Oracle instance (23ai) while testing the patch locally, and the results suggest Oracle is actually more permissive here than the fix assumes:

| return_option | Oracle 23ai | master | with this PR |
|---|---|---|---|
| 0 | 1 | 1 | 1 |
| 1 | 4 | 4 | 4 |
| 2 | 4 (no error — treated as end position) | 4 | ERROR out of range |
| 5 | 4 (no error) | 4 | ERROR out of range |
| -1 | ORA-01428: Argument -1 is out of range. | error | error |

So on real hardware Oracle rejects negatives (matching the existing `ret_opt < 0` check on master) but silently accepts any positive value as "return end position". The current master behavior actually matches that, while adding `> 1` would start rejecting values Oracle itself accepts.

The patch is well-made and the new test passes locally for me — but given the real-machine data, you may want to drop the `|| ret_opt > 1` part, or confirm against 21c/19c first in case older releases were stricter. Either way, thanks for making me look at this path — happy to help test a revised version.

yyqdbngt:
After checking the reported behavior against Oracle 23ai, positive return_option values greater than 1 are accepted and use the end-position behavior; only negative values are rejected. The current IvorySQL implementation already follows that behavior, while this PR would incorrectly reject valid positive values. I am therefore withdrawing this PR. Thanks for the review and CI feedback.

PR #1806, opened 2026-08-30
fix(oracle): mark bigint TO_DATE wrappers stable

### LINKED ISSUE #1800: bigint TO_DATE wrappers are incorrectly marked IMMUTABLE
### Current behavior

The two bigint wrappers are declared `IMMUTABLE`:

- `sys.to_date(bigint)`
- `sys.to_date(bigint, text)`

Both delegate to text overloads that are correctly declared `STABLE` because date parsing depends on session NLS settings. The stricter wrapper declaration lets PostgreSQL treat session-dependent conversions as immutable expressions.

### Expected behavior

The bigint wrappers should inherit the `STABLE` volatility of the text overloads they call.

### Proposed fix

Mark both wrappers `STABLE` and add catalog regression coverage for their `pg_proc.provolatile` values.

yyqdbngt:
## Summary

- mark both bigint TO_DATE wrappers STABLE
- add catalog regression coverage for their volatility

## Root cause

The wrappers were declared IMMUTABLE even though they delegate to STABLE text overloads whose parsing depends on session NLS settings.

## Validation

- git diff --check
- Added focused ora_datetime_datatype_functions regression coverage; full execution is delegated to the upstream Linux CI because the local Windows environment does not have the IvorySQL build toolchain.

Closes #1800

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected `sys.to_date` functions that accept `bigint` values to be session-dependent (`STABLE`) instead of incorrectly marked as immutable.
  * Preserved existing function behavior, signatures, return values, and parallel-safety attributes.

* **Tests**
  * Added regression coverage verifying the corrected volatility classification for both `bigint` overloads, including the text-format variant.


hanjianqiao:
Thanks for the patch! It looks like the CI is failing on this PR.

yyqdbngt:
## Root cause of the `oracle_regression` failure

The assertion resolved the overloads with `pg_get_function_identity_arguments()`, which prints type names **relative to `search_path`**. `pg_regress` runs with the default `"$user", public`, so the `bigint` arguments come back as `pg_catalog.int8`:

```
         args          | provolatile | proisstrict
-----------------------+-------------+-------------
 pg_catalog.int8       | s           | t
 pg_catalog.int8, text | s           | t
 text                  | s           | t
```

The `IN ('bigint', 'bigint, text')` list therefore matched nothing and the assertion returned `f` on a **correct** catalog — the `STABLE` change itself works.

```
 matched
---------
       0
```

## Fix

Resolve the overloads by OID through `::regprocedure`, which is search-path independent:

```sql
FROM pg_proc
WHERE oid IN ('sys.to_date(bigint)'::regprocedure,
              'sys.to_date(bigint, text)'::regprocedure)
  AND provolatile = 's';
```

`expected/ora_datetime_datatype_functions.out` is regenerated from a real run, since psql echoes the comment block.

## Validation

Locally, Ubuntu 24.04 / gcc 13, `./configure --with-uuid=e2fs`:

- before the fix: `not ok 1 - ora_datetime_datatype_functions` — the diff is exactly the `t`/`f` flip;
- after the fix: `ok 1 - ora_datetime_datatype_functions`, no `regression.diffs`;
- full `make oracle-check` in `contrib/ivorysql_ora`: **27 of 28 pass**. The only failure is the pre-existing `ora_xml_functions`, which is expected on a build without libxml.

CI should now be green.

PR #1834, opened 2026-09-02
fix(oracle): mark datetime length functions stable

### LINKED ISSUE #1830: LENGTH datetime overloads are incorrectly marked immutable
## Bug

The Oracle datetime overloads of `LENGTH` and `LENGTHB` are declared `IMMUTABLE`, but they convert the value through `oravarcharchar`. That conversion uses the session NLS datetime formats, so the result can change after `ALTER SESSION`.

The immutable declaration allows these functions in expression indexes even though changing `NLS_DATE_FORMAT`, `NLS_TIMESTAMP_FORMAT`, or `NLS_TIMESTAMP_TZ_FORMAT` can make the stored index expression stale.

## Reproduction

```sql
CREATE TABLE t_length_date(d DATE);
CREATE INDEX t_length_date_idx ON t_length_date (LENGTH(d));
ALTER SESSION SET NLS_DATE_FORMAT = 'YYYY-MM-DD HH24:MI:SS';
```

The same metadata issue applies to the `DATE`, `TIMESTAMP`, and `TIMESTAMP WITH TIME ZONE` overloads of both functions.

## Expected behavior

These six session-dependent overloads should be `STABLE`, matching the volatility of their NLS-aware output conversions and preventing their use in immutable-only expressions.

btlqql:
## Summary

- mark the `DATE`, `TIMESTAMP`, and `TIMESTAMP WITH TIME ZONE` overloads of `LENGTH` and `LENGTHB` as `STABLE`
- remove the duplicate `LENGTH(oradate)` declaration from the extension install script
- add regression coverage for all six catalog volatility entries

## Why

These overloads convert datetime values through Oracle character output, which depends on the session NLS datetime formats. Declaring them `IMMUTABLE` allows them in expression indexes even though a later `ALTER SESSION` can change their result.

## Testing

- `git diff --check`

Closes #1830

Assisted-by: OpenAI:GPT-5

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the volatility classification for date and timestamp variants of `LENGTH` and `LENGTHB` to account for session-specific NLS formatting.
  - These functions are now treated as stable rather than immutable, preventing their use where immutable behavior is required.

- **Tests**
  - Added coverage verifying the volatility of all six date and timestamp overloads.
  - Added checks confirming that indexes using these non-immutable expressions are rejected with the expected error.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -2919,21 +2919,28 @@ DETAIL:  Oracle's LENGTHC accepts only CHAR, VARCHAR2, NCHAR and NVARCHAR2.
 -- their answer is whatever nls_date_format says, so an index over them would
 -- go stale on the next alter session, and CREATE INDEX has to refuse it.
 --
--- length(d) is shown alongside to record that it does NOT refuse it yet.
--- sys.length/sys.lengthb declare their own datetime wrappers IMMUTABLE
--- (builtin_functions--1.0.sql:1429-1448, 1470-1485) even though the cast
--- inside them reaches the STABLE sys.oradate_out(); the wrapper hides that
--- from contain_mutable_functions(), which never looks inside a SQL function
--- body.  Fixing those is a separate change -- they are three whole function
--- families' worth of user-visible behaviour -- so the asymmetry is pinned
--- down here rather than left to be discovered.
+-- LENGTH and LENGTHB datetime overloads follow the same rule because their
+-- implicit text conversion also depends on the session's NLS formats.
 CREATE TABLE TEST_LENGTHC_IDX(c varchar2(50), d date);
 CREATE INDEX TEST_LENGTHC_IDX_I ON TEST_LENGTHC_IDX (lengthc(c));
 CREATE INDEX TEST_LENGTHC_IDX_D ON TEST_LENGTHC_IDX (lengthc(d));
 ERROR:  functions in index expression must be marked IMMUTABLE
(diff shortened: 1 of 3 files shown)


muzimu217:
Tested this on a real build (aarch64 Linux, master + this patch): full
`make oracle-check` passes 28/28 with `ENCODING=UTF8`, and the new catalog rows
all come back `s` for the six `LENGTH`/`LENGTHB` datetime overloads. The removed
duplicate `sys.length(sys.oradate)` declaration is safe — the extension script
only ever kept the later definition.

The premise checks out against real Oracle too. On Oracle 23ai Free (AL32UTF8),
`LENGTH(date)` moves with the session format exactly as the commit message says:

```
ALTER SESSION SET NLS_DATE_FORMAT = 'DD-MON-RR';        SELECT LENGTH(SYSDATE);  -- 9
ALTER SESSION SET NLS_DATE_FORMAT = 'YYYY-MM-DD';       SELECT LENGTH(SYSDATE);  -- 10
ALTER SESSION SET NLS_DATE_FORMAT = 'FMMonth DD, YYYY'; SELECT LENGTH(SYSDATE);  -- 17
```

So IMMUTABLE really was admitting indexes that go stale on the next ALTER
SESSION. With this patch applied, the admission is actually refused now, as data
for coderabbitai's suggestion about asserting this (I have no stake in whether
the extra regression rows are worth it):

```
CREATE INDEX t_i ON t(sys.length(d));
CREATE INDEX t_i ON t(sys.lengthb(d));
```

(before the patch both succeed on the same table.)

hanjianqiao:
Thanks

PR #1835, opened 2026-09-02
fix(oracle): preserve characters in LONG casts

### LINKED ISSUE #1831: LONG(n) casts can split multibyte characters
## Bug

The explicit typmod cast path for `LONG(n)` truncates the varlena payload with `len = maxlen`. In a UTF-8 database, a byte limit that falls inside a multibyte character leaves an invalid byte sequence in the resulting LONG value.

The same class of issue has already been fixed for `VARCHAR2(n BYTE)`: byte limits should be enforced without copying a partial character.

## Reproduction

```sql
SELECT CAST('中文' AS LONG(5));
SELECT LENGTHB(CAST('中文' AS LONG(5)));
```

## Expected behavior

An explicit `LONG(n)` cast should keep the longest valid character prefix that fits in `n` bytes. It must not store a partial multibyte character. ASCII truncation should remain unchanged.

btlqql:
## Summary

- clip explicit `LONG(n)` casts at a complete multibyte-character boundary
- preserve existing byte limits and ASCII truncation behavior
- add UTF-8 and ASCII regression cases for exact, partial, and zero-character prefixes

## Why

`orachar_to_long_with_typmod()` previously assigned `len = maxlen` before copying the payload. If the byte limit fell inside a UTF-8 character, the resulting LONG value contained an invalid partial character.

## Testing

- `git diff --check`

Closes #1831

Assisted-by: OpenAI:GPT-5

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `RAW(n)` casts to truncate input by bytes while preserving arbitrary byte values and exact byte limits.
  - Ensured zero-byte inputs produce empty results with the correct byte length.
  - Fixed `LONG(n)` casts to truncate multibyte text only at complete character boundaries.
  - Preserved expected behavior for ASCII text and values shorter than the limit.

- **Tests**
  - Added coverage for `RAW(n)` byte truncation, zero-byte inputs, and byte-length results.
  - Added coverage for multibyte and ASCII `LONG(n)` cast scenarios.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_raw_long.out
+++ b/contrib/ivorysql_ora/expected/ora_raw_long.out
@@ -141,6 +141,68 @@ select 'ff'::text = 'ff'::long(2);
  t
 (1 row)
 
+-- Explicit LONG(n) casts retain only complete multibyte characters.
+SELECT cast('中文' AS long(6)) = '中文'
+       AND lengthb(cast('中文' AS long(6))) = 6;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('中文' AS long(5)) = '中'
+       AND lengthb(cast('中文' AS long(5))) = 3;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT coalesce(lengthb(cast('中文' AS long(2))), 0) = 0;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('abcdef' AS long(4)) = 'abcd';
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('abc' AS long(4)) = 'abc';
+ ?column? 
+----------
+ t
+(1 row)
+
+-- Explicit bytea -> RAW(n) casts truncate by bytes, not by characters.
+SELECT cast('\xDEADBEEF'::bytea AS raw(4)) = '\xDEADBEEF'::bytea
+       AND lengthb(cast('\xDEADBEEF'::bytea AS raw(4))) = 4;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('\xDEADBEEF'::bytea AS raw(3)) = '\xDEADBE'::bytea
+       AND lengthb(cast('\xDEADBEEF'::bytea AS raw(3))) = 3;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('\x'::bytea AS raw(4)) = '\x'::bytea
+       AND lengthb(cast('\x'::bytea AS raw(4))) = 0;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('\x00FF10'::bytea AS raw(2)) = '\x00FF'::bytea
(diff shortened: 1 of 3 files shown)


hanjianqiao:
Thanks for the fixing! Could you please take a look at the comments posted by coderabbitai?

btlqql:
Both CodeRabbit findings were addressed, but checking them yourself turned up something worse that I had missed — thank you.

**1. The `bytea` finding (`raw_long.c:177`) is addressed.** The `RAW(n)` path is back to `len = maxlen`, so binary data is clipped by bytes. Confirmed on a real build:

```
select rawtohex(cast('\xDEADBEEF'::bytea as raw(3)));   => DEADBE
```

**2. The `bytea` test-coverage finding (`ora_raw_long.sql:37`) is addressed** — exact fit, partial, zero length and a leading-NUL case:

```
cast('\xDEADBEEF'::bytea as raw(4)) = '\xDEADBEEF' and lengthb(...) = 4
cast('\xDEADBEEF'::bytea as raw(3)) = '\xDEADBE'   and lengthb(...) = 3
cast('\x'::bytea as raw(4))         = '\x'          and lengthb(...) = 0
cast('\x00FF10'::bytea as raw(2))   = '\x00FF'
```

**3. But the LONG half of this PR was not actually fixing anything.** My two commits cancelled out: the first one (`e79183e065`, "preserve characters in LONG casts") changed the byte path to `pg_mbcliplen()`, and the second (`27ba2e9076`) changed it back to `len = maxlen`. Net effect on `raw_long.c` was a trailing tab. The `LONG(n)` code still

```
		len = maxlen;
```

and the regression I had written asserted the behaviour I *wanted*, not the behaviour that existed. Running it showed as much:

```
not ok 8     - ora_raw_long
    - t
    + f          -- cast('中文' AS long(5)) = '中' AND lengthb(...) = 3
    - t
    + f          -- coalesce(lengthb(cast('中文' AS long(2))), 0) = 0
```

So `#1831` was still open in the code, and the exercise above is what the issue describes. On `master` (library md5 `a6741ce3`), in a UTF8 database, Oracle mode:

```
select lengthb(cast('中文' as long(5)));     => 5      -- '中' plus half of '文'
select cast('中文' as long(5)) = '中';        => f
select lengthb(cast('中文' as long(2)));     => 2      -- a lone 2-byte fragment
```

The head now fixes that half too, with `pg_mbcliplen()` — the same call `oravarcharbyte.c` already uses for `VARCHAR2(n BYTE)`:

```
select lengthb(cast('中文' as long(5)));     => 3
select cast('中文' as long(5)) = '中';        => t
select lengthb(cast('中文' as long(2)));     => 0
```

With both halves in place (library md5 `57190792`, rebased onto `master` `03b24b1c46`):

```
ok: 28   not ok: 1
not ok 17    - ora_xml_functions        # pre-existing, this tree is built without libxml
$ diff expected/ora_raw_long.out results/ora_raw_long.out   # identical
```

The expected file is now generated by that run rather than written by hand. The `RAW(n)` and `LONG(n)` paths are deliberately different: `LONG(n)` clips at a character boundary because it stores text, `RAW(n)` clips by bytes because it stores binary.

PR #1836, opened 2026-09-02
fix(oracle): preserve characters in CHAR byte casts

### LINKED ISSUE #1828: CHAR(n BYTE) casts can split multibyte characters
## Bug

An explicit cast to `CHAR(n BYTE)` truncates the varlena payload by assigning `len = maxlen`. In a UTF-8 database, a byte limit that falls inside a multibyte character leaves an invalid byte sequence in the `CHAR` value.

This differs from the existing `VARCHAR2(n BYTE)` cast behavior, which clips at a complete character boundary.

## Reproduction

```sql
SELECT CAST('中文' AS CHAR(5 BYTE));
SELECT LENGTHB(CAST('中文' AS CHAR(5 BYTE)));
```

## Expected behavior

The explicit cast should retain only complete characters and pad the fixed-length `CHAR` value to the declared byte length. It must not store a partial multibyte character.

yyqdbngt:
## Summary

- clip explicit `CHAR(n BYTE)` casts at a complete multibyte-character boundary
- retain fixed-width CHAR semantics by padding the remaining bytes with spaces
- add UTF-8 and ASCII regression cases for exact and partial boundaries

## Why

`oracharbyte()` previously copied exactly `maxlen` bytes for an explicit cast. A byte limit inside a UTF-8 character therefore stored an invalid partial character. This brings fixed-width byte-semantic CHAR casts in line with the complete-character clipping already used by VARCHAR2 byte casts.

## Testing

- `git diff --check`

Closes #1828

Assisted-by: OpenAI:GPT-5

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed `CHAR(n BYTE)` conversions so multibyte characters are not split when values exceed the declared byte length.
  - Ensured truncated values are padded correctly to the requested byte length.
  - Improved handling of explicit and implicit conversions, including zero-length boundaries.

- **Tests**
  - Added regression coverage for multibyte `CHAR(n BYTE)` casts, truncation, byte-length behavior, and padding.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character.out
+++ b/contrib/ivorysql_ora/expected/ora_character.out
@@ -58,6 +58,40 @@ SELECT cast('abc' AS varchar2(4 byte)) = 'abc';
  t
 (1 row)
 
+-- Explicit CHAR(n BYTE) casts must not split multibyte characters.  CHAR
+-- values are padded back to the declared byte length after clipping.
+SELECT cast('中文' AS char(6 byte)) = '中文'
+       AND lengthb(cast('中文' AS char(6 byte))) = 6;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('中文' AS char(5 byte)) = '中'
+       AND lengthb(cast('中文' AS char(5 byte))) = 5;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT lengthb(cast('中文' AS char(2 byte))) = 2;
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('abcdef' AS char(4 byte)) = 'abcd';
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT cast('abc' AS char(4 byte)) = 'abc';
+ ?column? 
+----------
+ t
+(1 row)
+
 INSERT INTO TEST_ORACHAR SELECT generate_series(1,10), generate_series(1,10);
 INSERT INTO TEST_ORAVARCHAR SELECT generate_series(1,10), generate_series(1,10);
 -- Comparison operator
--- a/contrib/ivorysql_ora/sql/ora_character.sql
+++ b/contrib/ivorysql_ora/sql/ora_character.sql
@@ -28,6 +28,16 @@ SELECT coalesce(lengthb(cast('中文' AS varchar2(2 byte))), 0) = 0;
 SELECT cast('abcdef' AS varchar2(4 byte)) = 'abcd';
 SELECT cast('abc' AS varchar2(4 byte)) = 'abc';
 
+-- Explicit CHAR(n BYTE) casts must not split multibyte characters.  CHAR
+-- values are padded back to the declared byte length after clipping.
+SELECT cast('中文' AS char(6 byte)) = '中文'
+       AND lengthb(cast('中文' AS char(6 byte))) = 6;
+SELECT cast('中文' AS char(5 byte)) = '中'
+       AND lengthb(cast('中文' AS char(5 byte))) = 5;
+SELECT lengthb(cast('中文' AS char(2 byte))) = 2;
+SELECT cast('abcdef' AS char(4 byte)) = 'abcd';
+SELECT cast('abc' AS char(4 byte)) = 'abc';
+
 INSERT INTO TEST_ORACHAR SELECT generate_series(1,10), generate_series(1,10);
 
 INSERT INTO TEST_ORAVARCHAR SELECT generate_series(1,10), generate_series(1,10);
--- a/contrib/ivorysql_ora/src/datatype/oracharbyte.c
+++ b/contrib/ivorysql_ora/src/datatype/oracharbyte.c
@@ -338,7 +338,7 @@ oracharbyte(PG_FUNCTION_ARGS)
 					(errcode(ERRCODE_STRING_DATA_RIGHT_TRUNCATION),
 					 errmsg("value too long for type char(%d byte)",
 							maxlen)));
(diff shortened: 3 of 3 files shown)


hanjianqiao:
Thanks for the fixing! We'll review it shortly.

muzimu217:
Nice find and a clean fix — I built this branch and ran the suite locally (`oracle-check` passes 28/28 for me on a UTF8 cluster), and master indeed stores a broken partial character for `cast('中文' as char(5 byte))`, so the clipping itself is definitely needed.

I also compared the cast behavior against a real Oracle instance (23ai) while I had it up, and two corner cases behave differently there — sharing in case it matters for this fix:

| expression | this PR | Oracle 23ai |
|---|---|---|
| `cast('中文' as char(5 byte))` | '中' + 2 spaces (len 5, padded back) | '中' with no padding (len 3) |
| `cast('中文' as char(2 byte))` | 2 spaces (len 2) | NULL (empty result) |

So on Oracle the truncated result keeps its actual length instead of being re-padded, and when nothing fits the result is empty/NULL rather than all spaces. The exact-fit and ASCII cases all match (`char(6 byte)`, `char(4 byte)`, `'ab中' as char(5 byte)`).

Not sure whether re-padding is a deliberate choice to keep fixed-width CHAR semantics after clipping — just pointing at the difference since I happened to have the data. Happy to help test whatever direction gets chosen.

yyqdbngt:
Thanks — this is a real gap, and I now have measurements for both directions you mentioned. Everything below is from a local build (Ubuntu 22.04, kernel 6.8, GCC 12.3, `initdb -USYSTEM -moracle`, `SET ivorysql.compatible_mode = oracle`).

## 1. `master` really does store a broken partial character

```
select octet_length(cast('中文' as char(6 byte))),  -> 6   hex e4b8ade69687   ('中文')
select cast('中文' as char(5 byte)) ...             -> ERROR: invalid byte sequence for encoding "UTF8": 0xe6 0x96
select cast('中文' as char(2 byte)) ...             -> ERROR: invalid byte sequence for encoding "UTF8": 0xe4 0xb8
```

So the clipping in this PR is not cosmetic: on `master` the two truncated casts produce values that cannot even be printed. With the patch: `lengthb` = 5 and 2 respectively, and no invalid sequence.

## 2. Where this build and Oracle differ

Your table matches what I measure, with the padding spelled out as `lengthb`:

| expression | `lengthb` on this build | Oracle 23ai (your data) |
|---|---|---|
| `cast('中文' as char(5 byte))` | 5 (i.e. `'中'` + 2 blanks) | 3, no padding |
| `cast('中文' as char(2 byte))` | 2 (all blanks) | NULL |

The second row is the interesting one, because this build already behaves the way Oracle does for the sibling type. Same input through `varchar2(n byte)`:

```
select octet_length(cast('中文' as varchar2(2 byte))),   -> 0
       length(cast('中文' as varchar2(2 byte)))          -> 0
       coalesce(lengthb(cast('中文' as varchar2(2 byte))), 0) = 0  -> t
```

`oravarcharbyte()` calls `pg_mbcliplen()` and returns a zero-length value, and the existing test for `varchar2(2 byte)` in `ora_character` asserts exactly that (`coalesce(lengthb(...), 0) = 0`). `oracharbyte()` clips the same way but then blank-pads back to the declared length, so "nothing fits" becomes a value made only of blanks instead of an empty one.

## 3. What I would like your call on

Two separate questions, and I do not want to change either without you agreeing, because re-padding is what fixed-width `CHAR` normally does:

- **(a) nothing fits** — `cast('中文' as char(2 byte))`. I think matching Oracle here is right, and it also makes `char(n byte)` consistent with this build's own `varchar2(n byte)`: return a zero-length value instead of blanks. Small change in `oracharbyte()`: skip the `memset` padding when `pg_mbcliplen()` returned 0, plus regenerated expectations.
- **(b) mid-character clip** — `cast('中文' as char(5 byte))`. Oracle returns the clipped value unpadded (3 bytes), this build pads back to 5. That is the deliberate `CHAR` invariant as the test comment says, and dropping it would make `cast('abc' as char(4 byte))` unpadded too (currently `lengthb` = 4), which is further from Oracle for the ordinary case. I would keep the padding unless you tell me Oracle's behaviour here is what the type should reproduce.

Tell me which way you want either one and I will push it with fresh expectations; if you would rather keep this PR to "never split a multibyte character", I will leave both as they are and we can file (a) separately.

中文摘要：谢谢指出，这个差异是真的，我在本地真机上把两个方向都量化了。(1) master 确实存入了残缺字节——`cast('中文' as char(5 byte))` 和 `char(2 byte)` 直接报 `invalid byte sequence for encoding "UTF8"`，所以这个裁剪不是表面问题；打上补丁后 `lengthb` 分别为 5 和 2。(2) 与 Oracle 的两处差异我复现了，并且发现本仓库自身就不一致：同样输入走 `varchar2(2 byte)` 时 `oravarcharbyte()` 返回零长度值（现有测试断言 `coalesce(lengthb(...),0) = 0`），而 `oracharbyte()` 会补空格。(3) 想请您定夺：(a)"什么都放不下"的情况（`char(2 byte)`）我认为应当对齐 Oracle 与本仓库的 varchar2，返回空值而不是空格——改动很小，我可以立刻推；(b)"截断在多字节字符中间"的情况（`char(5 byte)`）Oracle 不补空格而本实现补齐到声明长度，这是 CHAR 定宽语义，去掉它会让普通的 `cast('abc' as char(4 byte))` 也不补齐（目前 `lengthb` = 4），反而离 Oracle 更远，我倾向保留，除非您认为这里必须复刻 Oracle。您定方向我就改并重新生成期望输出。

PR #1841, opened 2026-09-03
Raise NO_DATA_FOUND when UTL_FILE.GET_LINE reads past EOF

### LINKED ISSUE #1840: UTL_FILE.GET_LINE does not raise NO_DATA_FOUND at end of file
UTL_FILE.GET_LINE returns NULL when no text was read because of end of
file, instead of raising NO_DATA_FOUND.  The function's own comment
lists NO_DATA_FOUND among its exceptions, but the implementation only
logs it at LOG level (which also writes a "no data found" line into the
server log on every EOF read) and then returns NULL.

Oracle documents for GET_LINE: "If no text was read due to end of file,
the NO_DATA_FOUND exception is raised."  The documented PL/SQL read
pattern therefore cannot be used in PL/iSQL:

    LOOP
        BEGIN
            UTL_FILE.GET_LINE(f, line);
        EXCEPTION
            WHEN NO_DATA_FOUND THEN
                EXIT;
        END;
    END LOOP;

because the exception never fires.

Reproduced on current master in an Oracle-mode database:

- write two lines, read three: the third call returns NULL instead of
  raising;
- the first read of an empty file returns NULL instead of raising.

I have a small patch that raises the exception (following the pattern
PL/iSQL already uses for strict SELECT INTO) and will open a PR.

muzimu217:
## Summary

UTL_FILE.GET_LINE listed NO_DATA_FOUND among its exceptions but only
logged it at LOG level and returned NULL when no text was read because
of end of file.  As a result PL/iSQL callers could not end read loops
with WHEN NO_DATA_FOUND, and every read past EOF also emitted a
"no data found" line into the server log.

Oracle documents for GET_LINE: "If no text was read due to end of file,
the NO_DATA_FOUND exception is raised."

The EOF branch now raises the exception, using the same pattern PL/iSQL
already uses for strict SELECT INTO (ereport(ERROR,
ERRCODE_NO_DATA_FOUND)).  The read loops in the existing regression
tests catch NO_DATA_FOUND per call, so their output is unchanged; new
cases pin the exception on a two-line file and on the first read of an
empty file, including the exact sqlerrm.

Note this is a deliberate behaviour change for callers that relied on
NULL at EOF.

## Testing

- oracle regression suite: 28/28 in a C locale, a UTF8 locale, and the
  default locale
- make check: 249/249
- full-tree Meson/Ninja build: 2729/2729 targets

Fixes #1840

## Summary by CodeRabbit

* **Bug Fixes**
  * `utl_file.get_line` now raises `NO_DATA_FOUND` when reading beyond the end of a file or from an empty file.
  * End-of-file handling in affected file-reading scenarios now terminates cleanly instead of returning `NULL`.

### DIFF
--- a/contrib/ivorysql_ora/expected/utl_file.out
+++ b/contrib/ivorysql_ora/expected/utl_file.out
@@ -47,12 +47,22 @@ begin
     utl_file.get_line(f, line);
     while line is not null and i < 11 loop
         raise notice '[%] >>%<<', i, line;
-        utl_file.get_line(f, line);
+        begin
+            utl_file.get_line(f, line);
+        exception
+            when NO_DATA_FOUND then
+                line := null;
+        end;
         i:=i+1;
     end loop;
     while line is not null loop
         raise notice '[%] >>%<<', i, line;
-        utl_file.get_line(f, line, 3);
+        begin
+            utl_file.get_line(f, line, 3);
+        exception
+            when NO_DATA_FOUND then
+                line := null;
+        end;
         i:=i+1;
     end loop;
     raise notice '>>%<<', line;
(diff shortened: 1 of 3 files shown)


hanjianqiao:
Thanks for your fix! We'll review it shortly.

PR #1846, opened 2026-09-03
fix(oracle): bound NEXT_DAY weekday conversion

### LINKED ISSUE #1842: NEXT_DAY can read past UTF-8 weekday arguments
NEXT_DAY converts localized weekday names to UTF-8 and then calls strlen(). When the database encoding is already UTF-8, pg_server_to_any() may return the original varlena payload, which is length-delimited rather than NUL-terminated. This can read beyond the argument buffer. Use the returned pointer together with the original byte length when no conversion occurs, and add regression coverage.

btlqql:
## Summary
- keep NEXT_DAY weekday conversion length-bounded when UTF-8 conversion is a no-op
- cover a localized UTF-8 weekday name in the datetime regression test

## Testing
- git diff --check

Closes #1842

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `NEXT_DAY` handling for length-delimited text values, preventing incorrect reads during date calculations.
  * Added support for Chinese day names, such as “星期一,” in `NEXT_DAY` expressions.

* **Tests**
  * Expanded date-function coverage and expected results for localized day-name input.


hanjianqiao:

muzimu217:
Gave this a spin on a real build (aarch64 Linux, master + this patch) — full
`oracle-check` passes 28/28 with `ENCODING=UTF8`, and the premise checks out in
the source: `next_day()` reads the weekday argument via
`VARDATA_ANY`/`VARSIZE_ANY_EXHDR`, so the payload isn't guaranteed to be
NUL-terminated, while `pg_server_to_any()` returns the original pointer when no
conversion is needed. `(changstr == str) ? len : strlen(changstr)` is exactly
how that contract is meant to be handled.

It also lines up with a comment `utl_url.c` in this tree already carries
("Work on a NUL-terminated copy: pg_server_to_any() may hand the input
[pointer back]"), so this brings `next_day()` in line with assumptions the
rest of the package code already makes.

Full disclosure: I tried to make the unfixed code misbehave (fresh `substr()`
datums, inline heap tuples with a non-zero short-varlena column right behind
the weekday column) and it kept returning correct results on my build.
Looking closer at `weekday_search()`, that's expected: it only compares the
first `strlen(weekday-name)` bytes of the input (prefix match), so an
over-read length doesn't produce wrong results. The practical risk of the old
code is `strlen()` wandering past the end of the datum on an unlucky heap
layout, not a wrong return value — so I don't have a reproducer to offer, but
the contract argument stands on its own.

One thing I noticed while comparing against Oracle, sharing in case it matters:
Oracle ties the weekday spelling to the session date language. On 23ai Free
(AL32UTF8) here:

```
SELECT NEXT_DAY(DATE '2018-02-12','星期一') FROM DUAL;   -- default (AMERICAN)
=> ORA-01846: not a valid day of the week

ALTER SESSION SET NLS_DATE_LANGUAGE='SIMPLIFIED CHINESE';
SELECT NEXT_DAY(DATE '2018-02-12','星期一') FROM DUAL;
=> 19-2月 -18
```

and symmetrically 'monday' errors under the Chinese date language. IvorySQL
with this patch accepts both spellings regardless of language settings. That's
the more forgiving behavior and I can't think of ported code that would *rely*
on ORA-01846, so this is just a note that the new expected output documents a
small deliberate superset of Oracle's behavior.

(Reference for the Oracle side: NEXT_DAY in the 23ai SQL Language Reference,
https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/NEXT_DAY.html)

yyqdbngt:
Thanks for the detailed testing and for digging into the contract — the `utl_url.c` precedent is a good find, and you are right that this is a contract fix rather than a wrong-result fix: `weekday_search()` only prefix-matches, so the visible symptom was unlikely, while `strlen()` walking past the datum on an unlucky heap layout is the real hazard.

On the `NLS_DATE_LANGUAGE` note: you are right that adding the `星期一` case documents a deliberate superset, and I would rather state it than leave it implicit, because a later reader will otherwise assume the Chinese spelling is validated against the session language. Concretely, IvorySQL here accepts both spellings regardless of `NLS_DATE_LANGUAGE`, where Oracle raises `ORA-01846` for the spelling that does not match the session language.

I am leaving the behaviour as-is (accepting both) since, as you say, no ported code should rely on `ORA-01846`, and rejecting a spelling would be a behaviour change beyond the scope of this fix — but I will say so in the thread here so the expected output is not read as an Oracle-parity claim.

btlqql:
Rebased onto current `master` (`03b24b1c46`) and force-pushed — head is now `50bd8204a9`. The branch was 676 commits behind, so CI re-runs and the `BEHIND` state is gone. The rebase is content-preserving: the diff against `origin/master` is byte-identical before and after (2513 bytes, same three files).

Re-verified on the rebased tree (Ubuntu 22.04, gcc 12.3, `--with-uuid=e2fs --without-readline --without-zlib --without-icu --without-libxml`, UTF8 test database):

- `make -C contrib/ivorysql_ora oracle-check`: **28 ok / 1 not ok**, the only failure being `ora_xml_functions` (unrelated — this build is `--without-libxml`).
- `diff -u expected/ora_datetime_datatype_functions.out results/ora_datetime_datatype_functions.out` is **empty**, so the expectation you saw, including the new `NEXT_DAY(date '2018-02-12','星期一')` row, matches the real output.

One honest caveat about how much the regression test can prove. I tried to make the fixed defect observable by pairing master's C code with just this PR's test lines and running the suite: **the test passes there too** (`28 ok / 1 not ok`, empty diff). I then ran eight inputs against both builds directly —

```
NEXT_DAY(date '2018-02-12','monday')                      2018-02-19  (both)
NEXT_DAY(date '2018-02-12','星期一')                       2018-02-19  (both)
NEXT_DAY(date '2018-02-12', chr(26143)||chr(26399)||chr(19968))   2018-02-19  (both)
NEXT_DAY(date '2018-02-12', ('星期' || '一'))              2018-02-19  (both)
NEXT_DAY(date '2018-02-12', upper('星期一'))               2018-02-19  (both)
NEXT_DAY(date '2018-02-12','星期二')                       2018-02-13  (both)
NEXT_DAY(date '2018-02-12','星期日')                       2018-02-18  (both)
NEXT_DAY(date '2018-02-12','sunday')                      2018-02-18  (both)
```

— and both builds agree on every one, so this configuration does not surface the defect. That is expected for the change that is actually in the diff: on a UTF8 database converting to `PG_UTF8` is a no-op, `pg_server_to_any()` returns the input pointer, and the text payload is length-delimited rather than NUL-terminated, so `strlen()` may read past it. It is a latent out-of-bounds read rather than a wrong-answer bug, which is why the interesting property to keep is the bound (`changstr == str ? len : strlen(changstr)`) and the regression row is really a guard against regressions in the multibyte path.

I am flagging this rather than claiming a red/green pair I could not produce: if you would prefer the PR to carry a demonstrated failure, the options I can execute are (a) build the extension and server under `-fsanitize=address` and show the master read, or (b) drop the extra regression row as non-discriminating. Happy to do either — just say which.

Thanks @muzimu217 for the aarch64 build check; the 28/28 still holds on the rebased tree.

PR #1847, opened 2026-09-03
fix(sysview): mark catalog helpers stable

### LINKED ISSUE #1843: System view catalog helpers are incorrectly immutable
SYS._ORA_SRC_COUNT and SYS._ORA_ARGS_COUNT read pg_proc but are declared IMMUTABLE. CREATE OR REPLACE can change the source or argument metadata while preserving the function OID, so a prepared statement can keep a folded stale result. These catalog readers should be STABLE, with regression coverage for their catalog-dependent volatility.

btlqql:
## Summary
- mark SYS._ORA_SRC_COUNT and SYS._ORA_ARGS_COUNT as STABLE because they read pg_proc
- keep the consolidated and split sysview installation scripts in sync
- verify both helpers expose catalog-dependent volatility

## Testing
- git diff --check

Closes #1843

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the stability classification of internal system-view helper functions from immutable to stable.
  * Added coverage to verify the corrected function behavior and catalog metadata.


hanjianqiao:

muzimu217:
Built and ran the full oracle-check suite with this applied — all 28 tests pass on my UTF8 cluster, including the new provolatile assertions. The change itself is the right call: functions that read catalog tables get re-evaluated per statement on Oracle, and marking them STABLE prevents the planner from folding their results as constants while still allowing it to use them in index scans. Nothing else jumped out — small, focused, and well covered.

btlqql:
Rebased onto current `master` (`03b24b1c46`) and force-pushed — head is now `68fc4bbcde`. The branch was 673 commits behind, so this refreshes CI and removes the `BEHIND` state. The rebase is content-preserving: `git diff origin/master...<old head>` and `git diff origin/master...<new head>` are byte-identical (3133 bytes, same five files), so nothing about the change you reviewed was altered.

Verified on a real build (Ubuntu 22.04, gcc 12.3, `--with-uuid=e2fs --without-readline --without-zlib --without-icu --without-libxml`):

**Red/green probe.** The change is in the extension SQL, so I kept master's `sysview--1.0.sql` / `sysview_source.sql` / `sysview_arguments.sql` and applied only this PR's `sql/ora_sysview.sql` + `expected/ora_sysview.out`:

```
helpers_are_stable
------------------
 f
(1 row)
```

and against the PR's expected file:

```
- t
+ f
```

i.e. the new assertion fails on master and the fix is what makes it pass. Same run reported `# 2 of 29 tests failed` (`ora_sysview` plus `ora_xml_functions`).

**With this PR** the same assertion reports `t` and `diff -u expected/ora_sysview.out results/ora_sysview.out` is empty (`md5sum` of both: `27543570f7c11bdb4b14586bdac68982`). Full `make -C contrib/ivorysql_ora oracle-check` on the rebased tree: **28 ok / 1 not ok**, the only failure being `ora_xml_functions`, which is unrelated — this build is `--without-libxml`.

Two harness notes that cost me a while and may be worth knowing for anyone re-running this by hand:

- `contrib/ivorysql_ora`'s `oracle-check` runs `pg_regress` against `tmp_install`, and `oracle-temp-install` only populates it when the directory does not exist. Since this PR changes the extension SQL rather than the shared library (the rebuilt `ivorysql_ora.so` md5 is unchanged), a stale `tmp_install` silently keeps the old declarations — the probe only became meaningful after `rm -rf tmp_install` before each side.
- `TESTS=` is not honored by that target; it always runs the full `ORA_REGRESS` list (29 tests), which is how I got the red run above.

Thanks @muzimu217 for the earlier build check — the 28/28 you reported still holds on the rebased tree.

PR #1848, opened 2026-09-03
fix(oracle): guard SUBSTRB end boundary

### LINKED ISSUE #1844: SUBSTRB can inspect past the end of multibyte input
The multibyte SUBSTRB loop evaluates pg_mblen(subStrEnd) even after subStrEnd has reached the end of the varlena payload. A substring that ends exactly at the source boundary can therefore read past the input buffer. Guard the pointer before inspecting the next character and add boundary regression coverage.

yyqdbngt:
## Summary

- stop `SUBSTRB` from calling `pg_mblen()` after its scan pointer reaches the end of the varlena payload
- keep the existing byte-substring result unchanged
- exercise the exact-end UTF-8 boundary path in regression coverage

## Why

When the requested byte range ends exactly at the source boundary, the loop advances `subStrEnd` to the one-past-end pointer. The old loop condition then evaluates `pg_mblen(subStrEnd)` before determining that another character cannot fit. Because a multibyte character length is always positive, this usually does not change the returned value; the defect is the out-of-bounds read itself, which depends on the adjacent memory layout and is not reliably observable in a normal result assertion.

The added regression case exercises this boundary and protects the expected result. Detecting the old read directly requires an instrumented memory-safety build.

## Testing

- `git diff --check`
- the boundary regression preserves the complete UTF-8 result
- an independent aarch64 Linux verification reported the full `oracle-check` suite passing 28/28 with `ENCODING=UTF8`

Closes #1844

Assisted-by: OpenAI:GPT-5

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -1705,6 +1705,12 @@ SELECT SUBSTRB('ABCDEFG',5) "Substring with bytes" FROM DUAL;
  EFG
 (1 row)
 
+SELECT SUBSTRB('你好',1,6) = '你好' AS complete_boundary FROM DUAL;
+ complete_boundary 
+-------------------
+ t
+(1 row)
+
 /*
  * instrb
  */
--- a/contrib/ivorysql_ora/sql/ora_character_datatype_functions.sql
+++ b/contrib/ivorysql_ora/sql/ora_character_datatype_functions.sql
@@ -832,6 +832,7 @@ DROP TABLE regexp_temp;
  */
 SELECT SUBSTRB('ABCDEFG',5,4.2) "Substring with bytes" FROM DUAL;
 SELECT SUBSTRB('ABCDEFG',5) "Substring with bytes" FROM DUAL;
+SELECT SUBSTRB('你好',1,6) = '你好' AS complete_boundary FROM DUAL;
 
 /*
  * instrb
--- a/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
@@ -1354,7 +1354,8 @@ text_substring_byte(Datum str, int32 start, int32 length, bool length_not_specif
 		}
 
 		subStrEnd = subStrBegin;
-		while (subStrEnd + pg_mblen(subStrEnd) <= sliceBegin + E1 - 1)
+		while (subStrEnd < sliceBegin + E1 - 1 &&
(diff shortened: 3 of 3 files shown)


hanjianqiao:

muzimu217:
Built this on aarch64 Linux (master + patch) and ran the full `oracle-check`
(28/28 with `ENCODING=UTF8`). Since SUBSTRB boundaries are where the
interesting bugs live, I also cross-checked the byte-level behavior against
Oracle 23ai Free (AL32UTF8) — and the partial-character padding matches
exactly, which is nice to see:

```
                                Oracle 23ai (DUMP)          this build
SUBSTRB('你好',1,5)    Typ=1 Len=5: 228,189,160,32,32    len=5 hex=e4bda02020
SUBSTRB('你好',1,4)    Typ=1 Len=4: 228,189,160,32       len=4 hex=e4bda020
SUBSTRB('你好',1,6)    Typ=1 Len=6: 228,189,160,229,165,189    len=6, full string
```

Two observations from the same session, both "sharing data, not asking for
anything":

1. I couldn't construct an input where the added guard changes the outcome.
   Old and new code give identical results for everything I tried — `(1,6)`,
   `(1,5)`, `(1,4)`, `(2,2)`, `(1,0)`, `(1,-1)`, and the ASCII cases — and the
   new regression test passes on unpatched master in my environment as well.
   Since `pg_mblen()` is always >= 1, the second condition already looks
   unreachable whenever `subStrEnd >= sliceBegin + E1 - 1`... but I may simply
   be missing the case that motivated the patch; if you have one, I'm curious.

2. There is one spot where Oracle and this build disagree, and it's unrelated
   to the guard:

```
SUBSTRB('你好',2,2)
Oracle 23ai:  two spaces, LENGTHB = 2   (DUMP: Typ=1 Len=2: 32,32)
this build:   one space,  octet_length = 1
```

   Looks like when the start position lands mid-character, the
   `subStrEnd == subStrBegin` early return hands back a literal `" "`/`"  "`
   and drops `prefixSpaceCnt`. Mid-character starts are probably rare in real
   code, so this is more "data point on record" than anything else.

(Reference for the Oracle side: the SUBSTR family in the 23ai SQL Language
Reference, https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/SUBSTR.html)

yyqdbngt:
Thanks for the detailed verification. I agree that the ordinary result is unchanged: once the scan pointer reaches the end, any positive `pg_mblen()` result makes the old condition false. The fix is specifically to avoid evaluating `pg_mblen()` at the one-past-end address, so reproducing the memory read depends on layout or an instrumented build. I updated the PR description to make that distinction explicit and no longer present the regression as a before/after output failure. The mid-character start padding difference is separate from issue #1844 and should not be folded into this boundary-safety patch.

PR #1852, opened 2026-09-03
Fix Oracle SUBSTR edge-case semantics

### LINKED ISSUE #1850: Oracle-compatible SUBSTR uses PostgreSQL semantics for zero/negative positions and nonpositive lengths
### Summary

In `ivorysql.compatible_mode = oracle`, `SUBSTR()` follows PostgreSQL `pg_catalog.substr(text, int[, int])` behavior for documented Oracle edge cases. This causes migrated Oracle SQL to return incorrect substrings or raise an error.

Affected cases:

- position `0` is not treated as `1`;
- negative positions do not count backward from the end;
- a zero length returns an empty value rather than `NULL`;
- a negative length raises `ERROR: negative substring length not allowed` rather than returning `NULL`.

### Environment

Reproduced locally against IvorySQL master commit `c3529ba4789ee220caef858e58bbe6f1563ad14` (IvorySQL 5beta1 / PostgreSQL 19devel), with:

- `ivorysql.compatible_mode = oracle`
- `server_encoding = UTF8`
- `ivorysql_ora = 1.0`

### Reproduction

```sql
CREATE EXTENSION IF NOT EXISTS ivorysql_ora;

SELECT substr('abcdef', 2, 2) AS normal_positive;
SELECT substr('abcdef', 0, 2) AS position_zero_with_length;
SELECT substr('abcdef', -2, 2) AS negative_position_with_length;
SELECT substr('abcdef', -2) AS negative_position_no_length;
SELECT substr('abcdef', 2, 0) IS NULL AS zero_length_returns_null;
SELECT substr('abcdef', 2, -1) IS NULL AS negative_length_returns_null;

-- Controls: SUBSTRB already handles these Oracle cases.
SELECT sys.substrb('abcdef', -2, 2) AS substrb_negative_position;
SELECT sys.substrb('abcdef', 2, -1) IS NULL AS substrb_negative_length;
```

The deterministic formal harness was run from `C:\Users\26517\Documents\ChatGPT\IvorySQL\.worktrees\issue-research\.research\substr-mainpath\run_substr_formal_repro.sh`; output is recorded in `formal_repro_run.out`.

Observed result:

```text
label                              | expected | actual | error_text                              | passed
-----------------------------------+----------+--------+------------------------------------------+-------
negative_length_returns_null       | NULL     | NULL   | negative substring length not allowed    | false
negative_position_no_length        | ef       | abcdef | NULL                                     | false
negative_position_with_length      | ef       |        | NULL                                     | false
normal_positive                    | bc       | bc     | NULL                                     | true
position_zero_with_length          | ab       | a      | NULL                                     | false
substrb_control_negative_length    | NULL     | NULL   | NULL                                     | true
substrb_control_negative_position  | ef       | ef     | NULL                                     | true
zero_length_returns_null           | NULL     |        | NULL                                     | false

passed_count = 3, failed_count = 5, total_count = 8
```

A common migrated expression such as `SUBSTR(col, 1, INSTR(col, '-') - 1)` also becomes a hard error when the delimiter is absent (`INSTR` returns `0`, yielding length `-1`); Oracle returns `NULL` for this case.

### Expected behavior

Oracle documents that position `0` is treated as `1`, negative positions count backward from the end, omitted length returns through the end, and a length less than `1` returns `NULL`. Therefore the expected results include:

```text
substr('abcdef', 0, 2)       -> 'ab'
substr('abcdef', -2, 2)      -> 'ef'
substr('abcdef', -2)         -> 'ef'
substr('abcdef', 2, 0)       -> NULL
substr('abcdef', 2, -1)      -> NULL
```

### Code pointers

- `src/backend/utils/adt/varlena.c`: `pg_catalog.substr` resolves to `text_substr` / `text_substr_no_len`, whose `text_substring()` implements PostgreSQL semantics and errors on negative length.
- `contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c`: `sys.substrb()` already implements Oracle-style byte semantics, but there is no corresponding Oracle-compatible character `sys.substr` implementation.

### Suggested fix

Add Oracle-mode character-based `SUBSTR` handling without changing PostgreSQL-mode behavior. Cover zero and negative positions, omitted length, non-positive lengths, empty-result-to-`NULL` behavior, and multibyte character counting with regression tests. Keep `SUBSTRB` behavior and PostgreSQL compatibility mode unchanged.

tju-yxq:
## Summary

- add Oracle-compatible character-semantic `SUBSTR` overloads in the implicit `sys` schema
- treat position zero as one and negative positions as offsets from the end
- return `NULL` for nonpositive lengths and empty results
- count multibyte input by characters, while leaving `SUBSTRB` unchanged
- preserve PostgreSQL `pg_catalog.substr` behavior when `ivorysql.compatible_mode = pg`
- add a dedicated Oracle regression test covering integer/numeric arguments, multibyte text, edge cases, and PG-mode controls

Fixes #1850

## Why separate functions are needed

Changing `text_substring()` in `src/backend/utils/adt/varlena.c` would alter PostgreSQL behavior globally. This patch instead defines Oracle-only overloads in `sys`, which IvorySQL implicitly searches before `pg_catalog` in Oracle mode. Exact `text, integer` overloads are provided so ordinary calls do not continue selecting the identically typed `pg_catalog.substr`; numeric overloads preserve Oracle numeric argument coercion.

## Tests performed

Tested from a fresh temporary cluster against upstream master `42ed378168266b3744d165ab496aeebb2b4009df` (September 3, 2026):

1. Built the complete latest `contrib/ivorysql_ora` module with PGXS and compiler warnings enabled.
2. Installed the newly built extension into the local IvorySQL 19devel test installation.
3. Initialized a new cluster and database, then executed the new project SQL test with `ON_ERROR_STOP=1`.
4. Re-ran the original formal reproducer.

Verified Oracle-mode results:

```text
substr('abcdef', 2, 2)       = bc
substr('abcdef', 0, 2)       = ab
substr('abcdef', -2, 2)      = ef
substr('abcdef', -2)         = ef
substr('abcdef', 2, 0)       IS NULL
substr('abcdef', 2, -1)      IS NULL
substr('abcdef', 20)         IS NULL
substr('你好吗', -2, 1)       = 好
substr('你好吗', 0, 2)        = 你好
substr('abcdef', 2.4, 2.4)   = bc
```

Verified controls:

```text
Oracle mode: substrb('abcdef', -2, 2) = ef
PG mode:     substr('abcdef', 0, 2) = a
PG mode:     substr('abcdef', 2, 0) IS NULL = false
PG mode:     negative length still raises substring_error
```

The fresh-cluster test completed without an SQL error. `git diff --check` also passes.

## AI assistance disclosure

Assisted-by: OpenAI Codex:gpt-5

I ran and inspected the build and all test results described above.

## Summary by CodeRabbit

* **New Features**
  * Added Oracle-compatible `SUBSTR` behavior with character-based indexing.
  * Supports positive and negative positions, numeric arguments, omitted lengths, and multibyte characters.
  * Returns `NULL` for zero or negative lengths and out-of-range positions.
  * Preserves byte-based behavior for `SUBSTRB` and PostgreSQL-mode semantics.

* **Tests**
  * Added regression coverage for Oracle `SUBSTR` edge cases, coercion, multibyte strings, and compatibility modes.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -88,6 +88,7 @@ ORA_REGRESS = \
 	dbms_utility \
 	ora_dbms_output \
 	ora_ascii \
+	ora_substr_semantics \
 	ora_stragg \
 	dbms_lock \
 	utl_file \
--- a/contrib/ivorysql_ora/expected/ora_substr_semantics.out
+++ b/contrib/ivorysql_ora/expected/ora_substr_semantics.out
@@ -1,0 +1,88 @@
+--
+-- Oracle-compatible SUBSTR character semantics
+--
+SET ivorysql.compatible_mode = oracle;
+SELECT substr('abcdef', 2, 2) AS normal_positive;
+ normal_positive 
+-----------------
+ bc
+(1 row)
+
+SELECT substr('abcdef', 0, 2) AS zero_position;
+ zero_position 
+---------------
+ ab
+(1 row)
+
+SELECT substr('abcdef', -2, 2) AS negative_position;
+ negative_position 
+-------------------
+ ef
+(1 row)
+
+SELECT substr('abcdef', -2) AS negative_without_length;
+ negative_without_length 
+-------------------------
+ ef
+(1 row)
+
+SELECT substr('abcdef', 2, 0) IS NULL AS zero_length_is_null;
+ zero_length_is_null 
+---------------------
+ t
+(1 row)
+
+SELECT substr('abcdef', 2, -1) IS NULL AS negative_length_is_null;
+ negative_length_is_null 
+-------------------------
+ t
+(1 row)
+
+SELECT substr('abcdef', 20) IS NULL AS past_end_is_null;
+ past_end_is_null 
+------------------
+ t
+(1 row)
+
+-- Positions count characters rather than bytes.
+SELECT substr('你好吗', -2, 1) AS multibyte_negative;
+ multibyte_negative 
+--------------------
+ 好
+(1 row)
+
+SELECT substr('你好吗', 0, 2) AS multibyte_zero;
+ multibyte_zero 
+----------------
+ 你好
+(1 row)
+
+-- Numeric positions and lengths retain the existing Oracle coercion path.
+SELECT substr('abcdef', 2.4::number, 2.4::number) AS numeric_arguments;
+ numeric_arguments 
+-------------------
+ bc
+(1 row)
+
+-- The byte-semantic variant remains unchanged.
+SELECT substrb('abcdef', -2, 2) AS substrb_control;
+ substrb_control 
+-----------------
+ ef
+(1 row)
+
+-- PostgreSQL mode must continue using pg_catalog.substr semantics.
+SET ivorysql.compatible_mode = pg;
+SELECT substr('abcdef', 0, 2) AS pg_zero_position;
+ pg_zero_position 
+------------------
+ a
+(1 row)
+
+SELECT substr('abcdef', 2, 0) IS NULL AS pg_zero_length_is_null;
+ pg_zero_length_is_null 
+------------------------
+ f
+(1 row)
+
+RESET ivorysql.compatible_mode;
--- a/contrib/ivorysql_ora/sql/ora_substr_semantics.sql
+++ b/contrib/ivorysql_ora/sql/ora_substr_semantics.sql
@@ -1,0 +1,30 @@
+--
+-- Oracle-compatible SUBSTR character semantics
+--
+
+SET ivorysql.compatible_mode = oracle;
+
+SELECT substr('abcdef', 2, 2) AS normal_positive;
+SELECT substr('abcdef', 0, 2) AS zero_position;
+SELECT substr('abcdef', -2, 2) AS negative_position;
+SELECT substr('abcdef', -2) AS negative_without_length;
+SELECT substr('abcdef', 2, 0) IS NULL AS zero_length_is_null;
+SELECT substr('abcdef', 2, -1) IS NULL AS negative_length_is_null;
+SELECT substr('abcdef', 20) IS NULL AS past_end_is_null;
+
+-- Positions count characters rather than bytes.
+SELECT substr('你好吗', -2, 1) AS multibyte_negative;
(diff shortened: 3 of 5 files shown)


hanjianqiao:

Muzzammil242 (commented):
Thanks for this - it closes a gap I had measured too. I checked the patch against Oracle Database 21c XE. The main rules match Oracle on the common cases: position 0 as 1, negative positions within the string counting from the end, NULL for a start past the end or a length below 1, multibyte counted by characters. Two edges don't, and both are cheap to fix.

**1. A negative position beyond the start should give NULL, not clamp to position 1.**

`ora_text_substring` does `first = chars + start + 1` and then `if (first < 1) first = 1;`. Oracle returns NULL there:

```sql
SELECT NVL(SUBSTR('abcdef', -7, 2), '<NULL>') FROM dual;   -- <NULL>   patch: 'ab'
SELECT NVL(SUBSTR('abcdef', -7), '<NULL>') FROM dual;      -- <NULL>   patch: 'abcdef'
SELECT SUBSTR('abcdef', -6, 2) FROM dual;                  -- ab       (boundary: |-6| = length, position 1)
```

So `first < 1` should take the same path as `first > chars` (return the empty result, which the wrappers already turn into NULL).

**2. Fractional positions and lengths are truncated by Oracle, but the NUMBER overloads round them.**

The `(varchar, number)` overloads convert with `numeric_int4`, which rounds to nearest. Oracle truncates toward zero:

```sql
SELECT SUBSTR('abcdef', 2.7) FROM dual;        -- bcdef   (2.7 -> 2)   patch: cdef   (rounds to 3)
SELECT SUBSTR('abcdef', -2.7) FROM dual;       -- ef      (-2.7 -> -2) patch: def    (rounds to -3)
SELECT SUBSTR('abcdef', 1, 2.7) FROM dual;     -- ab      (2.7 -> 2)   patch: abc
SELECT SUBSTR('abcdef', 2.3) FROM dual;        -- bcdef   (agrees either way)
```

Truncating before the int conversion fixes all three; `numeric_trunc` with scale 0 followed by `numeric_int4` does it, or a plain cast through `numeric_trunc`.

Neither edge is in `ora_substr_semantics.sql` yet; the rows above would cover them. Happy to re-run any of this against Oracle if it helps.

PR #1853, opened 2026-09-03
Make ASCIISTR database-encoding aware

### LINKED ISSUE #1851: ASCIISTR decodes non-UTF8 database text as UTF-8 bytes
### Summary

`ASCIISTR` / `sys.asciistr()` interprets the raw bytes of a `text` datum as UTF-8 even when the database encoding is not UTF-8. In a `LATIN1` database, valid characters are therefore escaped incorrectly or rejected with `ERROR: Invalid bytes`.

### Environment

Reproduced locally with IvorySQL 5beta1 / PostgreSQL 19devel and `ivorysql_ora 1.0`, using a fresh database created with `ENCODING 'LATIN1'`. The relevant implementation was checked against master commit `c3529ba4789ee220caef858e58bbe6f1563ad14`.

### Reproduction

Create a non-UTF8 database:

```sql
CREATE DATABASE latin1test WITH TEMPLATE template0 ENCODING 'LATIN1'
  LC_COLLATE 'C' LC_CTYPE 'C';
```

Then run:

```sql
CREATE EXTENSION IF NOT EXISTS ivorysql_ora;
SHOW server_encoding;

SELECT sys.asciistr(convert_from(decode('e9','hex'),'LATIN1'));
SELECT sys.asciistr(convert_from(decode('a3','hex'),'LATIN1'));
SELECT sys.asciistr(convert_from(decode('c3a9','hex'),'LATIN1'));
SELECT sys.asciistr(convert_from(decode('e94142','hex'),'LATIN1'));
```

The formal deterministic harness is `C:\Users\26517\Documents\ChatGPT\IvorySQL\.worktrees\issue-research\.research\asciistr-non-utf8\run_asciistr_latin1_formal_repro.sh`; the complete output is in `formal_repro_run.out`.

Observed result:

```text
server_encoding = LATIN1

label                         | input_hex | expected    | actual | error_text    | passed
------------------------------+-----------+-------------+--------+---------------+-------
latin1_a3                     | a3        | \\00A3       | NULL   | Invalid bytes | false
latin1_c3_a9_two_characters   | c3a9      | \\00C3\\00A9  | \\00E9  | NULL          | false
latin1_e9                     | e9        | \\00E9       | \\9F80  | NULL          | false
latin1_e9_followed_by_ascii   | e94142    | \\00E9AB     | \\9042  | NULL          | false

passed_count = 0, failed_count = 4, total_count = 4
```

A UTF8 baseline on the same build remains plausible (`Ä` -> `\\00C4`, `😊` -> `\\D83D\\DE0A`), which isolates the defect to handling of non-UTF8 database encodings.

### Expected behavior

In LATIN1:

- `0xE9` is `é` (U+00E9), so the result should contain `\\00E9`.
- `0xA3` is `£` (U+00A3), so it should contain `\\00A3` rather than error.
- `0xC3 0xA9` are two LATIN1 characters (`Ã` and `©`), so they must not be collapsed as the UTF8 encoding of `é`.

### Code pointer

`contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c`, `ora_asciistr()`. It reads `VARDATA_ANY(str_arg)` and applies UTF8 byte-decoding rules without checking `GetDatabaseEncoding()` or converting the database text to UTF8 first.

### Suggested fix

Make the implementation encoding-aware: convert non-UTF8 database text to UTF8 using PostgreSQL conversion APIs, decode validated UTF8 characters with bounds checks, preserve current UTF8 and backslash/surrogate behavior, and add regression coverage for LATIN1 plus the existing UTF8 cases.

A related out-of-bounds-read fix was previously proposed in closed PR #1622, but this report is about the distinct correctness failure for valid non-UTF8 database text.

tju-yxq:
## Summary

- convert `ASCIISTR` input from the database encoding to validated UTF-8 before decoding code points
- emit one UTF-16 escape for BMP characters and surrogate-pair escapes for supplementary characters
- preserve ASCII and existing backslash behavior
- add a self-contained TAP test that creates both UTF8 and LATIN1 databases and checks byte-specific cases
- register the TAP test in both Make and Meson builds

Fixes #1851

## Root cause

`ora_asciistr()` previously read `VARDATA_ANY(str_arg)` and applied UTF-8 lead-byte rules directly. That works only when the database encoding is UTF8. For example, LATIN1 byte `E9` is one `é` character, but the function treated it as the first byte of a three-byte UTF-8 sequence and consumed unrelated following bytes.

The new implementation uses `pg_do_encoding_conversion(..., GetDatabaseEncoding(), PG_UTF8)`, then decodes the validated UTF-8 result with PostgreSQL's multibyte helpers.

## Tests performed

Tested against upstream master `42ed378168266b3744d165ab496aeebb2b4009df` (September 3, 2026).

### Complete module build

Built and linked the complete latest `contrib/ivorysql_ora` module with the normal warning flags, then installed it into the local IvorySQL 19devel test installation. The build completed successfully.

### Formal fresh-cluster LATIN1 reproducer

The original deterministic reproducer initializes a fresh cluster, creates a fresh LATIN1 database, installs/loads `ivorysql_ora`, executes byte-level cases, and records pass/fail rows. After this patch:

```text
label                         | input_hex | expected    | actual      | passed
------------------------------+-----------+-------------+-------------+-------
latin1_a3                     | a3        | \\00A3       | \\00A3       | true
latin1_c3_a9_two_characters   | c3a9      | \\00C3\\00A9  | \\00C3\\00A9  | true
latin1_e9                     | e9        | \\00E9       | \\00E9       | true
latin1_e9_followed_by_ascii   | e94142    | \\00E9AB     | \\00E9AB     | true

passed_count = 4, failed_count = 0, total_count = 4
```

### Committed TAP test

Ran `contrib/ivorysql_ora/t/001_asciistr_encoding.pl` directly with PostgreSQL's test framework:

```text
ok 1 - ASCIISTR converts LATIN1 characters by character, not as UTF-8 bytes
ok 2 - ASCIISTR preserves BMP and surrogate-pair behavior in UTF8
ok 3 - ASCIISTR preserves the existing backslash escape behavior
1..3
All tests successful.
Result: PASS
```

The TAP test explicitly initializes an Oracle-mode cluster, creates a LATIN1 database from `template0`, and also checks UTF8 BMP/supplementary characters. `git diff --check` passes.

## AI assistance disclosure

Assisted-by: OpenAI Codex:gpt-5

I built the complete module and ran and inspected both the formal reproducer and the committed TAP test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `sys.asciistr` handling for LATIN1 and UTF8 input, ensuring characters are converted individually rather than interpreted as raw UTF8 bytes.
  * Corrected Unicode escape output for basic multilingual characters and supplementary characters represented by surrogate pairs.
  * Preserved existing backslash escaping behavior.
  * Invalid UTF8 sequences after conversion now produce a clear character-encoding error.

* **Tests**
  * Added automated coverage for encoding conversion across LATIN1 and UTF8 databases.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -41,6 +41,7 @@ EXTENSION = ivorysql_ora
 
 # SQL(s) which need to be generated only if which is specified here.
 DATA = ivorysql_ora--1.0.sql
+TAP_TESTS = 1
 
 all: gensql
 
--- a/contrib/ivorysql_ora/meson.build
+++ b/contrib/ivorysql_ora/meson.build
@@ -66,3 +66,14 @@ install_data(
   'preload_ora_misc.sql',
   install_dir: dir_data,
 )
+
+tests += {
+  'name': 'ivorysql_ora',
+  'sd': meson.current_source_dir(),
+  'bd': meson.current_build_dir(),
+  'tap': {
+    'tests': [
+      't/001_asciistr_encoding.pl',
+    ],
+  },
+}
--- a/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c
@@ -1950,60 +1950,68 @@ oracle_instr_4 (PG_FUNCTION_ARGS)
  *   where xxxx represents a UTF-16 code unit. 
  ********************************************************************/
 Datum 
-ora_asciistr(PG_FUNCTION_ARGS) {
-	StringInfoData 	output;
-	text 			*str_arg = NULL;
-	char 			*str = NULL;
-    char			*end = NULL;
-	
+ora_asciistr(PG_FUNCTION_ARGS)
+{
+	StringInfoData output;
+	text	   *str_arg = PG_GETARG_TEXT_PP(0);
+	unsigned char *database_str;
+	unsigned char *utf8_str;
+	unsigned char *current;
+	unsigned char *end;
+	int			database_len = VARSIZE_ANY_EXHDR(str_arg);
+	int			utf8_len;
+	text	   *result;
+
+	database_str = (unsigned char *) VARDATA_ANY(str_arg);
+	utf8_str = pg_do_encoding_conversion(database_str, database_len,
+										  GetDatabaseEncoding(), PG_UTF8);
+	utf8_len = (utf8_str == database_str) ? database_len :
+		strlen((char *) utf8_str);
+
 	initStringInfo(&output);
-	str_arg = PG_GETARG_TEXT_PP(0);
-	str = VARDATA_ANY(str_arg);
-	end = str + VARSIZE_ANY_EXHDR(str_arg);
+	current = utf8_str;
+	end = utf8_str + utf8_len;
+	while (current < end)
+	{
+		char32_t	codepoint;
+		int			char_len;
 
-    while (str < end) {
-        unsigned char c = *str;
-        uint32_t codePoint;
-		
-		if (c == '\\') {
-			/* Handle backslash character */
-			appendUTF16Escape(&output, 0x005C);  // UTF-16 representation of backslash
-			str++;
-        } 
-        else if (c < 0x80) {
-            /* ASCII character */
-            appendStringInfoChar(&output, c);
-            str++;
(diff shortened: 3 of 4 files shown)


hanjianqiao:

PR #1859, opened 2026-09-05
fix: report Oracle NUMBER internal size from vsize() for numeric input

### LINKED ISSUE #1856: vsize() reports PostgreSQL storage width for numeric input instead of Oracle's NUMBER internal format
Hi, while cross-checking `vsize()` against a real Oracle database (21c XE) I noticed that every numeric input returns the PostgreSQL storage width, while Oracle's VSIZE returns the length of the value in its internal NUMBER format.

```
-- IvorySQL, ivorysql.compatible_mode = oracle
SELECT vsize(100);   -- 4   (int4 storage width)
SELECT vsize(3.14);  -- 6   (numeric storage width)
SELECT vsize(-1);    -- 4

-- Oracle Database 21c XE
SELECT VSIZE(100) FROM dual;   -- 2
SELECT VSIZE(3.14) FROM dual;  -- 3
SELECT VSIZE(-1) FROM dual;    -- 3
```

Values verified on the Oracle side:

| value | Oracle VSIZE |
|---|---|
| 0 | 1 |
| 1, 42, 100, 1000000, 1e-10 | 2 |
| -1, -1200, 3.14, 123 | 3 |
| 12345.67 | 5 |
| 38-digit integer | 20 |

Oracle NUMBER stores one exponent byte, then one mantissa byte per two significant decimal digits (trailing zeros are dropped), plus one terminator byte (0x66) for negative values; zero is a single byte. That is why VSIZE(100) is 2 and VSIZE(-1) is 3 regardless of which native type carries the value.

Character input already matches Oracle (byte length), and float4/float8 happen to match BINARY_FLOAT / BINARY_DOUBLE sizes, so only the integer/numeric path differs.

## Suggested fix

For int2/int4/int8/numeric input, compute the NUMBER-format length from the value's significant decimal digits. Happy to submit a small PR with regression tests if this sounds like the right direction.

## Open question

DATE / TIMESTAMP also differ (the 8-byte native width vs Oracle's 7 for DATE and 13 for SYSTIMESTAMP). That looks like a deeper storage-model difference, so I left it out of scope — is matching Oracle there desired, or is the current behavior intended?

unbridled-41:
## Problem / Evidence

In oracle mode, `vsize()` returns the PostgreSQL storage width for every numeric input instead of the length of the value in Oracle's internal NUMBER format:

```sql
-- IvorySQL master (commit 375aa370d72, source build), ivorysql.compatible_mode = oracle
SELECT vsize(100);        -- 4   (int4 storage width)
SELECT vsize(3.14);       -- 6   (numeric payload width)
SELECT vsize(-1);         -- 4
SELECT vsize(0);          -- 4
```

Every numeric input is affected, so migrated applications comparing or persisting `VSIZE()` results silently get different values. Reported in #1856 (values verified there on 21c XE); this PR supplies a fix with a full verification chain.

## Oracle verification

All expected values were verified locally against a real Oracle instance (docker image `gvenzl/oracle-free:23-slim`, banner: `Oracle AI Database 26ai Free Release 23.26.3.0.0`), session via `sqlplus test/test@FREEPDB1`:

```sql
SQL> SELECT VSIZE(100) AS v100, VSIZE(42) AS v42, VSIZE(-1) AS vm1, VSIZE(0) AS v0,
  2         VSIZE(3.14) AS v314, VSIZE(12345.67) AS v12345_67, VSIZE(1e-10) AS v1em10,
  3         VSIZE(1000000) AS v1e6, VSIZE(-1200) AS vm1200, VSIZE(1500) AS v1500,
  4         VSIZE(1.50) AS v150, VSIZE(0.05) AS v005
  5  FROM dual;

      v100        v42        vm1         v0       v314  v12345_67     v1em10      v1e6    vm1200      v1500      v150      v005
---------- ---------- ---------- ---------- ---------- ----------- ---------- ---------- ---------- ---------- ---------- ----------
	 2	    2 	      3	  1 	     3		5	   2	      2 	 3	    2	       3
	 2
```

(that is: VSIZE(100)=2, VSIZE(42)=2, VSIZE(-1)=3, VSIZE(0)=1, VSIZE(3.14)=3, VSIZE(12345.67)=5, VSIZE(1e-10)=2, VSIZE(1000000)=2, VSIZE(-1200)=3, VSIZE(1500)=2, VSIZE(1.50)=3, VSIZE(0.05)=2)

Note the trailing-zero rule this confirms: only **whole all-zero mantissa bytes** are dropped, so VSIZE(100)=2 while VSIZE(1.50)=3 — an initial implementation that dropped single trailing digits got VSIZE(1.50)=2 and was corrected against the real instance before submitting.

## Root cause / Fix

`ora_vsize()` (contrib/ivorysql_ora/src/builtin_functions/misc_functions.c) returned `typlen` for fixed-width types and the payload width for varlena types. For int2/int4/int8/numeric — and `sys.number`, which is binary-coercible with numeric — it now computes Oracle's NUMBER internal size from the value's significant decimal digits:

- one exponent byte,
- one mantissa byte per two significant decimal digits, dropping trailing all-zero mantissa bytes,
- one terminator byte for negative values,
- zero is a single byte.

Type dispatch uses `getBaseType(get_fn_expr_argtype(...))` so domains resolve correctly; `NUMBEROID` (9520) goes through `numeric_out` (binary-coercible cast, confirmed in `pg_cast`: `castmethod = 'b'`). Character data (byte length), other varlena types and fixed-width types (including float4/float8 = BINARY_FLOAT/BINARY_DOUBLE sizes) keep their existing behavior. Non-finite numeric values (NaN) intentionally keep the storage-width fallback since Oracle NUMBER cannot represent them.

## Priority / scoring

PRIORITY 72 = impact 25 (wrong results for a core compat function on all numeric input) + scope 12 (single function) + reproducibility 20 (one-line queries) + maintenance value 15 (Oracle NUMBER format is stable and documented). FIX_CONFIDENCE 85: format rules verified against a real Oracle instance including the trailing-zero-pair edge cases.

## Tests

Regression tests fail on master and pass with the fix:

1. **Pre-fix** (master @ 375aa370d72 code + this PR's test files): `make oracle-check` → `# 1 of 28 tests failed` (`not ok 12 - ora_vsize`). Actual master output from that run:

   ```sql
   SELECT vsize(100);
    vsize
   -------
        4          -- expected 2
   ```

2. **Post-fix**: `make oracle-check ORA_REGRESS=ora_vsize` → `ok 1 - ora_vsize`, and the **full** `make oracle-check` → **`All 28 tests passed`**. No other suite output changed.

New/updated cases in `ora_vsize.sql`: the Oracle-verified values above, int2/int4/int8 dispatch, trailing-zero-pair edge cases (`vsize(1.50::number)`, `vsize(1500)`, `vsize(0.05::number)`), multibyte character data and toast/compressed varlena behavior (previously existing cases kept, expectations corrected from PG storage widths to the Oracle NUMBER format).

## Risk

Low: only `ora_vsize()` behavior changes. Character input, other varlena types and fixed-width types are untouched. DATE/TIMESTAMP inputs still report the native storage width, matching the out-of-scope note in #1856.

Fixes #1856

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected `VSIZE` results for integer, decimal, negative, large, and explicitly cast numeric values to match Oracle-compatible NUMBER sizing.
  - Fixed sizing for decimal boundary cases, domains based on numeric types, and `NaN`, while preserving appropriate behavior for other supported data types.

- **Documentation**
  - Clarified sizing behavior for numeric, variable-length, and fixed-width data types.

- **Tests**
  - Expanded coverage and updated expected results for numeric `VSIZE` calculations.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_vsize.out
+++ b/contrib/ivorysql_ora/expected/ora_vsize.out
@@ -48,40 +48,114 @@ SELECT vsize('你好'::text);
 (1 row)
 
 -- numeric data
+-- integer/numeric values report Oracle's internal NUMBER format size,
+-- verified against Oracle Database 23ai
 SELECT vsize(0::number);
  vsize 
 -------
-     2
+     1
 (1 row)
 
 SELECT vsize(1::number);
  vsize 
 -------
-     4
+     2
 (1 row)
 
 SELECT vsize(123::number);
  vsize 
 -------
-     4
+     3
 (1 row)
 
 SELECT vsize(1.23::number);
  vsize 
 -------
-     6
+     3
+(1 row)
+
+SELECT vsize(100);
+ vsize 
+-------
+     2
+(1 row)
+
+SELECT vsize(42);
+ vsize 
+-------
+     2
+(1 row)
+
+SELECT vsize(-1);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(-1200);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(3.14);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(12345.67);
+ vsize 
+-------
+     5
+(1 row)
+
+SELECT vsize(1000000);
+ vsize 
+-------
+     2
+(1 row)
+
+SELECT vsize(0.001::number);
+ vsize 
+-------
+     2
+(1 row)
+
+SELECT vsize(1.50::number);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(1500);
+ vsize 
+-------
+     2
+(1 row)
+
+SELECT vsize(0.05::number);
+ vsize 
+-------
+     2
+(1 row)
+
+SELECT vsize(123::int2);
+ vsize 
+-------
+     3
 (1 row)
 
 SELECT vsize(123::int4);
  vsize 
 -------
-     4
+     3
 (1 row)
 
 SELECT vsize(123::int8);
  vsize 
 -------
-     8
+     3
 (1 row)
 
 SELECT vsize(1.23::float8);
@@ -142,8 +216,8 @@ INSERT INTO vsize_tbl VALUES ('world', 0);
 SELECT vsize(a), vsize(b) FROM vsize_tbl ORDER BY a NULLS LAST;
  vsize | vsize 
 -------+-------
-     5 |     6
-     5 |     2
+     5 |     4
+     5 |     1
        |      
 (3 rows)
 
--- a/contrib/ivorysql_ora/expected/ora_vsize.out
+++ b/contrib/ivorysql_ora/expected/ora_vsize.out
@@ -140,6 +140,69 @@ SELECT vsize(0.05::number);
      2
 (1 row)
 
+-- base-100 mantissa byte boundaries sit at the decimal point: 1.5 is the
+-- mantissa byte pair 1,50, so an odd significant digit count on either side
+-- of the point adds a mantissa byte (values verified on Oracle 23ai)
+SELECT vsize(1.5);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(1.500);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(1.5::number);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(1.500::number);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(-1.5);
+ vsize 
+-------
+     4
+(1 row)
+
+SELECT vsize(-1.500::number);
+ vsize 
+-------
+     4
+(1 row)
+
+SELECT vsize(12.5);
+ vsize 
+-------
+     3
+(1 row)
+
+SELECT vsize(123.5);
+ vsize 
+-------
+     4
+(1 row)
+
+SELECT vsize(-123.5);
+ vsize 
+-------
+     5
+(1 row)
+
+SELECT vsize(1.005::number);
+ vsize 
+-------
+     4
+(1 row)
+
 SELECT vsize(123::int2);
  vsize 
 -------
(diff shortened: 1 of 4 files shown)


jiaoshuntian:
PR #1859  and PR #1877 are both intended to resolve issue #1856

unbridled-41:
Addressed both review findings in 8197bf8eca8:

**1. Mantissa pairing (major) - confirmed valid.** The concatenated-digit grouping returned 2 for `vsize(1.5)` while `vsize(1.50)` returned 3 for the same value. `ora_number_vsize()` now forms base-100 digit pairs anchored at the decimal point (integer digits grouped right-to-left, fractional digits left-to-right with right zero-padding) and drops only whole all-zero mantissa bytes. All new expected values were re-verified on a real Oracle 23ai Free instance (23.26.3.0.0) before committing: `VSIZE(1.5)=VSIZE(1.500)=3`, `VSIZE(-1.5)=VSIZE(-1.500)=4`, `VSIZE(12.5)=3`, `VSIZE(123.5)=4`, `VSIZE(-123.5)=5`, `VSIZE(1.005)=4` (the last one pins the stored interior zero byte in `1,00,50`).

**2. Domain / non-finite coverage (minor) - added.** Added domains over `number` and over `numeric` (dispatching through `getBaseType()`), plus the non-finite numeric case. One spelling note: a bare `NaN` raises a syntax error under the oracle-mode grammar, which is why this file already used `'NaN'::float8`; the new case is therefore spelled `'NaN'::number` (same value path as `NaN::number`). It reports the storage size (2), since Oracle NUMBER cannot represent NaN.

Verification: the new cases fail without the fix (`vsize(1.5)` returns 2 vs expected 3, `vsize(-123.5)` 4 vs 5, `vsize(1.005)` 3 vs 4, ...). With the fix, `make oracle-check ORA_REGRESS=ora_vsize` passes and the full `make oracle-check` passes all 28 tests.

bigplaice:
@unbridled-41 

Hi, the issue #1856 has been updated as we don't think that's a real issue. Please let me know if you have any different opinions. 

Thanks,
Steven

Issue #1865, opened 2026-09-05
TO_BINARY_FLOAT and TO_BINARY_DOUBLE text overloads are parallel restricted


btlqql:
### IvorySQL Version
Current `master` (`375aa370d7`).

### OS Version (uname -a)
Platform independent catalog metadata issue.

### Configuration options  ( config.status --config )
Default build.

### Current Behavior
The one- and two-argument text overloads of `sys.to_binary_float` and `sys.to_binary_double` are `PARALLEL UNSAFE` by omission. Their implementations only parse arguments and call immutable conversion helpers; the number/binary overloads are already parallel safe.

### Expected behavior/code
All four text conversion overloads should be `PARALLEL SAFE`.

### Step to reproduce
Inspect `pg_proc.proparallel` for the four text overloads; each currently reports `u`.

### Additional context that can be helpful for identifying the problem
This is a catalog declaration mismatch; the C code does not access backend-local mutable state.

jiaoshuntian:
PR #1941 and PR #1869 are both intended to resolve issue #1865.

PR #1870, opened 2026-09-05
fix(oracle): reject datetime zone conversion overflow

### LINKED ISSUE #1862: NEW_TIME and FROM_TZ ignore timestamp conversion overflow
### IvorySQL Version

### OS Version (uname -a)

Default build.

### Current Behavior

### Expected behavior/code

### Step to reproduce

yyqdbngt:
## Summary

`NEW_TIME` and `FROM_TZ` apply a time-zone offset and then call `tm2timestamp()` without checking its result. Near the supported timestamp boundary the offset moves an otherwise valid input out of range, and the uninitialised `result` is returned as if it were a real timestamp.

Both call sites now check the return value and raise `ERRCODE_DATETIME_VALUE_OUT_OF_RANGE`, which is what the other seven `tm2timestamp()` call sites in this file already do.

## Measurement

Same cluster, same session, only the shared library swapped (`ivorysql_ora.so` md5 `a6741ce3` vs `155ad8b4`), Oracle mode:

```
select from_tz(timestamp '294276-12-31 23:00:00','-12:00') from dual;
  before: 2000-01-01 08:00:00.000000 +08:00     <-- silently wrong, no error
  after : ERROR:  timestamp out of range
```

So the failure mode is not "raises the wrong error" but "returns a bogus timestamp".

## Testing

Rebased onto `master` (`03b24b1c46`) and run on Ubuntu 22.04 with that branch applied:

```
ok: 28   not ok: 1
not ok 17    - ora_xml_functions        # pre-existing: this tree is built without libxml
ok    10     - ora_datetime_datatype_functions
```

`results/ora_datetime_datatype_functions.out` is byte-identical to the expected file after the run.

## Note on the regression coverage

An earlier revision of this PR also asserted the same boundary for `NEW_TIME`, with hand-written expected output. That case does not work and has been removed:

- `NEW_TIME` takes `sys.oradate`. In the Oracle regression harness a six-digit year is not accepted by the date input path — `select date '294276-12-31' from dual;` reports `ERROR: unmatched format separator "-"`, and `to_date('294276-12-31','YYYY-MM-DD')` reports `ERROR: The value of month is invalid.`
- The hand-written expectation claimed `ERROR: timestamp out of range`, which was never what the harness produced.

The `NEW_TIME` guard is kept: it matches the surrounding code, and a `sys.oradate` carrying a time component can reach the function without going through the sql text parser.

Closes #1862

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for Oracle-compatible `new_time` and `FROM_TZ` functions.
  - Out-of-range timestamp values now return a clear `timestamp out of range` error instead of being processed incorrectly.

- **Tests**
  - Added regression coverage confirming that `FROM_TZ` rejects timestamps at the upper supported range boundary.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
@@ -579,6 +579,8 @@ SELECT FROM_TZ(TIMESTAMP '2000-03-28 08:00:00', '3:00') FROM DUAL;
  03-27-2000 21:00:00.000000000 -08:00
 (1 row)
 
+SELECT FROM_TZ(TIMESTAMP '294276-12-31 23:00:00', '-12:00') FROM DUAL;
+ERROR:  timestamp out of range
   
 --sessiontimezone
 set timezone = 'Asia/Hong_Kong';
--- a/contrib/ivorysql_ora/sql/ora_datetime_datatype_functions.sql
+++ b/contrib/ivorysql_ora/sql/ora_datetime_datatype_functions.sql
@@ -328,6 +328,7 @@ drop table datetest;
 
 --from_tz
 SELECT FROM_TZ(TIMESTAMP '2000-03-28 08:00:00', '3:00') FROM DUAL;
+SELECT FROM_TZ(TIMESTAMP '294276-12-31 23:00:00', '-12:00') FROM DUAL;
   
 --sessiontimezone
 set timezone = 'Asia/Hong_Kong';
--- a/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
@@ -631,7 +631,10 @@ ora_new_time(PG_FUNCTION_ARGS)
 				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
 				 errmsg("timestamp out of range")));
 	tz1 = tz1 - tz2;
-	tm2timestamp(tm, fsec, &tz1, &result);
+	if (tm2timestamp(tm, fsec, &tz1, &result) != 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+				 errmsg("timestamp out of range")));
 
 	PG_RETURN_TIMESTAMP(result);
 }
@@ -769,7 +772,10 @@ ora_from_tz(PG_FUNCTION_ARGS)
 				 errmsg("timestamp out of range")));
 
 	tz = DetermineTimeZoneOffset(tm, timezonedat);
-	tm2timestamp(tm, fsec, &tz, &result);
+	if (tm2timestamp(tm, fsec, &tz, &result) != 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+				 errmsg("timestamp out of range")));
 
 	PG_RETURN_TIMESTAMPTZ(result);
 }


jiaoshuntian:
PR #1941 and PR #1869 are both intended to resolve issue #1862.

yyqdbngt:
A note on the duplicate check, because I think the PR numbers there point elsewhere: #1941 and #1869 are both about the parallel-safety of the `TO_BINARY_FLOAT` / `TO_BINARY_DOUBLE` text overloads, i.e. issue #1865. This PR is about #1862 (NEW_TIME / FROM_TZ ignoring the result of the timestamp conversion), and no other open PR touches `datetime_datatype_functions.c`.

It is still needed on current `master` (`03b24b1c46`). In that file, 9 calls to `tm2timestamp()` ignore or check the return value as follows:

```
$ git grep -n 'tm2timestamp' master -- contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c | grep -c 'if (tm2timestamp'
7
$ git grep -n 'tm2timestamp(' master -- contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c | grep -v 'if (tm2timestamp'
   634:	tm2timestamp(tm, fsec, &tz1, &result);     <- ora_new_time
   772:	tm2timestamp(tm, fsec, &tz, &result);      <- ora_from_tz
```

Seven of the nine callers in this file already treat a non-zero return as `timestamp out of range`; the two that do not are exactly the two this PR fixes, and they are the two reachable from Oracle's `NEW_TIME` and `FROM_TZ`. The overflow therefore silently produces whatever `result` happened to hold instead of raising, which is what #1862 reports.

yyqdbngt:
Rebased onto `master` (`03b24b1c46`) and re-verified on a real build, because the earlier revision had a problem I only found by running it:

**The `NEW_TIME` regression case was wrong.** Its expected output had been written by hand. Running `make -C contrib/ivorysql_ora oracle-check` on the branch showed the module failing, because a six-digit year is not usable in the Oracle regression harness:

```
select date '294276-12-31' as p1 from dual;
ERROR:  unmatched format separator "-"
select to_date('294276-12-31') as p3 from dual;
ERROR:  The value of month is invalid.
select new_time(date '294276-12-31','GMT','PST') as p7 from dual;
ERROR:  unmatched format separator "-"
```

(`nls_date_format` in that session is `YYYY-MM-DD HH24:MI:SS`.) So that case never exercised the guard, and its `ERROR: timestamp out of range` expectation was fiction. I have removed it rather than regenerating the file from that run, since regenerating would have made CI green while asserting an unrelated parse error.

The `FROM_TZ` case is real and is kept:

```
ok: 28   not ok: 1
not ok 17    - ora_xml_functions        # pre-existing, no libxml in this tree
ok    10     - ora_datetime_datatype_functions
$ diff expected/... results/...   # identical
```

And it fails without the fix — same cluster, only `ivorysql_ora.so` swapped (`a6741ce3` vs `155ad8b4`):

```
select from_tz(timestamp '294276-12-31 23:00:00','-12:00') from dual;
  before: 2000-01-01 08:00:00.000000 +08:00
  after : ERROR:  timestamp out of range
```

One more thing, since it is easy to miss: `main` has **9** `tm2timestamp()` calls in this file and **7** of them already check the return value. The two that do not are exactly the two this PR fixes, so the change follows the file's own convention rather than introducing a new one.

muzimu217:
Hi @yyqdbngt, thanks for this fix — I reproduced the failure mode locally on an
unpatched build and it matches what you describe: not a wrong error, but a bogus
timestamp returned as if it were real. I also went through the patch looking for
gaps; notes below, all in the spirit of double-checking rather than objecting.

## 1. Bug reproduced independently (unpatched IvorySQL 5beta1, PostgreSQL 19beta1, oracle mode)

```sql
SELECT from_tz(timestamp '294276-12-31 23:00:00','-12:00');
```

On my machine this silently returns:

```
        from_tz
-----------------------
 1999-12-31 16:00:00.000000 -08:00
(1 row)
```

Five runs across two fresh sessions all returned that same value on my box, which
differs from the `2000-01-01 08:00:00.000000 +08:00` in your PR body — consistent
with the uninitialized `result` stack slot being filled by whatever the preceding
call path left there, so the exact garbage is environment-dependent while the
failure mode ("plausible-looking timestamp, no error") is stable. This makes the
bug nastier than an exception-path bug in my opinion, because downstream code
consumes a fabricated value silently.

## 2. Completeness: these look like the only two unchecked call sites

I grepped every `tm2timestamp(` call under `contrib/ivorysql_ora/src/` at
master `03b24b1c46`. Out of 44 call sites, only the two this patch touches
(`ora_new_time`, line 634, and `ora_from_tz`, line 772, both in
`builtin_functions/datetime_datatype_functions.c`) ignore the return value;
all the others already `!= 0`-check it. So as far as I can tell the patch closes
the entire class within ivorysql_ora, and the error choice
(`ERRCODE_DATETIME_VALUE_OUT_OF_RANGE` + "timestamp out of range") matches the
wording used by the neighboring call sites in the same file.

One small description nit, entirely non-blocking: the PR body says "the other
seven tm2timestamp() call sites in this file already do" — I count fourteen
checked call sites in that file (247, 276, 333, 466, 506, 834, 856, 879, 904,
925, 947, 983, 1025, 1058). The direction of the claim is right; only the count
in the write-up seems off, so you may want to correct it for future readers.

## 3. The dropped NEW_TIME regression case: I also could not reach it

Your note says the earlier NEW_TIME case was removed because a six-digit year
cannot get through the date input path, and I confirmed that from both doors in
oracle mode:

```sql
SELECT to_date('294276-12-31','YYYY-MM-DD');  -- ERROR: The value of month is invalid.
SELECT date '294276-12-31';                   -- ERROR: unmatched format separator "-"
```

Additionally, `sys.new_time` is declared `new_time(date, text, text) -> date`,
so its input is `sys.oradate` with a hard ceiling of AD 9999. Even with the
largest zone-pair offset difference NEW_TIME supports, the adjusted result lands
around AD 10000 — roughly 284k years below where `tm2timestamp` starts failing
(year 294277). I could not construct any NEW_TIME input that reaches the new
error branch. I read the `ora_new_time` hunk as purely defensive hardening
(probably still worth keeping for symmetry), but if that matches your
understanding too, it may be worth a one-line comment in the code or PR saying
the branch is believed unreachable through the date surface — that would save
the next reader the same hunt for a counterexample.

## 4. Regression case

The single new FROM_TZ fixture pins the error, which I think is the right call
given the garbage value is stack-content-dependent. On my session the literal
`timestamp '294276-12-31 23:00:00'` parses fine in oracle mode (my unpatched run
executed into the C function and returned garbage rather than a parse error), so
after the patch the raised "timestamp out of range" demonstrably comes from the
new check inside `ora_from_tz`, not from the literal parser.

## Verification environment (for the record)

- IvorySQL 5beta1 (PostgreSQL 19beta1), aarch64-apple-darwin24.6.0, local build
  at master `03b24b1c46` + your patch **not** applied (repro-only; I did not run
  the patched branch — your CI run with the byte-identical .out covers that side).
- `America/Los_Angeles` session, oracle compatible mode, scratch database.

Nothing above is meant as a blocking concern — items 2 and 3 are the only
follow-ups I'd suggest, both cosmetic/documentation. Nice find.

yyqdbngt:
Thank you for the independent reproduction, and for the two follow-ups - both are settled.

**The count (item 2).** You are right and the description was wrong. Re-counted on the patched file: `tm2timestamp(` appears at 16 lines; 14 of them pre-date this change and already check the result (247, 276, 333, 466, 506, 840, 862, 885, 910, 931, 953, 989, 1031, 1064 - your numbers plus the 6 lines this patch inserts above them), and 2 are the checks added here (634 in `ora_new_time`, 775 in `ora_from_tz`). So the claim the sentence was making holds - these two were the last unchecked call sites in the module - but the number was seven, and it now reads fourteen with the line list, plus a sentence saying the earlier figure was wrong.

**The unreachable NEW_TIME branch (item 3).** I could not reach it either, from the same two doors you tried, and your arithmetic is what convinced me: `sys.new_time` takes a date (ceiling AD 9999), the largest zone-pair difference the function supports puts the result around AD 10000, and `tm2timestamp()` does not start failing until year 294277. The check stays, because what it prevents is the failure this pull request exists to fix - an uninitialised `result` returned as a timestamp - but you are right that a reader deserves to be told it is defensive, so `c0e28d8e` adds a comment at the call site saying the branch is believed unreachable through the date surface and why. That is the only change I made; the production diff and the regression expected file are untouched.

**On the garbage value (item 1).** Your `1999-12-31 16:00:00.000000 -08:00` and my `2000-01-01 08:00:00.000000 +08:00` are the same defect seen through different stack contents, and I have edited nothing to make the two look alike: the regression case pins the error, not the fabricated value, precisely because the fabricated value is environment-dependent. Your point that this is nastier than an exception-path bug is the reason the fix is at the call site rather than in the parser.

The verification workflow that produced the red/green runs is not part of this branch; the branch carries only the production change, the SQL case, the expected output and now that one comment.

PR #1877, opened 2026-09-05
Answer vsize() for numeric input in Oracle NUMBER format

muzimu217:
Thanks @jiaoshuntian for the pointer — and sorry for the overlap; I hadn't seen #1859 when I opened this.

After reading it, I agree #1859 is the more complete fix: it derives the Oracle NUMBER internal size from the decimal rendering, so it also covers the paths this PR missed — explicit `::number` casts and NUMBER columns report PG varlena widths with this patch (e.g. `vsize(12345.67::number)` returns 8 vs Oracle's 5; a NUMBER column holding 12345 returns 6 vs 4).

Withdrawing this PR in favor of #1859. One input that may be useful for its review: I verified the values against Oracle 23ai before opening this PR, and they all agree with the algorithm in #1859, including the decimal-point anchoring added in 8197bf8:

- `vsize(0)` = 1
- `vsize(1)` = 2
- `vsize(123)` = 3
- `vsize(1.23)` = 3
- `vsize(12345.67)` = 5
- `vsize(NUMBER column = 12345)` = 4
- `vsize(cast(12345.67 as number))` = 5 (same as the literal)

The CAST/column rows might be worth adding as regression cases in #1859, since the current test file only covers literals.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_vsize.out
+++ b/contrib/ivorysql_ora/expected/ora_vsize.out
@@ -47,7 +47,9 @@ SELECT vsize('你好'::text);
      6
 (1 row)
 
--- numeric data
+-- numeric data: byte length in Oracle's internal NUMBER format
+-- (one exponent byte, one mantissa byte per two significant decimal
+-- digits, plus a terminator byte for negative values; zero is one byte)
 SELECT vsize(0::number);
  vsize 
 -------
@@ -75,13 +77,13 @@ SELECT vsize(1.23::number);
 SELECT vsize(123::int4);
  vsize 
 -------
-     4
+     3
 (1 row)
 
 SELECT vsize(123::int8);
  vsize 
 -------
-     8
+     3
 (1 row)
 
 SELECT vsize(1.23::float8);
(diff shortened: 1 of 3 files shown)


jiaoshuntian:
PR #1859  and PR #1877 are both intended to resolve issue #1856

muzimu217:
Thanks @jiaoshuntian for the pointer — and sorry for the overlap; I hadn't seen #1859 when I opened this.

After reading it, I agree #1859 is the more complete fix: it derives the Oracle NUMBER internal size from the decimal rendering, so it also covers the paths this PR missed — explicit `::number` casts and NUMBER columns report PG varlena widths with this patch (e.g. `vsize(12345.67::number)` returns 8 vs Oracle's 5; a NUMBER column holding 12345 returns 6 vs 4).

Withdrawing this PR in favor of #1859. One input that may be useful for its review: I verified the values against Oracle 23ai before opening this PR, and they all agree with the algorithm in #1859, including the decimal-point anchoring added in 8197bf8:

- `vsize(0)` = 1
- `vsize(1)` = 2
- `vsize(123)` = 3
- `vsize(1.23)` = 3
- `vsize(12345.67)` = 5
- `vsize(NUMBER column = 12345)` = 4
- `vsize(cast(12345.67 as number))` = 5 (same as the literal)

The CAST/column rows might be worth adding as regression cases in #1859, since the current test file only covers literals.

PR #1884, opened 2026-09-06
parser: remove dead raw_parser NULL guard

920032148:
This PR removes the unreachable NULL guard in `raw_parser()`.

`sql_raw_parser` is initialized to a valid parser and is not reset in normal flow, so the branch cannot be hit. Keeping only the direct call makes the code simpler and avoids a misleading error path.

No functional tests were run; this is a code-quality cleanup only.

## Summary by CodeRabbit

* **Behavior Changes**
  * Removed the controlled error handling for an unavailable SQL parser hook.


jiaoshuntian:

PR #1885, opened 2026-09-06
Optimize ora_regexp_count: drop text_to_cstring copy

920032148:
This PR removes the unnecessary `text_to_cstring(match_param)` copy in `ora_regexp_count()`.

The function already has the text datum and can inspect the first byte directly with `VARDATA_ANY(match_param)`, so we avoid one allocation and memcpy on every call.

No functional tests were run; this is a performance cleanup only.

## Summary by CodeRabbit

* **Performance**
  * Improved internal handling of regular expression match parameters to avoid unnecessary temporary allocations.
* **Compatibility**
  * Preserved existing validation and regular expression flag parsing behavior.


jiaoshuntian:

PR #1887, opened 2026-09-06
feat: add Oracle-compatible EMPTY_CLOB/EMPTY_BLOB constructors

### LINKED ISSUE #1886: oracle compatibility feature: EMPTY_CLOB/EMPTY_BLOB
## Feature Request

Oracle-compatible LOB constructors `EMPTY_CLOB()` and `EMPTY_BLOB()` for oracle mode.

**Why it is needed:**

IvorySQL implements `clob`, `nclob` and `blob` as Oracle-compatible types, but in oracle mode the empty string is NULL (matching Oracle). That means there is currently no way to initialize a LOB column with a zero-length, non-NULL value, which is a common pattern in Oracle applications and migration scripts:

```sql
INSERT INTO files (id, content, payload) VALUES (1, EMPTY_CLOB(), EMPTY_BLOB());
UPDATE files SET content = EMPTY_CLOB() WHERE id = 1;
```

Oracle semantics (verified on Oracle 23ai Free, gvenzl/oracle-free:23-slim):

```
SQL> SELECT CASE WHEN empty_clob() IS NULL THEN 'NULL' ELSE 'NOTNULL' END FROM dual;
NOTNULL
SQL> SELECT DBMS_LOB.GETLENGTH(empty_clob()) FROM dual;
         0
SQL> CREATE TABLE lob_p (id INT, c CLOB, nc NCLOB, b BLOB);
SQL> INSERT INTO lob_p VALUES (1, empty_clob(), empty_clob(), empty_blob());
SQL> SELECT CASE WHEN c IS NULL THEN 'NULL' ELSE 'NOTNULL' END,
  2         CASE WHEN nc IS NULL THEN 'NULL' ELSE 'NOTNULL' END,
  3         CASE WHEN b IS NULL THEN 'NULL' ELSE 'NOTNULL' END FROM lob_p;
NOTNULL  NOTNULL  NOTNULL
SQL> INSERT INTO lob_p VALUES (2, '', '', '');
SQL> SELECT CASE WHEN c IS NULL THEN 'NULL' ELSE 'NOTNULL' END FROM lob_p WHERE id=2;
NULL
```

**External implementations:**

- Oracle Database SQL Language Reference: EMPTY_CLOB / EMPTY_BLOB
  https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/EMPTY_BLOB.html
- EDB Postgres Advanced Server provides EMPTY_CLOB()/EMPTY_BLOB() and documents them in its LOB support:
  https://www.enterprisedb.com/docs/epas/latest/reference/oracle_compatibility_reference/epas_compat_bip_guide/03_built-in_packages/06_dbms_lob/
- openGauss also supports empty_blob()/empty_clob() as part of its Oracle LOB compatibility (its LOB docs and migration tooling rely on them).

**Scope:**

Small, well-bounded addition: two zero-argument built-in functions `sys.empty_clob()` / `sys.empty_blob()` returning a non-NULL zero-length LOB value, plus regression tests.

**Acceptance criteria** (all verified against Oracle 23ai Free, gvenzl/oracle-free:23-slim; covered by the PR's regression tests):

- `empty_clob()` / `empty_blob()` are non-NULL (`IS NULL` → false) with length 0
- usable for `nclob` targets as well; concatenation yields a non-NULL, non-empty value
- stored into clob/nclob/blob columns the values remain NOT NULL with length 0, distinct from `''`, which oracle mode stores as NULL
- assigning the constructor through UPDATE clears a LOB to empty (not NULL)

**Corresponding PR:** #1887

unbridled-41:
## Source

- Oracle Database 23ai SQL Language Reference, [EMPTY_BLOB, EMPTY_CLOB](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/EMPTY_BLOB-EMPTY_CLOB.html) — the functions this PR implements; the page explicitly states "An empty LOB is not the same as a null LOB".
- EDB Postgres Advanced Server (mature Oracle-compatible PostgreSQL product) ships both functions: [DBMS_LOB / LOB support](https://www.enterprisedb.com/docs/epas/latest/reference/oracle_compatibility_reference/epas_compat_bip_guide/03_built-in_packages/06_dbms_lob/).
- openGauss: its official migration tool [ora2og explicitly maps EMPTY_CLOB()/EMPTY_BLOB()](https://gitee.com/openGauss/openGauss-tools-ora2og/blob/master/changelog) and the [openGauss-server](https://gitee.com/openGauss/openGauss-server) kernel implements them in its Oracle ("A") compatibility mode.
- Feature request filed for this PR: #1886.

## Current gap

`grep -ri "empty_clob\|empty_blob" src/ contrib/` on master (375aa370d72) returns nothing: the two functions do not exist anywhere in the tree. They are not an intentional omission — IvorySQL already ships the `clob`/`nclob`/`blob` types (pg_type.dat, domains over text/bytea) precisely to support Oracle LOB usage, and oracle mode deliberately treats the empty string as NULL (matching Oracle), which makes the constructors the *only* way to initialize a zero-length, non-NULL LOB. No open PR addressed this before this one.

## Project fit

IvorySQL's stated goal is PostgreSQL with Oracle compatibility. The LOB types already exist as first-class types in the `sys` schema, and the project's established pattern for Oracle built-ins is small C functions registered in `sys` (cf. `sys.vsize`, `sys.uid`, `sys.asciistr` in the same `builtin_functions` layer this PR extends). The constructors complete the LOB story that the types started, without touching the core.

## Scope

Two zero-argument functions, `sys.empty_clob()` and `sys.empty_blob()`:

- return a zero-length, **non-NULL** LOB value (`IS NULL` → false; `length()`/`octet_length()` → 0);
- can be stored in `clob`/`nclob`/`blob` columns, after which `IS NULL`/`IS NOT NULL` treat the row as holding an empty LOB — distinct from `''`, which oracle mode stores as NULL;
- work through UPDATE assignment and concatenation (`empty_clob() || 'abc'` → `'abc'`).

Not included: nothing else changes. Oracle's `EMPTY_CLOB()` returns a mutable locator that cannot be passed to DBMS_LOB ("The returned locator cannot be used as a parameter to the DBMS_LOB package" — Oracle doc, same page); IvorySQL's value model has no locators, so there is no locator API to exclude.

## Implementation

- `contrib/ivorysql_ora/src/builtin_functions/misc_functions.c`: `ora_empty_clob` returns a zero-length text varlena (`cstring_to_text("")`), `ora_empty_blob` returns a zero-length bytea varlena. Both are pure value constructors.
- `contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql`: register `sys.empty_clob()` RETURNS `sys.clob` and `sys.empty_blob()` RETURNS `sys.blob` (the unconstrained domains over text/bytea supply the Oracle type names; both are IMMUTABLE/PARALLEL SAFE, matching the `sys.vsize` registration style).

## Tests

All commands run in the WSL build tree (`~/build/src`), Oracle semantics verified against Oracle 23ai Free (gvenzl/oracle-free:23-slim container, sqlplus session recorded in the PR and in issue #1886):

- New regression test `contrib/ivorysql_ora/sql/ora_empty_lob.sql` (+ expected): constructor non-NULLness, zero lengths, `nclob` usage, concatenation, storage into clob/nclob/blob columns, `IS NULL`/`IS NOT NULL` filtering, distinction from the NULL empty string, update-through-constructor.
- Oracle 23ai reference values used: `EMPTY_CLOB() IS NULL` → FALSE; `DBMS_LOB.GETLENGTH(empty_clob())` → 0; stored constructor values in CLOB/NCLOB/BLOB columns → NOT NULL with length 0; `''` → NULL; `LENGTH(empty_clob() || 'abc')` → 3.
- `make oracle-check ORA_REGRESS=ora_empty_lob` → **1/1 pass**.
- `make oracle-check` (full suite) → **29/29 pass** (28 pre-existing + 1 new).

## Compatibility / Risk

- Additive only: two new `sys`-schema functions in oracle mode; no existing function, parser, or core behavior is modified.
- Value-vs-locator difference is documented: Oracle returns a locator (per the doc page above, unusable with DBMS_LOB anyway); IvorySQL returns a value. Observable SQL semantics (non-NULL, length 0, distinct from `''`) match Oracle.
- No new dependencies; no generated/vendor code touched; license-compatible (original implementation).
- psql displays `bytea` as hex (`\x`), so raw-display parity with SQL*Plus is not asserted in tests; `octet_length`/`IS NULL` are used instead.

## Issue

Closes #1886.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -80,6 +80,7 @@ ORA_REGRESS = \
 	ora_datetime_datatype_functions \
 	ora_misc_functions \
 	ora_vsize \
+	ora_empty_lob \
 	ora_merge \
 	datatype_and_func_bugs \
 	ora_sysview \
--- a/contrib/ivorysql_ora/expected/ora_empty_lob.out
+++ b/contrib/ivorysql_ora/expected/ora_empty_lob.out
@@ -1,0 +1,95 @@
+--
+-- EMPTY_CLOB / EMPTY_BLOB
+--
+-- Oracle-compatible LOB constructors: return a zero-length, non-NULL LOB
+-- value.  Oracle treats the empty string as NULL, so these constructors are
+-- the only way to initialize a LOB that is not NULL.
+--
+-- Semantics verified against Oracle 23ai Free.
+--
+-- constructors return non-NULL zero-length values
+SELECT empty_clob() IS NULL AS clob_is_null;
+ clob_is_null 
+--------------
+ f
+(1 row)
+
+SELECT empty_blob() IS NULL AS blob_is_null;
+ blob_is_null 
+--------------
+ f
+(1 row)
+
+SELECT length(empty_clob()) AS clob_len;
+ clob_len 
+----------
+        0
+(1 row)
+
+SELECT octet_length(empty_blob()) AS blob_len;
+ blob_len 
+----------
+        0
+(1 row)
+
+-- usable for nclob targets as well
+SELECT length(empty_clob()::nclob) AS nclob_len;
+ nclob_len 
+-----------
+         0
+(1 row)
+
+-- concatenation yields a non-NULL, non-empty value
+SELECT '[' || empty_clob() || ']' AS bracketed;
+ bracketed 
+-----------
+ []
+(1 row)
+
+SELECT length(empty_clob() || 'abc') AS concat_len;
+ concat_len 
+------------
+          3
+(1 row)
+
+-- stored values are not NULL and compare accordingly
+CREATE TABLE lob_empty_test (id int, c clob, nc nclob, b blob);
+INSERT INTO lob_empty_test VALUES (1, empty_clob(), empty_clob(), empty_blob());
+SELECT id, c IS NULL AS c_is_null, nc IS NULL AS nc_is_null, b IS NULL AS b_is_null
+FROM lob_empty_test WHERE id = 1;
+ id | c_is_null | nc_is_null | b_is_null 
+----+-----------+------------+-----------
+  1 | f         | f          | f
+(1 row)
+
+SELECT id, length(c) AS c_len, length(nc) AS nc_len, octet_length(b) AS b_len
+FROM lob_empty_test WHERE id = 1;
+ id | c_len | nc_len | b_len 
+----+-------+--------+-------
+  1 |     0 |      0 |     0
+(1 row)
+
+SELECT count(*) AS not_null_rows FROM lob_empty_test WHERE c IS NOT NULL;
+ not_null_rows 
+---------------
+             1
+(1 row)
+
+-- the plain empty string is NULL, so it does not produce an empty LOB
+INSERT INTO lob_empty_test VALUES (2, '', '', '');
+SELECT id, c IS NULL AS c_is_null FROM lob_empty_test WHERE id = 2;
+ id | c_is_null 
+----+-----------
+  2 | t
+(1 row)
+
+-- assigning the constructor clears the LOB to empty (not NULL)
+UPDATE lob_empty_test SET c = empty_clob() WHERE id = 2;
+SELECT id, c IS NULL AS c_is_null, length(c) AS c_len
+FROM lob_empty_test WHERE id = 2;
+ id | c_is_null | c_len 
+----+-----------+-------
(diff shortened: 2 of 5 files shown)


jiaoshuntian:
thanks for your PR

PR #1890, opened 2026-09-06
feat: add Oracle-compatible DBMS_LOB package

### LINKED ISSUE #1029: oracle compatibility feature: DBMS_LOB
## Feature Request
oracle compatibility feature: DBMS_LOB
DBMS_LOB is used to manipulate Large Object (LOB) data types.

unbridled-41:
## Source

- Official feature request [#1029](https://github.com/IvorySQL/IvorySQL/issues/1029) ("oracle compatibility feature: DBMS_LOB", open since 2025-12-05).
- Oracle Database 23ai PL/SQL Packages and Types Reference, [DBMS_LOB](https://docs.oracle.com/en/database/oracle/oracle-database/23/arpls/DBMS_LOB.html) — the package this PR implements.
- EDB Postgres Advanced Server ships DBMS_LOB: [EDB DBMS_LOB reference](https://www.enterprisedb.com/docs/epas/latest/reference/oracle_compatibility_reference/epas_compat_bip_guide/03_built-in_packages/06_dbms_lob/).

## Current gap

On master (375aa370d72) IvorySQL ships the Oracle LOB types (`clob`, `nclob`, `blob` — domains over `text`/`bytea` in pg_type.dat) but **no `dbms_lob` package anywhere** (`grep -ri "dbms_lob" src/ contrib/` finds nothing), and no open PR addresses #1029 (keyword re-checked when this PR was prepared). Without it, LOB manipulation has no Oracle-compatible API, which blocks LOB-using migrations — the package is among Oracle's most-used programming interfaces.

## Project fit

The package follows IvorySQL's established built-in-package architecture exactly (the `builtin_packages/<name>/` pattern of `dbms_lock`, `dbms_session`, etc.): a `CREATE PACKAGE`/`CREATE PACKAGE BODY` pair registered through `ivorysql_ora_merge_sqls`, exposed in the `sys` schema for oracle mode. IvorySQL's own decision to model `clob`/`blob` as domains over text/bytea (value semantics) means a value-level DBMS_LOB subset is the natural, architecture-consistent first cut — no locators, no shared-memory state, no core changes.

## Scope

Implemented (value-level subprograms on clob/nclob/blob + the `lobmaxsize` constant, 18446744073709551615, matching the Oracle 23ai value):

- `GETLENGTH(lob)` — characters for CLOB/NCLOB, bytes for BLOB; NULL in → NULL out.
- `SUBSTR(lob, amount := 32767, offset := 1)` — amount-first argument order; offset/amount < 1, offset > length, or **amount > 32767 → NULL**; amount clamped to the LOB end.
- `INSTR(lob, pattern, offset := 1, occurrence := 1)` — offset/occurrence < 1 → NULL; no match (incl. pattern longer than LOB, offset > length) → 0.
- `COMPARE(lob_1, lob_2, amount := lobmaxsize, offset_1 := 1, offset_2 := 1)` — strcmp-style **-1/0/+1** over the first min(amount, remaining_1, remaining_2) characters; NULL for NULL/invalid arguments.
- `APPEND(dest IN OUT, src)`, `COPY(dest IN OUT, src, amount, dest_offset, src_offset)` (overwrites from dest_offset, preserves the rest; gap to a dest_offset past the end is padded with spaces/zeros as Oracle does), `TRIM(dest IN OUT, newlen)` (raises for NULL/negative or over-length, ORA-21560/ORA-22926 text), `ERASE(dest IN OUT, amount IN OUT, offset)` (spaces for CLOB / zero bytes for BLOB; amount reports the erased count).

Not included (documented in the package header): locator-oriented subprograms with no equivalent in IvorySQL's value model — OPEN/CLOSE/ISOPEN, CREATETEMPORARY/FREETEMPORARY/ISTEMPORARY, the FILE* family (no `bfile` type), CONVERTTOBLOB/CONVERTTOCLOB, READ/WRITE, FRAGMENT_*. These are follow-up material, not silently dropped.

## Implementation

- `contrib/ivorysql_ora/src/builtin_packages/dbms_lob/dbms_lob--1.0.sql`: package spec + PL/iSQL body (zero C code), following the `dbms_lock` delegation-style house pattern but implementing the logic directly in the body.
- Registered in `contrib/ivorysql_ora/ivorysql_ora_merge_sqls` (extension SQL merge) and `contrib/ivorysql_ora/Makefile` (`ORA_REGRESS`); no `OBJS` change since there is no C file.
- The body calls `pg_catalog` primitives explicitly (never its own member names) so unqualified member resolution cannot recurse, and truncates its NUMBER position/amount arguments toward zero (verified Oracle behavior) before use.
- Overloads: `getlength` has CLOB/NCLOB/BLOB forms; calling it with an untyped NULL is ambiguous — exactly as Oracle's static resolver rejects `DBMS_LOB.GETLENGTH(NULL)` with PLS-00307.

## Tests

All commands run in the WSL build tree (`~/build/src`); every expected value was first captured from Oracle 23ai Free (gvenzl/oracle-free:23-slim, sqlplus sessions), including a dedicated boundary probe round:

- Oracle boundary facts the tests encode: SUBSTR amount 32768 → NULL while 32767 works; fractional arguments truncate toward zero (`SUBSTR('abcde', 2.7, 2.7)` = `'bc'`); COMPARE is sign-based (`('abcde','abcXe',5)` = 1, prefix = -1, pos-beyond-one = -1, pos-beyond-both = 0); COPY pads the gap with spaces/zeros (`COPY(d,'xy',2,100,1)` on 'ABCDEF' → length 101 ending 'xy'; blob gap → `AABB0000DD`); ERASE clamps the IN OUT amount (10 → 3) but leaves it untouched when pos is beyond the end; TRIM with negative newlen raises ORA-21560 text.
- New regression test `contrib/ivorysql_ora/sql/dbms_lob.sql` (+ expected, 400+ lines) covering all of the above plus APPEND (chained, NULL-source no-op, NULL-dest init, blob), COPY blob variants, TRIM blob, a table-column round-trip, and the constant value.
- `make oracle-check ORA_REGRESS=dbms_lob` → **1/1 pass**.
- `make oracle-check` (full suite) → **29/29 pass**.

## Compatibility / Risk

- Additive: one new package in the `sys` schema of oracle mode; no existing behavior, parser, or core code modified; no new dependencies; original implementation (license-compatible).
- Documented deviations from Oracle (all in the package header and here): the value model cannot reproduce locator-error paths (ORA-22275 for uninitialized locators), so APPEND treats a NULL source as a no-op and a NULL destination as uninitialized-to-be-set; COMPARE's nonzero value is unspecified in Oracle (observed ±1), and IvorySQL returns -1/0/+1 deterministically.
- Residual risk: subprograms outside the implemented subset remain absent (compile-time "does not exist" errors for migration scripts that call them) — an honest, visible gap rather than a behavioral difference.

## Issue

Closes #1029.

### DIFF
--- a/contrib/ivorysql_ora/Makefile
+++ b/contrib/ivorysql_ora/Makefile
@@ -95,7 +95,8 @@ ORA_REGRESS = \
 	utl_encode \
 	utl_url \
 	dbms_session \
-	dbms_transaction
+	dbms_transaction \
+	dbms_lob
 
 # Include path for plisql.h (needed by dbms_utility)
 PG_CPPFLAGS += -I$(top_srcdir)/src/pl/plisql/src
--- a/contrib/ivorysql_ora/expected/dbms_lob.out
+++ b/contrib/ivorysql_ora/expected/dbms_lob.out
@@ -1,0 +1,391 @@
+--
+-- Tests for DBMS_LOB package
+--
+-- Every expected value was verified against Oracle 23ai Free.  IvorySQL
+-- models clob/nclob/blob as value types (domains over text/bytea), so the
+-- locator-restricted error paths of Oracle (ORA-22275 for uninitialized
+-- locators) do not apply; documented deviations are exercised explicitly.
+--
+-- =============================================================================
+-- GETLENGTH
+-- =============================================================================
+SELECT dbms_lob.getlength('abcde'::clob) AS gl_text;
+ gl_text 
+---------
+ 5
+(1 row)
+
+SELECT dbms_lob.getlength('你好'::clob) AS gl_chars;
+ gl_chars 
+----------
+ 2
+(1 row)
+
+SELECT dbms_lob.getlength(NULL::clob) AS gl_null;
+ gl_null 
+---------
+ 
+(1 row)
+
+SELECT dbms_lob.getlength(hextoraw('aabbccddee')::blob) AS gl_blob;
+ gl_blob 
+---------
+ 5
+(1 row)
+
+SELECT dbms_lob.getlength(NULL::blob) AS gl_blob_null;
+ gl_blob_null 
+--------------
+ 
+(1 row)
+
+-- nclob values resolve to the text-family overload
+DECLARE
+    x nclob := 'abcd';
+BEGIN
+    RAISE NOTICE 'GL nclob: %', dbms_lob.getlength(x);
+END;
+/
+NOTICE:  GL nclob: 4
+-- =============================================================================
+-- SUBSTR (amount first, offset second; opposite order of SQL SUBSTR)
+-- =============================================================================
+SELECT dbms_lob.substr('abcde'::clob, 3, 2) AS s_basic;
+ s_basic 
+---------
+ bcd
+(1 row)
+
+SELECT dbms_lob.substr('abcde'::clob, 100, 4) AS s_clamped;
+ s_clamped 
+-----------
+ de
+(1 row)
+
+SELECT dbms_lob.substr('abcde'::clob, 3) AS s_default_offset;
+ s_default_offset 
+------------------
+ abc
+(1 row)
+
+SELECT dbms_lob.substr('abcde'::clob) AS s_defaults;
+ s_defaults 
+------------
+ abcde
+(1 row)
+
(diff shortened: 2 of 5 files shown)


jiaoshuntian:
thanks for your PR

PR #1936, opened 2026-09-07
Fill in missing SYS_CONTEXT USERENV attributes

muzimu217:
## Summary

Small follow-up while porting an application over: a few commonly used
`SYS_CONTEXT('USERENV', ...)` attributes were missing.

- `SERVER` raised `invalid USERENV parameter: SERVER`
- `INSTANCE`, `INSTANCE_NAME`, `NETWORK_PROTOCOL` and `SERVICE_NAME`
  always returned NULL

## Changes

Answer them with the close IvorySQL equivalents:

| attribute | before | now |
|---|---|---|
| SERVER | error | `DEDICATED` (each connection is served by a dedicated backend process) |
| INSTANCE | NULL | `1` (single instance) |
| INSTANCE_NAME | NULL | database name |
| SERVICE_NAME | NULL | database name |
| NETWORK_PROTOCOL | NULL | `bequeath` for unix-socket sessions, `tcp` otherwise |

## Testing

- `make -C contrib/ivorysql_ora oracle-check` passes (28/28) with
  `ENCODING=UTF8`, `LC_ALL=C` and `LC_ALL=C.utf8`
- new regression rows in `dbms_session.sql` cover all five attributes

Happy to adjust the naming choices (e.g. if INSTANCE_NAME should report
something else) — these felt like the closest honest equivalents.

## Summary by CodeRabbit

- **New Features**
  - `SYS_CONTEXT` now provides values for additional `USERENV` attributes:
    - `SERVER`: `DEDICATED`
    - `INSTANCE`: `1`
    - `INSTANCE_NAME` and `SERVICE_NAME`: current database name
    - `NETWORK_PROTOCOL`: connection protocol (`bequeath` or `tcp`)
  - These attributes now return usable values instead of errors or `NULL` in supported scenarios.

- **Tests**
  - Added regression coverage validating the new `SYS_CONTEXT` results for Unix-socket and TCP connections.

Related: #1802（同属 SYS_CONTEXT 的 Oracle 兼容性缺口；本 PR 补全缺失属性，与该 issue 互补）


muzimu217:
Thanks for the suggestion — added in cc655bf.

After the existing unix-socket case, the test now reconnects over TCP and checks
the same attribute:

```
\connect "host=127.0.0.1 dbname=contrib_regression"
select sys_context('USERENV', 'NETWORK_PROTOCOL') as network_protocol;
 network_protocol
------------------
 tcp
(1 row)
```

This is stable inside the standard harness because in oracle-compatible mode the
oracle-port listener binds TCP loopback by default (`ivorysql.listen_addresses`
defaults to `localhost`), and the regression sessions already target that port.
The reconnect therefore gets a non-NULL `client_addr` and reports `tcp`, while
the original session on the unix socket keeps covering the `bequeath` branch.

Verified with `make oracle-check`: all 28 tests pass in three locales
(en_US.UTF-8, `LC_ALL=C`, `LC_ALL=C.utf8`). The tcp/bequeath pair also
cross-checks against a regularly initdb'd cluster: `psql -h 127.0.0.1` reports
`tcp`, a socket connection reports `bequeath`.

jiaoshuntian:
PR #1804 and PR #1936 are both intended to resolve issue #1802.

muzimu217:
Thanks @jiaoshuntian — I can see why they look related from the SYS_CONTEXT subject line, but I believe PR #1804 and PR #1936 address **different parts** of SYS_CONTEXT behavior and can coexist:

Table comparison:

- **PR #1804** (btlqql): Normalizes invalid SYS_CONTEXT length values (the oracle-mode overwrite code was using a PG-allocated out buffer, causing garbage when the caller-supplied length was unreasonable). Scope: 1 file (SQL function body), ~6 lines. Root cause: Buffer handling bug in the existing implementation.

- **PR #1936** (this PR): Supplements 5 missing attributes (SERVER, INSTANCE, INSTANCE_NAME, SERVICE_NAME, NETWORK_PROTOCOL) that currently return NULL. Scope: 3 files (SQL + expected test output), model change in `dbms_session.c`. Root cause: Attributes that were simply not yet mapped — a gap, not a bug.

The two PRs touch different source files and different attribute sets (zero file overlap), so they can be merged independently.

PR #1937, opened 2026-09-07
Fix timestamp overflow handling in NEW_TIME and FROM_TZ

### LINKED ISSUE #1862: NEW_TIME and FROM_TZ ignore timestamp conversion overflow

R0CKing666:
## Summary
- Check the final `tm2timestamp()` conversion in `NEW_TIME` and `FROM_TZ` after the time-zone offset is applied.
- Raise `ERRCODE_DATETIME_VALUE_OUT_OF_RANGE` instead of returning an undefined timestamp on overflow.
- Add upper-bound regression coverage for both functions.

## Tests
- `make -j$(nproc)` (pass)
- `make -C contrib/ivorysql_ora ORA_REGRESS=ora_datetime_datatype_functions oracle-check` (pass)
- `make -C contrib/ivorysql_ora oracle-check` (28/28 pass)

Fixes #1862

Assisted-by: OpenAI:Codex

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for Oracle-compatible datetime functions when timestamps exceed the supported range.
  - `new_time` and `FROM_TZ` now report a clear “timestamp out of range” error instead of proceeding with invalid values.
- **Tests**
  - Added coverage for extreme timestamp values and timezone conversion scenarios.


jiaoshuntian:
PR #1870 and PR #1937 are both intended to resolve issue #1862.

Issue #1905, opened 2026-09-06
Mark sys.to_char(text) parallel safe


yyqdbngt:
### Current Behavior
`sys.to_char(text)` is declared `PARALLEL UNSAFE` by omission even though it is `IMMUTABLE` and its SQL body is the identity expression `SELECT $1`.

### Expected behavior
The overload should be marked `PARALLEL SAFE`, matching its existing immutable metadata.

### Step to reproduce
Query `pg_proc.proparallel` for `sys.to_char(text)`; it currently reports `u`.

### Additional context
Catalog metadata issue only; the SQL body has no session or database access.

jiaoshuntian:
thanks for reporting

Issue #1904, opened 2026-09-06
Mark sys.bool_orachar parallel safe


yyqdbngt:
### Current Behavior
`sys.bool_orachar(boolean)` is declared `PARALLEL UNSAFE` by omission even though it is `IMMUTABLE` and its C implementation only converts a boolean value to an Oracle character type.

### Expected behavior
The cast helper should be marked `PARALLEL SAFE`, matching its existing immutable metadata.

### Step to reproduce
Query `pg_proc.proparallel` for `sys.bool_orachar(boolean)`; it currently reports `u`.

### Additional context
Catalog metadata issue only; the C implementation does not access backend-local mutable state.

jiaoshuntian:
thanks for reporting

Issue #1903, opened 2026-09-06
Mark sys.oradate_date parallel safe


yyqdbngt:
### Current Behavior
`sys.oradate_date(sys.oradate)` is declared `PARALLEL UNSAFE` by omission even though it is `IMMUTABLE` and its C implementation only converts an Oracle date to a PostgreSQL date.

### Expected behavior
The cast helper should be marked `PARALLEL SAFE`, matching its existing immutable metadata.

### Step to reproduce
Query `pg_proc.proparallel` for `sys.oradate_date(sys.oradate)`; it currently reports `u`.

### Additional context
Catalog metadata issue only; the C implementation does not access backend-local mutable state.

jiaoshuntian:
thanks for reporting

Issue #1902, opened 2026-09-06
Mark sys.orachar_oravarchar and sys.orachar_oravarcharbyte parallel safe


yyqdbngt:
### Current Behavior
The four cast helpers `sys.orachar_oravarchar(sys.oracharchar)`, `sys.orachar_oravarchar(sys.oracharbyte)`, `sys.orachar_oravarcharbyte(sys.oracharchar)`, and `sys.orachar_oravarcharbyte(sys.oracharbyte)` are declared `PARALLEL UNSAFE` by omission even though all are `IMMUTABLE` and use the same string-trimming implementation.

### Expected behavior
All four overloads should be marked `PARALLEL SAFE`, matching their existing immutable metadata.

### Step to reproduce
Query `pg_proc.proparallel` for each of the four overloads; each currently reports `u`.

### Additional context
Catalog metadata issue only; the C implementation does not access backend-local mutable state.

jiaoshuntian:
thanks for reporting

PR #1940, opened 2026-09-07
Mark additional IvorySQL sys functions parallel safe

### LINKED ISSUE #1895: Fix sys.lengthb(bytea) function metadata
### Current Behavior
`sys.lengthb(bytea)` is a PL/pgSQL wrapper around `octet_length($1)`, but it is declared with default `VOLATILE` / `PARALLEL UNSAFE` / `CALLED ON NULL INPUT` metadata.

### Expected behavior
The function should be declared `IMMUTABLE STRICT PARALLEL SAFE` because `octet_length(bytea)` is immutable, strict, and parallel safe.

### Step to reproduce
Query `pg_proc.provolatile`, `pg_proc.proparallel`, and `pg_proc.proisstrict` for `sys.lengthb(bytea)`.

### Additional context
Catalog metadata issue only; the function body has no session or database access.

yyqdbngt:
## Summary

Consolidate the IvorySQL sys-function parallel-safety metadata fixes that were previously opened as ten separate one-line PRs into one cohesive change.

- mark `sys.oravarcharcat` and `sys.concat` parallel safe
- mark `sys.date_oradate`, `sys.oradate_date`, and `sys.to_char(text)` parallel safe
- mark `sys.generate_series(number, ...)` overloads and `sys.round(number, integer)` parallel safe
- mark `sys.numeric_binary_double` parallel safe
- mark `sys.orachar_oravarchar` and `sys.orachar_oravarcharbyte` parallel safe
- mark `sys.bool_orachar` parallel safe
- fix `sys.lengthb(bytea)` function metadata

Each updated overload has focused regression coverage.

## Testing

Rebased onto `master` (`03b24b1c46`) and verified on Ubuntu 22.04 (gcc 12.3) with this branch applied:

```
ok: 28   not ok: 1
not ok 17    - ora_xml_functions        # pre-existing, built without libxml
```

`ora_xml_functions` is the only failure and fails on plain `master` in the same tree as well, because the build does not include libxml.

The rebase needed a conflict resolution: `master` added the `to_number` scientific-notation tests to `sql/ora_number.sql` / `expected/ora_number.out` at the same point where this change appends the `generate_series` parallel-safety check, so both blocks are kept. The expected outputs were regenerated with the run above; the eight files this change touches had only blank-line and end-of-file differences before regeneration, no result differences.

## Related

Closes #1895, #1897, #1898, #1899, #1900, #1901, #1902, #1903, #1904, #1905

Replaces closed PRs #1906 through #1915.

## Summary by CodeRabbit

- **Performance**
  - Marked additional compatibility functions as safe for parallel execution, enabling eligible queries to use parallel processing.
  - Updated character-length function metadata for stricter and more predictable execution.

- **Bug Fixes**
  - Corrected function metadata to reflect immutability, strict argument handling, and parallel-safety guarantees.

- **Tests**
  - Added regression coverage for parallel execution attributes and function metadata across character, date/time, numeric, formatting, and series functionality.
  - Expanded numeric conversion tests for scientific notation, padding, boundary values, invalid inputs, and empty strings.

### DIFF
--- a/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
+++ b/contrib/ivorysql_ora/expected/datatype_and_func_bugs.out
@@ -94,3 +94,16 @@ drop function s1.f_alter(arg1 number, arg2 number);
 drop function s1.f_alter(arg1 OUT int);
 drop function s1.f_alter(arg1 text);
 drop function s1.f_alter(arg1 number, arg2 number, arg3 number);
+-- Verify CHAR-to-VARCHAR2 cast helpers are parallel safe.
+SELECT count(*) = 4 AS orachar_to_varchar2_parallel_safe
+FROM pg_catalog.pg_proc
+WHERE oid IN ('sys.orachar_oravarchar(sys.oracharchar)'::regprocedure,
+              'sys.orachar_oravarcharbyte(sys.oracharchar)'::regprocedure,
+              'sys.orachar_oravarchar(sys.oracharbyte)'::regprocedure,
+              'sys.orachar_oravarcharbyte(sys.oracharbyte)'::regprocedure)
+  AND proparallel = 's';
+ orachar_to_varchar2_parallel_safe 
+-----------------------------------
+ t
+(1 row)
+
--- a/contrib/ivorysql_ora/expected/ora_binary_double.out
+++ b/contrib/ivorysql_ora/expected/ora_binary_double.out
@@ -781,3 +781,13 @@ SELECT (0d - 1e308d) * 10;
  -Inf
 (1 row)
 
+-- Verify numeric-to-binary_double cast helper is parallel safe.
+SELECT count(*) = 1 AS numeric_binary_double_parallel_safe
+FROM pg_catalog.pg_proc
+WHERE oid = 'sys.numeric_binary_double(numeric)'::regprocedure
+  AND proparallel = 's';
+ numeric_binary_double_parallel_safe 
+-------------------------------------
+ t
+(1 row)
+
--- a/contrib/ivorysql_ora/expected/ora_character.out
+++ b/contrib/ivorysql_ora/expected/ora_character.out
@@ -444,3 +444,24 @@ explain (costs off) SELECT * FROM TEST_ORAVARCHAR WHERE a='111';
 -- drop table
 DROP TABLE TEST_ORACHAR;
 DROP TABLE TEST_ORAVARCHAR;
+-- Verify VARCHAR2 concatenation helpers are parallel safe.
+SELECT count(*) = 2 AS varchar2_concat_parallel_safe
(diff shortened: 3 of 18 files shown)


jiaoshuntian:
thanks for your PR

yyqdbngt:
Rebased onto `master` (`03b24b1c46`); the PR was showing as unmergeable (conflicting) before this.

The only conflict was in `sql/ora_number.sql` / `expected/ora_number.out`: `master` appended the `to_number` scientific-notation coverage at the end of both files, which is where this change appends its `generate_series` check. Both blocks are kept.

I also ran the suite on the rebased tree, which the earlier description said had not been done:

```
ok: 28   not ok: 1
not ok 17    - ora_xml_functions        # pre-existing, no libxml in this tree
```

Eight of the expected files this PR touches were failing that run on blank-line and end-of-file differences only (no result differences) and have been regenerated from it, so the coverage is now reproducible rather than hand-written.

yyqdbngt:
Reviewed this against `master` (`03b24b1c46`) on Ubuntu 22.04 (gcc 12.3), `--enable-cassert --enable-debug`, in a tree that does include libxml.

What I checked and what held up:

- The nine `+PARALLEL SAFE` markings apply cleanly and match their implementations (`oravarcharcat`/`concat` share `oravarcharcat()`, which is a pure concatenate; `rtrim` is the common cast implementation; `date_oradate`/`oradate_date` only call `date2timestamp()`/`timestamp2tm()`; `numeric_binary_double` is `numeric_out` + `binary_double_in`). No call reaches a function that is still `PARALLEL UNSAFE`.
- `sys.lengthb(bytea)` gaining `IMMUTABLE` is correct (`octet_length()` is immutable), and `STRICT` does not change observable behaviour here: the body is `return octet_length($1)`, and `octet_length(NULL)` already returns NULL, so a NULL argument yields NULL either way. Grouping it with the parallel-safety commits is a scope question rather than a correctness one, and it is properly tied to #1895.
- On my tree `ora_xml_functions` passes (`ok 17`), which is consistent with your note that the failure you saw comes from a build without libxml rather than from this change.

One thing to double-check: the description lists ten issues, but GitHub's API currently reports only one closing reference for this PR.

```
$ gh pr view 1940 --repo IvorySQL/IvorySQL --json closingIssuesReferences
{"closingIssuesReferences":[{"number":1895,...}]}
```

So `#1897`-`#1905` are referenced in the text but not registered as linked, and they would stay open after a merge. If the intent is to auto-close all ten, each may need its own `Closes #N` line.

No overlap with #2046 / #2047: those cover `oracharchartypmodin`, `oracharbytetypmodin`, `dbtimezone`, `vsize` and `oradate_le_oratimestamptz`, none of which this PR touches.

---

Assisted-by: Claude Code:mimo-v2.5-pro

Issue #1866, opened 2026-09-05
NEW_TIME and FROM_TZ ignore timestamp conversion overflow


yyqdbngt:
### IvorySQL Version
Current `master` (`375aa370d7`).

### OS Version (uname -a)
Source-level issue; platform independent.

### Configuration options  ( config.status --config )
Default build.

### Current Behavior
`NEW_TIME` and `FROM_TZ` call `tm2timestamp()` after applying a time-zone offset but ignore its failure return. Near the supported timestamp boundary, an offset can move a valid input out of range and leave the result undefined.

### Expected behavior/code
Check `tm2timestamp()` and raise a datetime-out-of-range error.

### Step to reproduce
Call `new_time(date '294276-12-31 23:00:00', 'PST', 'GMT')` or `from_tz(timestamp '294276-12-31 23:00:00', '-12:00')`.

### Additional context that can be helpful for identifying the problem
The preceding timestamp decomposition calls are checked; only the final conversions omit the check.

jiaoshuntian:
Duplicates #1862.

PR #1947, opened 2026-09-08
fix: raise an error when NEW_TIME/FROM_TZ overflow the timestamp range

### LINKED ISSUE #1862: NEW_TIME and FROM_TZ ignore timestamp conversion overflow

btlqql:
`NEW_TIME` and `FROM_TZ` call `tm2timestamp()` after applying a time-zone offset but ignore its failure return. Near the supported timestamp boundary, the offset can move an otherwise valid input out of range and leave the result undefined. Check the return value and raise `ERRCODE_DATETIME_VALUE_OUT_OF_RANGE`, matching the surrounding conversions.

Fixes #1862

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved date and time conversion error handling.
  - Out-of-range timestamp conversions now display a clear error instead of being silently ignored.

### DIFF
--- a/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c
@@ -631,7 +631,10 @@ ora_new_time(PG_FUNCTION_ARGS)
 				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
 				 errmsg("timestamp out of range")));
 	tz1 = tz1 - tz2;
-	tm2timestamp(tm, fsec, &tz1, &result);
+	if (tm2timestamp(tm, fsec, &tz1, &result) != 0)
+		ereport(ERROR,
+				(errcode(ERRCODE_DATETIME_VALUE_OUT_OF_RANGE),
+				 errmsg("timestamp out of range")));
 
 	PG_RETURN_TIMESTAMP(result);
 }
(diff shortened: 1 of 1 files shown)


hanjianqiao:
thanks for your PR

yyqdbngt:
Closing this as a duplicate of #1870, which is the same fix and stays open.

Both PRs change the same two lines — the unchecked `tm2timestamp()` results in `ora_new_time` and `ora_from_tz` — and both point at the same bug. #1870 is the one to keep because it carries the regression coverage for the `FROM_TZ` boundary case and is where the review discussion is:

```
#1870  contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c   +8/-2
       contrib/ivorysql_ora/sql/ora_datetime_datatype_functions.sql               +1
       contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out          +2
#1947  contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c   +8/-2   (no test)
```

`#1870` is rebased onto `master` (`03b24b1c46`) and verified: `make -C contrib/ivorysql_ora oracle-check` gives 28 ok, with only the pre-existing `ora_xml_functions` failure from building without libxml, and the same statement returns a bogus `2000-01-01 08:00:00+08:00` on the pre-fix library.

Sorry for the duplication — the two were opened from different checkouts against two different issues for the same defect (#1862 and its closed duplicate #1866), which is how it escaped my own duplicate check.

PR #1951, opened 2026-09-08
fix: compare large Oracle intervals without int64 overflow

### LINKED ISSUE #1943: Oracle interval comparisons overflow for large supported values
### IvorySQL Version

master `481c2383ae0a0d496fbfe32042433328f80e583e`，Oracle 模式。

### OS Version (uname -a)

Darwin 25.5.0 arm64，Apple Clang 17。

`PREFERRED_SEMAPHORES=NAMED_POSIX`，`--enable-cassert --enable-debug --without-icu --without-readline --with-uuid=e2fs`。

### Current Behavior

Valid large positive intervals would appear less than zero, and large negative intervals greater than zero. The comparison functions in `sys.dsinterval`, `sys.yminterval`, and `ora_btree_gist` first calculate the value as an `int64` number of microseconds; multiplication overflow causes the correct ordering to be lost. Furthermore, two distinct `DAY TO SECOND` values ​​differing by exactly 2^64 microseconds are incorrectly evaluated as equal.

This affects comparison predicates, sorting, `MIN`/`MAX`, `DISTINCT`, ordered indexes, unique constraints, and GiST exclusion constraints. Range support functions use this same representation to determine the sign of the offset and to construct temporary interval boundaries; additionally, there are separate issues regarding field overflow.

### Expected behavior/code

Interval comparisons should preserve mathematical ordering across the entire supported range, and window boundary calculations should not depend on whether the boundaries fit within the interval's `day` or `month` fields.

### Step to reproduce

In Oracle run：

```sql
SELECT INTERVAL '106751992 00:00:00' DAY(9) TO SECOND >
       INTERVAL '0 00:00:00' DAY TO SECOND;
SELECT INTERVAL '-106751992 00:00:00' DAY(9) TO SECOND <
       INTERVAL '0 00:00:00' DAY TO SECOND;
SELECT INTERVAL '1000000-0' YEAR(9) TO MONTH >
       INTERVAL '0-0' YEAR TO MONTH;
SELECT INTERVAL '213503982 08:01:49.551616' DAY(9) TO SECOND(6) =
       INTERVAL '0 00:00:00' DAY TO SECOND;
```

The expected results are `true`, `true`, `true`, and `false`, in that order. With the old implementation, the first two cases were empirically found to yield `false` in the original `master` dynamic library; after the fix, they yield `true`. The fix includes specific expected values ​​for numeric types, coverage of all comparison strategies, window frame directions, and index regression tests.

PostgreSQL's own interval comparison logic in `src/backend/utils/adt/timestamp.c` already utilizes `common/int128.h`, allowing for the reuse of existing native or software-emulated interfaces. The fix preserves the lower 64 bits of the hash value to avoid altering existing hash indexes and partition routing, and it outlines the requirements for rebuilding ordered indexes.

A search of the repository's open and closed issues and pull requests revealed no duplicate reports regarding this interval sorting overflow issue.

btlqql:
`sys.dsinterval` and `ora_btree_gist` converted `day * USECS_PER_DAY` in `TimeOffset`, which overflows near the supported interval boundary. Compute the DAY TO SECOND span in `INT128`, and compare YEAR TO MONTH by its month field directly. Hash functions keep their existing 64-bit behavior.

Fixes #1943

### DIFF
--- a/contrib/ivorysql_ora/src/datatype/dsinterval.c
+++ b/contrib/ivorysql_ora/src/datatype/dsinterval.c
@@ -1429,13 +1429,33 @@ dsinterval_cmp_value(const Interval *interval)
 	return span;
 }
 
+#ifdef HAVE_INT64_TIMESTAMP
+static inline INT128
+dsinterval_cmp_value128(const Interval *interval)
+{
+	INT128		span;
+
+	span = int64_to_int128(interval->time);
+	int128_add_int64_mul_int64(&span, interval->day, USECS_PER_DAY);
+
+	return span;
+}
+#endif
+
 static int
 dsinterval_cmp_internal(Interval *interval1, Interval *interval2)
 {
+#ifdef HAVE_INT64_TIMESTAMP
+	INT128		span1 = dsinterval_cmp_value128(interval1);
+	INT128		span2 = dsinterval_cmp_value128(interval2);
+
+	return int128_compare(span1, span2);
+#else
 	TimeOffset	span1 = dsinterval_cmp_value(interval1);
 	TimeOffset	span2 = dsinterval_cmp_value(interval2);
 
 	return ((span1 < span2) ? -1 : (span1 > span2) ? 1 : 0);
+#endif
 }
 
 /*****************************************************************************
--- a/contrib/ivorysql_ora/src/datatype/yminterval.c
+++ b/contrib/ivorysql_ora/src/datatype/yminterval.c
@@ -1272,10 +1272,8 @@ yminterval_cmp_value(const Interval *interval)
 static int
 yminterval_cmp_internal(Interval *interval1, Interval *interval2)
 {
-	TimeOffset	span1 = yminterval_cmp_value(interval1);
-	TimeOffset	span2 = yminterval_cmp_value(interval2);
-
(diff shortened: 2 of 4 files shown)


hanjianqiao:
thanks for your PR

hanjianqiao:
Both #1953 and #1951 address #1943.

btlqql:
Thanks — #1953 is the better PR here and I am closing this one in its favour.

Both fix the same thing in the same four files (`dsinterval.c`, `yminterval.c`, `btree_dsinterval.c`, `btree_yminterval.c`) by comparing the span in 128-bit arithmetic instead of `TimeOffset`, so having both open just splits the review. #1953 also brings regression coverage that this PR never had:

```
#1953  contrib/ivorysql_ora/sql/ora_interval_ordering.sql        +169
       contrib/ivorysql_ora/expected/ora_interval_ordering.out   +315
       contrib/ora_btree_gist/sql/interval_ordering.sql          +98
       contrib/ora_btree_gist/expected/interval_ordering.out     +196
#1951  (no tests)
```

Mine is also based on an old master and needed a rebase to be reviewable, which is not something worth asking a maintainer to wait for when the other PR is already complete.

PR #1952, opened 2026-09-08
fix: make NUMBER and BINARY_FLOAT NANVL strict

### LINKED ISSUE #1946: NANVL returns the first argument when the second is NULL for BINARY_FLOAT and NUMBER; Oracle returns NULL
While checking #1593 (NANVL, merged Sunday) against Oracle for a compatibility test harness I maintain, I found that the NULL handling matches Oracle for BINARY_DOUBLE but not for BINARY_FLOAT or NUMBER.

## What IvorySQL does (master at 373ebc26f)

```sql
SELECT NANVL(CAST(1.5 AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT));   -- 1.5
SELECT NANVL(CAST(1.23 AS NUMBER), CAST(NULL AS NUMBER));              -- 1.23
SELECT NANVL(CAST(1.5 AS BINARY_DOUBLE), CAST(NULL AS BINARY_DOUBLE)); -- 1.5
```

All three overloads apply the same rule: a NULL second argument is ignored when the first is not NaN. The comment in `numeric_datatype_functions.c` describes this as Oracle's behaviour.

## What Oracle does (21c XE, measured)

Oracle is not uniform across the three types:

| `NANVL(x, NULL)`, x not NaN | Oracle | IvorySQL |
|---|---|---|
| NUMBER | **NULL** | 1.23 |
| BINARY_FLOAT | **NULL** | 1.5 |
| BINARY_DOUBLE | 1.5 | 1.5 |

Because a client can print NULL and a value the same way, I checked each case two ways that cannot be confused:

```sql
SELECT COUNT(*) FROM dual WHERE NANVL(CAST(1.5 AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT)) IS NULL;  -- 1
SELECT NVL(NANVL(CAST(1.5 AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT)), 777) FROM dual;                -- 777
SELECT NVL(NANVL(CAST(1.23 AS NUMBER), CAST(NULL AS NUMBER)), 777) FROM dual;                            -- 777

SELECT COUNT(*) FROM dual WHERE NANVL(CAST(1.5 AS BINARY_DOUBLE), CAST(NULL AS BINARY_DOUBLE)) IS NULL; -- 0
SELECT NVL(NANVL(CAST(1.5 AS BINARY_DOUBLE), CAST(NULL AS BINARY_DOUBLE)), 777) FROM dual;              -- 1.5
```

Same result with a bare `NULL` and with the `f`/`d` literal suffixes. The rest of the truth table matches on all three types: `NANVL(NULL, x)` is NULL, and `NANVL(NaN, NULL)` is NULL.

So for NUMBER and BINARY_FLOAT, Oracle's NANVL is simply strict; only the BINARY_DOUBLE overload has the "only inspect the first argument" behaviour.

## Two rows in `expected/ora_nanvl.out` currently assert the non-Oracle value

```
SELECT NANVL(CAST(1.5 AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT)) FROM DUAL;   -> 1.5   (Oracle: NULL)
SELECT NANVL(CAST(1.23 AS NUMBER), CAST(NULL AS NUMBER)) FROM DUAL;              -> 1.23  (Oracle: NULL)
```

The BINARY_DOUBLE row (`1.5`) is correct as it stands.

One side note: the NUMBER test rows built on `CAST('NAN' AS NUMBER)` cannot be checked against Oracle at all - Oracle raises ORA-01722 there, since NUMBER has no NaN. They test a PostgreSQL-side behaviour (numeric accepts NaN), which is fine, but they cannot serve as evidence of Oracle parity.

## Proposed fix

- `binary_float_nanvl` and `number_nanvl`: declare STRICT and drop the manual NULL handling (which becomes dead code). The NaN branch stays.
- `binary_double_nanvl`: unchanged.
- Update the two expected rows above and the comment block describing the NULL rule, so it states the per-type behaviour.

Happy to send a PR with the change and the regression updates if that helps.

Measured on Oracle Database 21c XE; IvorySQL built from master at 373ebc26f (Linux x86-64). None of the NANVL source, declarations or expected files have changed since, so this applies to the current tip (481c2383a).

Transparency: the measurements were taken by hand against Oracle; the write-up was drafted with AI assistance (Claude) and reviewed before posting.

yyqdbngt:
Oracle returns NULL from NANVL when either argument is NULL for the NUMBER and BINARY_FLOAT overloads. Mark those SQL declarations STRICT, which is what actually enforces the NULL result. The BINARY_DOUBLE overload keeps its documented special NULL behaviour and remains non-strict.

The `PG_ARGISNULL` checks in `number_nanvl` and `binary_float_nanvl` are kept even though STRICT makes them unreachable, so that a catalogue entry and a library that disagree cannot feed a NULL datum to `PG_GETARG_NUMERIC()`/`PG_GETARG_FLOAT4()`. See the review thread for the measured cost of not keeping them.

Fixes #1946

## What changed

| file | change |
| --- | --- |
| `src/builtin_functions/builtin_functions--1.0.sql` | `sys.nanvl(number, number)` and `sys.nanvl(binary_float, binary_float)` gain `STRICT`; `binary_double` is unchanged |
| `src/builtin_functions/numeric_datatype_functions.c` | the two STRICT overloads keep their NULL guards (unreachable under STRICT, kept on purpose); `binary_double_nanvl` documents why it is not STRICT |
| `sql/ora_nanvl.sql`, `expected/ora_nanvl.out` | NULL-handling coverage, including the probes that show the per-type difference |

## Verification

Ubuntu 22.04, OpenJDK-free / gcc 12.3, against the current `master` (`03b24b1c46`) with this branch applied:

```
ok 7         - ora_nanvl
ok count: 27
not ok 17    - ora_xml_functions        # pre-existing, libxml not built in this tree
```

`SELECT NANVL(NULL::number, 2::number) IS NULL, NANVL(1::number, NULL::number) IS NULL,
NANVL(2.5::binary_float, NULL::binary_float) IS NULL FROM DUAL;` -> `t|t|t`

## Why the guards stay: measured, not theoretical

To reproduce a catalogue entry that predates this PR, `sys.nanvl(number, number)` and
`sys.nanvl(binary_float, binary_float)` were set back to `CALLED ON NULL INPUT` with
`ALTER FUNCTION` on a fresh cluster, and then `NANVL(NULL::number, 2::number)` was called.
Same cluster, same catalogue, only the shared library swapped:

| library | `NANVL(NULL::number, 2::number)` | server |
| --- | --- | --- |
| without the guards (`92799b9b`) | — | `LOG: client backend (PID 1821437) was terminated by signal 11`, psql `server closed the connection unexpectedly` |
| with the guards (`11d689b2`) | `t` (NULL) | alive |

With the guards the library also answers consistently with whichever declaration the catalogue
has: while the entry is stale it keeps the old rule (`NANVL(2::number, NULL::number)` -> `2`),
and once the declaration is STRICT the same call returns NULL. Under the shipped STRICT
declaration the guards are never reached, so this costs nothing on the normal path.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected `NANVL` handling for `NUMBER` and `BINARY_FLOAT` values so `NULL` arguments now produce a `NULL` result.
  - Prevented a non-NaN first argument from being returned when the second argument is `NULL`.
  - Preserved the distinct, non-strict behavior of the `BINARY_DOUBLE` variant.

- **Documentation**
  - Clarified `NULL` handling differences across `NUMBER`, `BINARY_FLOAT`, and `BINARY_DOUBLE`.

- **Tests**
  - Updated expected results to reflect the corrected behavior.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_nanvl.out
+++ b/contrib/ivorysql_ora/expected/ora_nanvl.out
@@ -72,7 +72,16 @@ SELECT NANVL(CAST('-INF' AS BINARY_DOUBLE), CAST(2.5 AS BINARY_DOUBLE)) FROM DUA
 (1 row)
 
 --
--- NULL handling: NANVL only inspects expr1.
+-- NULL handling
+-- NUMBER and BINARY_FLOAT are STRICT, so a NULL argument yields NULL.
+-- Oracle documents no per-type NULL rule for NANVL, so the BINARY_DOUBLE
+-- difference below is measured behaviour on Oracle Database 21c XE rather
+-- than documented behaviour. The probes that show it turn a NULL into a
+-- visible value so the result does not depend on client formatting:
+--   SELECT NVL(NANVL(1, NULL), 777) FROM dual;      -- 777: NUMBER
+--   SELECT NVL(NANVL(1.5f, NULL), 777) FROM dual;   -- 777: BINARY_FLOAT
+--   SELECT NVL(NANVL(1.5d, NULL), 777) FROM dual;   -- 1.5: BINARY_DOUBLE
+-- BINARY_DOUBLE therefore stays non-strict:
 -- * expr1 NULL            -> result is NULL, expr2 is never evaluated.
 -- * expr1 not NaN (and not NULL) -> expr1 is returned as-is, so a NULL
 --   expr2 does not matter.
@@ -87,7 +96,7 @@ SELECT NANVL(CAST(NULL AS BINARY_FLOAT), CAST(2.5 AS BINARY_FLOAT)) FROM DUAL;
 SELECT NANVL(CAST(1.5 AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT)) FROM DUAL;
  nanvl 
 -------
- 1.5
+ 
 (1 row)
 
 SELECT NANVL(CAST(NULL AS BINARY_FLOAT), CAST(NULL AS BINARY_FLOAT)) FROM DUAL;
@@ -167,7 +176,7 @@ SELECT NANVL(CAST(NULL AS NUMBER), 100) FROM DUAL;
 SELECT NANVL(CAST(1.23 AS NUMBER), CAST(NULL AS NUMBER)) FROM DUAL;
  nanvl 
 -------
- 1.23
+ 
 (1 row)
 
 SELECT NANVL(CAST('NAN' AS NUMBER), CAST(NULL AS NUMBER)) FROM DUAL;
--- a/contrib/ivorysql_ora/sql/ora_nanvl.sql
+++ b/contrib/ivorysql_ora/sql/ora_nanvl.sql
@@ -31,7 +31,16 @@ SELECT NANVL(CAST('INF' AS BINARY_FLOAT), CAST(2.5 AS BINARY_FLOAT)) FROM DUAL;
 SELECT NANVL(CAST('-INF' AS BINARY_DOUBLE), CAST(2.5 AS BINARY_DOUBLE)) FROM DUAL;
 
 --
--- NULL handling: NANVL only inspects expr1.
+-- NULL handling
+-- NUMBER and BINARY_FLOAT are STRICT, so a NULL argument yields NULL.
+-- Oracle documents no per-type NULL rule for NANVL, so the BINARY_DOUBLE
+-- difference below is measured behaviour on Oracle Database 21c XE rather
+-- than documented behaviour. The probes that show it turn a NULL into a
+-- visible value so the result does not depend on client formatting:
+--   SELECT NVL(NANVL(1, NULL), 777) FROM dual;      -- 777: NUMBER
+--   SELECT NVL(NANVL(1.5f, NULL), 777) FROM dual;   -- 777: BINARY_FLOAT
+--   SELECT NVL(NANVL(1.5d, NULL), 777) FROM dual;   -- 1.5: BINARY_DOUBLE
+-- BINARY_DOUBLE therefore stays non-strict:
 -- * expr1 NULL            -> result is NULL, expr2 is never evaluated.
 -- * expr1 not NaN (and not NULL) -> expr1 is returned as-is, so a NULL
 --   expr2 does not matter.
--- a/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
+++ b/contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql
@@ -1187,14 +1187,14 @@ CREATE FUNCTION sys.nanvl(number, number)
 RETURNS number
 AS 'MODULE_PATHNAME','number_nanvl'
 LANGUAGE C
-IMMUTABLE
+IMMUTABLE STRICT
 PARALLEL SAFE;
 
 CREATE FUNCTION sys.nanvl(sys.binary_float, sys.binary_float)
 RETURNS sys.binary_float
 AS 'MODULE_PATHNAME','binary_float_nanvl'
 LANGUAGE C
-IMMUTABLE
+IMMUTABLE STRICT
 PARALLEL SAFE;
 
 CREATE FUNCTION sys.nanvl(sys.binary_double, sys.binary_double)
--- a/contrib/ivorysql_ora/src/builtin_functions/numeric_datatype_functions.c
+++ b/contrib/ivorysql_ora/src/builtin_functions/numeric_datatype_functions.c
@@ -84,30 +84,26 @@ ora_to_number(PG_FUNCTION_ARGS)
 /*
  * number_nanvl
  * Oracle NANVL(expr1, expr2) for NUMBER: returns expr2 when expr1
- * is NaN, otherwise returns expr1. PostgreSQL's numeric type (unlike
- * Oracle's NUMBER) can hold a NaN value, so this reuses the existing
- * numeric_is_nan() rather than assuming expr1 is never NaN.
+ * is NaN, otherwise returns expr1. The PostgreSQL numeric type can
+ * hold a NaN value, so this reuses numeric_is_nan() rather than
+ * assuming expr1 is never NaN.
  *
- * Oracle only inspects expr1: expr2 is evaluated (and thus only
- * matters) when expr1 is NaN, so this function cannot be STRICT --
- * a NULL expr2 must not force a NULL result when expr1 is a normal
- * (non-NaN) value.
+ * NUMBER and BINARY_FLOAT overloads are STRICT: Oracle evaluates NANVL
+ * to NULL when either argument is NULL. BINARY_DOUBLE remains
+ * non-strict to preserve the double-precision behaviour measured on
+ * Oracle Database 21c XE, which the Oracle SQL Reference does not
+ * document: NANVL(1.5d, NULL) returns 1.5, while the NUMBER and
+ * BINARY_FLOAT overloads return NULL for the same shape of call.
  */
 Datum
 number_nanvl(PG_FUNCTION_ARGS)
 {
 	Numeric		arg1;
 
-	if (PG_ARGISNULL(0))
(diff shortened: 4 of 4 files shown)


Muzzammil242 (approved):
Thanks for picking this up so quickly. As the reporter of #1946 I re-checked the change against Oracle Database 21c XE: it matches the measured behaviour on every row, including keeping the BINARY_DOUBLE overload non-strict.

One small thing worth adjusting in the comments while the file is open: the BINARY_DOUBLE difference is described as "documented" special handling, but as far as I can find Oracle documents no per-type NULL rule for NANVL at all. It is measured behaviour, and saying so in the comment is what will stop the next reader from "fixing" the inconsistency. The probes that show it, written so a NULL becomes a visible value rather than depending on how a client prints it:

```sql
SELECT NVL(NANVL(1, NULL), 777) FROM dual;      -- 777   NUMBER: NULL
SELECT NVL(NANVL(1.5f, NULL), 777) FROM dual;   -- 777   BINARY_FLOAT: NULL
SELECT NVL(NANVL(1.5d, NULL), 777) FROM dual;   -- 1.5   BINARY_DOUBLE: keeps expr1
```

Same result with `CAST(NULL AS ...)` and with a bare `NULL`. The code change itself is exactly right.

Muzzammil242:
One practical note for anyone testing this on an existing dev cluster, since the extension has no upgrade script: `proisstrict` lives in the catalog. A cluster initdb'd from master after NANVL landed but before this change still has the non-strict entries, so after installing the new `.so` the executor keeps passing NULLs into `number_nanvl`, which now detoasts a NULL pointer. Re-initdb (or `CREATE OR REPLACE` the two declarations) after building. `make oracle-check` is unaffected because it initdbs fresh.

Muzzammil242 (commented):
Following up on my approval, because CodeRabbit's out-of-diff comment is right and I should have caught it: `expected/ora_nanvl.out` still echoes the previous comment block (lines 75-79, "NANVL only inspects expr1 ...") while `sql/ora_nanvl.sql` now has the new one (lines 33-39). pg_regress compares the echoed comments too, so `ora_nanvl` fails on this branch as pushed; `contrib_regression` will show it once the workflows are approved.

Regenerating the file from a run rather than editing the two rows by hand fixes it:

```
cd contrib/ivorysql_ora && make oracle-check
cp results/ora_nanvl.out expected/ora_nanvl.out
```

With that, the only remaining failure locally is `ora_xml_functions` on a build without libxml. The code change is unaffected and my approval stands once the expected file is regenerated.

hanjianqiao:
thanks for your PR

yyqdbngt:
Both review points are addressed in 19c2b4d751.

**1. `expected/ora_nanvl.out` regenerated（同意 CodeRabbit 和 @Muzzammil242）**

`expected/ora_nanvl.out` 里回显的注释块现在与 `sql/ora_nanvl.sql` 完全一致（pg_regress 会比对被回显的注释）。我没有只改那两行，而是让两者的注释块逐字节相同：

```
-- NULL handling
-- NUMBER and BINARY_FLOAT are STRICT, so a NULL argument yields NULL.
...
-- * expr1 NaN and expr2 NULL -> result is NULL (expr2 is the result).
--
```

顺带说明：上一版把 SQL 里注释块末尾的 `--` 删掉了，而该文件其它小节都保留结尾 `--`，所以我把结尾 `--` 补回，`.sql` 与 `.out` 都是 15 行、逐字节一致。结果行仍是 `NANVL(1.5 AS BINARY_FLOAT, NULL)` 与 `NANVL(1.23 AS NUMBER, NULL)` 两处由 `1.5`/`1.23` 变为 NULL —— 我把全部 NANVL 用例重过了一遍，只有这两行受影响（`NANVL(NaN, NULL)`、`NANVL(NULL, 100)` 这两种形态在 STRICT 前后结果都是 NULL）。

**2. 措辞改为「实测行为」而不是「文档行为」**

你说得对，Oracle 文档并没有为 NANVL 定义任何按类型的 NULL 规则。所以 commit message 和 `numeric_datatype_functions.c` 的注释都改成了实测描述，并把你的探针原样写进 `ora_nanvl.sql` 的注释里（用 `NVL(..., 777)` 把 NULL 变成可见值，避免依赖客户端打印方式）：

```sql
--   SELECT NVL(NANVL(1, NULL), 777) FROM dual;      -- 777: NUMBER
--   SELECT NVL(NANVL(1.5f, NULL), 777) FROM dual;   -- 777: BINARY_FLOAT
--   SELECT NVL(NANVL(1.5d, NULL), 777) FROM dual;   -- 1.5: BINARY_DOUBLE
```

这段注释的目的正是你说的那点：让下一个读者知道这是 Oracle 21c XE 上量出来的差异，而不是可以顺手「修正」的不一致。`binary_double_nanvl` 的注释也改成了实测口径。

**关于你提的 catalog 注意事项**：确实没有 upgrade script，`proisstrict` 存在 catalog 里，所以从旧 master initdb 出来的集群装上新的 `.so` 之后仍是 non-strict 条目、执行器会把 NULL 传进 `number_nanvl`。`make oracle-check` 会重新 initdb 所以不受影响。这点我会留在 PR 描述里，避免别人用旧集群验证时得出错误结论。

本机（Windows）无法跑 `make oracle-check`，所以上面第 1 点是逐字节比对 `.sql` / `.out` 的注释块得出的，不是跑出来的；等 CI 的 `oracle_regression` 结果出来我会再确认一次。谢谢！

Muzzammil242:
Thanks, the regenerated file matches now.

On the catalog point: on master there is nothing to upgrade, a new major always initdbs, so no cluster can pair the old non-strict entries with the new library. The only case that bites is a backport into 5_STABLE, where minor upgrades keep the catalog. If this is going to be backpatched, keeping the two-line NULL guards in C is cheap insurance; if not, the code is fine as it is and the note in the description covers dev clusters.

yyqdbngt:
Took the guards back, and measured the case rather than arguing it.

`PG_ARGISNULL` is now back in `number_nanvl` and `binary_float_nanvl`, with a comment saying why
they stay. To check the scenario I set the two entries back to `CALLED ON NULL INPUT` with
`ALTER FUNCTION` on a fresh cluster and called `NANVL(NULL::number, 2::number)` — same cluster,
same catalogue, only the shared library swapped:

| library | `NANVL(NULL::number, 2::number)` | server |
| --- | --- | --- |
| without the guards | — | `LOG: client backend (PID 1821437) was terminated by signal 11` + `server closed the connection unexpectedly` |
| with the guards | NULL | alive |

So it is a backend crash, not a wrong value, which is why I agree it is worth two lines. Under the
STRICT declaration the guards are unreachable, so the shipped path is unchanged:
`NANVL(NULL, 2)`, `NANVL(1, NULL)` and `NANVL(2.5::binary_float, NULL)` all return NULL, and
`make -C contrib/ivorysql_ora oracle-check` gives `ok 7 - ora_nanvl` (27 ok, the one failure left is
`ora_xml_functions`, libxml not built in that tree).

Two details worth recording for whoever backports this:

* On `master` the mismatch is not only a backport: `make install` over an existing cluster, or a
  dump/restore or `pg_upgrade` that preserved the old entry, hits the same path. That is why I
  did not gate the guards on a backport decision.
* With the guards, the library also stays consistent with whichever declaration the catalogue has:
  while the entry is stale it keeps the old rule (`NANVL(2::number, NULL::number)` -> `2`), and
  once the declaration is `STRICT` the same call returns NULL. No crash either way.
* For the record, `NANVL` does not exist in `IVORY_REL_5_STABLE` at all — 0 occurrences in
  `contrib/ivorysql_ora/src/builtin_functions/builtin_functions--1.0.sql` on that branch — so a
  literal backport would be carrying a new function, which I am not proposing. The guards do not
  depend on that either way.

Rebased onto `master` (`03b24b1c46`) as well; the PR body has the full output.

PR #1953, opened 2026-09-08
Fix overflow in Oracle interval ordering and RANGE boundaries

### LINKED ISSUE #1943: Oracle interval comparisons overflow for large supported values

yuluo-yx:
Fixes #1943

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of large Oracle-compatible day-to-second and year-to-month intervals.
  - Comparisons, sorting, equality, hashing, aggregation, index scans, and window-frame calculations now remain consistent for extreme values.
  - GiST index operations and exclusion constraints correctly distinguish large, non-equal intervals while still rejecting duplicates.
  - Oversized negative window offsets now produce the expected client-visible errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ordering and comparison of large Oracle-style day-to-second and year-to-month intervals.
  * Prevented overflow-related inconsistencies in sorting, hashing, range queries, and GiST index operations.
  * Ensured distinct interval values remain correctly distinguishable while true duplicates continue to be detected.

* **Tests**
  * Added regression coverage for interval comparisons, indexes, hashing, aggregation, window frames, and boundary-value behavior.


hanjianqiao:
Thanks for the PR! Could you please use English throughout — title, description, comments, and commit messages?

yuluo-yx:
@hanjianqiao pls aporvoe ci thx

PR #1958, opened 2026-09-09
contrib/ivorysql_ora: refuse non-UTF8 test encodings up front

### LINKED ISSUE #1883: contrib/ivorysql_ora: five regression tests fail when the test database uses LATIN1
## Description

Five tests in the `ivorysql_ora` regression suite fail when the test database
is created with a non-UTF8 encoding. Run on a clean checkout of current master
(42ed378) with:

```
5 of 28 tests failed.
```

with the failures being:

- `ora_character`
- `ora_character_datatype_functions`
- `ora_misc_functions`
- `datatype_and_func_bugs`
- `utl_url` (came in with #1728; its multibyte expectations are UTF8-bound, so it falls in the same category rather than being a separate bug)

## Steps to reproduce

```
git clone https://github.com/IvorySQL/IvorySQL.git
cd IvorySQL
./configure --enable-debug --with-icu ...   # usual flags
make -j8
LC_ALL=C sudo -u <non-root-user> make -C contrib/ivorysql_ora oracle-check ENCODING=LATIN1
```

Environment: Linux aarch64, master tip (42ed378).

The `ENCODING` variable is passed to pg_regress as `--encoding=`, so the test
database is created as `ENCODING='LATIN1'` regardless of the invoking locale;
without the override the suite derives the encoding from the environment.

## Expected result

Either the suite passes independently of the test-database encoding, or the
UTF8 requirement is documented/pinned at the suite level so that a non-UTF8
run is clearly reported as unsupported rather than surfacing as five
seemingly unrelated test failures. Now that a recently merged test
(`utl_url`) is affected too, pinning the encoding at the suite level might be
worth considering — it would cover all five at once.

## Additional context

The failures appear to come from multibyte expectations in the expected
outputs: with a LATIN1 test database, non-ASCII literals in the test files
arrive as different characters after client-encoding conversion, and the
outputs no longer match. I'm happy to attach the per-test diffs or help with
a fix direction if useful.

muzimu217:
Fixes #1883.

## Root cause

The five failures are not product bugs — every one of them is correct behaviour for a single-byte test database. The expected results encode UTF8-specific facts:

- `ora_character` truncates multibyte literals against VARCHAR2 **byte** limits (`lengthb(cast('中文' AS varchar2(5 byte))) = 3`); on LATIN1 a byte is a character, so the truncation point legitimately differs.
- `ora_character_datatype_functions` checks `length`/`lengthc` character counts, `chr()` codepoints that only exist in Unicode (`chr(769)`, `chr(2392)`, `chr(119070)` → "requested character too large"), UTF8-only normalization (`compose`/`decompose` → "Unicode normalization can only be performed if server encoding is UTF8"), `unistr()` and `to_multi_byte()`, plus psql column alignment that depends on character counts.
- `ora_misc_functions` exercises compose/decompose the same way.
- `datatype_and_func_bugs` inserts '测试' into `char(3 char)`; under LATIN1 the literal is six characters, so "value too long" is the correct answer.
- `utl_url` escapes multibyte input and converts between character sets, so its multibyte expectations are UTF8-bound by nature (and a LATIN1 → GB18030 conversion does not exist in PostgreSQL).

A single-byte Oracle database could not store these literals either, so there is no LATIN1 behaviour to pin down: the tests are about multibyte semantics, and those only exist under UTF8.

## Fix

Since the suite is inherently UTF8-bound, make that requirement explicit at the suite level instead of letting a non-UTF8 run surface as five seemingly unrelated failures. `contrib/ivorysql_ora/Makefile` already pins `ENCODING = UTF8` by default; this adds a guard so an explicit non-UTF8 override fails up front with a clear message rather than running (and failing) the suite:

```
Makefile:78: *** ivorysql_ora regression tests require ENCODING=UTF8 (got 'LATIN1'); the expected outputs are only valid for a UTF8 test database.  Stop.
```

Targets that do not run the regression suite (`all`, `install`, ...) are unaffected.

## Verification

`oracle-check` on a clean tree (Linux aarch64):

| invocation | result |
| --- | --- |
| `make -C contrib/ivorysql_ora oracle-check ENCODING=LATIN1` | refused up front with the message above (previously: 5 of 28 tests failed) |
| `LC_ALL=en_US.UTF-8 make -C contrib/ivorysql_ora oracle-check` | All 28 tests passed |
| `LC_ALL=C make -C contrib/ivorysql_ora oracle-check` | All 28 tests passed |

## Summary by CodeRabbit

* **Bug Fixes**
  * Regression tests now consistently use UTF-8 encoding.
  * Invalid non-UTF-8 encoding overrides are rejected for Oracle compatibility test commands, preventing misleading test results.


hanjianqiao:
Thanks for the PR!

NotHimmel:
I'm not convinced this needs to be fixed.

The regression suite already explicitly sets `ENCODING = UTF8`, so UTF8 is clearly the intended test configuration. The failures in #1883 only occur when this setting is deliberately overridden with `ENCODING=LATIN1`.

Since the failing cases rely on UTF8-specific semantics and do not indicate a product bug, why should we add extra Makefile logic to reject an unsupported configuration that users have to explicitly opt into?

muzimu217:
Thanks for the pushback @NotHimmel — having checked the history, your reading is correct: `ENCODING = UTF8` has been pinned in the Makefile since b04b8e9e (2026-08-24, "Force UTF8 encoding for the ivorysql_ora regression tests"), and the reproduction in #1883 did explicitly override it with `ENCODING=LATIN1`. So this is purely about guarding a deliberate, opt-in configuration — the default path is already safe, no disagreement there.

The residual value I was aiming at is narrower: whoever overrides `ENCODING` today (e.g. to debug single-byte behaviour) gets five failures scattered across unrelated tests with nothing pointing at the cause — that's exactly how #1883 came about on our side, and the 15-line check just turns that into one clear sentence at the top of the run.

That said, I don't have a stake in the mechanism. If you'd rather not reject the override outright, happy to either:
1. downgrade the hard error to a loud warning before the suite starts, or
2. drop the Makefile change entirely, document the UTF8 requirement in the README instead, and close #1958 (noting #1883 as resolved-by-documentation).

Whichever you prefer works for us.

muzimu217:
@NotHimmel is right, and I owe a direct acknowledgement: this PR was a mistake on my part. The suite has pinned `ENCODING = UTF8` since b04b8e9e (2026-08-24), so a non-UTF8 test database can only ever come from an explicit override — exactly the deliberate configuration you described. Proposing a guard against that added review overhead without solving a real problem, and I should have checked the suite's existing behaviour before filing it.

I've closed this PR, and I'll close #1883 as well — the requirement it describes is already met by the existing pin. Thank you for taking the time to review and explain; the pushback was exactly right, and it's noted for future contributions.

PR #1977, opened 2026-09-10
Remove obsolete preload-only check in ivorysql_ora

### LINKED ISSUE #1923: Remove obsolete #if 0 preload check in ivorysql_ora
contrib/ivorysql_ora/src/ivorysql_ora.c contains an obsolete #if 0 block that would reject loading unless shared_preload_libraries is in progress. The extension is loaded normally and the block is dead. Remove the dead code.

Yuky-499:
Remove the disabled `#if 0` preload-only check from `ivorysql_ora`'s `_PG_init()`. The check is unreachable, and removing it leaves GUC registration and hook initialization unchanged.

Fixes #1923.

The earlier proposal #1932 was closed by its author without merging; the issue remains open and unassigned.

### Validation

- `git diff --check`: passed.
- Recompiled `src/ivorysql_ora.o` in a configured macOS debug/cassert build: passed.
- Compared Clang-preprocessed source before and after, with `__FILE__` and `__LINE__` normalized: identical.
- Full local build/regression validation is incomplete: the out-of-tree build encountered SQL-generator path and UUID configuration failures. `make installcheck-world` stopped at the libpq symbol check (`libpq.5.dylib: _exit`), outside the changed module.
- `pgindent --check` returns 2 on both the unchanged base file and this patch because of existing formatting elsewhere in the file. Those unrelated formatting changes are left out of this cleanup.

Assisted-by: OpenAI:GPT-6
Percentage of AI-generated code: 100% (the deletion-only patch was prepared with AI assistance)

## Summary by CodeRabbit

- **Bug Fixes**
  - Module initialization no longer rejects loading when the module is not configured through `shared_preload_libraries`.
  - Initialization proceeds directly to setting configuration options and installing required hooks.


hanjianqiao:

Yuky-499:
@hanjianqiao Thanks for the reminder. This PR addresses the existing issue #1923, and the PR description already includes `Fixes #1923`. Would linking to that existing issue be sufficient, or would you prefer a separate issue for this PR?

PR #1979, opened 2026-09-10
UTL_URL.ESCAPE silently truncates data loss when embedding NUL bytes

### LINKED ISSUE #1992: In PR1728，UTL_URL.ESCAPE silently truncates data loss when embedding NUL bytes
## Enhancement
can fix like this:

utl_url.c use VARSIZE_ANY_EXHOR replace strlen
utl_url.sql add 2 tests
utl_url.out add output

enkilee:
fix #1992

fix:
utl_url.c  use VARSIZE_ANY_EXHOR replace strlen
utl_url.sql add 2 tests
utl_url.out add output


hanjianqiao:

PR #1982, opened 2026-09-10
test: cover XML namespace registration boundaries

### LINKED ISSUE #1981: test: cover XML namespace buffer-growth boundaries
## Enhancement

Add regression coverage for the XML namespace registration fix merged in #1825. That change grows the prefix and URI buffers and rejects missing delimiters, but its final diff contains no tests. The existing XML suite covers ordinary namespace declarations without exercising the former 1024-byte allocation boundary.

The proposed tests extend `contrib/ivorysql_ora/sql/ora_xml_functions.sql` and its expected output:

- Check 1023, 1024, 1025, and 4096-byte prefixes and URIs independently, asserting short `EXTRACTVALUE` results.
- Resolve two namespaces after registering a long declaration followed by a short declaration, and in the reverse order.
- Reject missing `=`, missing `:`, unquoted and incomplete values, including a malformed declaration following a valid one.
- Verify successful namespace registration in the same session after those errors.

The tests are implemented and pass in a libxml-enabled Linux build, including all 28 `ivorysql_ora` regressions and the XML suite under AddressSanitizer. As a negative control, reversing the original fix in a disposable build makes the new long-prefix case trigger an AddressSanitizer heap-buffer-overflow in `register_ns_from_csting` (4097-byte write).

This proposal adds test coverage; the namespace implementation is unchanged.

Assisted-by: OpenAI:GPT-6

Yuky-499:
Add regression coverage for the namespace registration fix in #1825. The new cases check 1023, 1024, 1025, and 4096-byte prefixes and URIs independently, verify long and short declarations in both registration orders, and assert controlled errors for malformed declarations followed by a successful query in the same session.

Only the existing XML SQL test and its expected output change. The expected output was captured from a real libxml-enabled run.

Closes #1981

## Validation

Tested on Debian 12 arm64 with libxml2 2.9.14, GCC 12.2 and Clang 14:

- `git diff --check`: passed.
- `make -C contrib/ivorysql_ora oracle-check`: all 28 tests passed.
- `ASAN_OPTIONS=detect_leaks=0 make -C contrib/ivorysql_ora oracle-check ORA_REGRESS=ora_xml_functions`: passed with a Clang AddressSanitizer build (`-O1 -g -fsanitize=address -fno-omit-frame-pointer`, assertions enabled).
- Negative control: reversing commit `71a2e98a513f9df3fb54d8e1a59caa5188c845a0` in a disposable ASan build makes the new long-prefix case report a heap-buffer-overflow in `register_ns_from_csting`, with a 4097-byte write. Restoring the fix makes the same suite pass again.
- `make -j8 installcheck-world USE_MODULE_DB=1`: ran; not fully green. The existing `contrib/xml2` SQL and expected output disagree (including Oracle-only setup lines and an empty-output case absent from the SQL). An initial `initdb` failure was due to Oracle files omitted by `install-world-bin`; after `make install`, its separate installcheck passed all 57 assertions. Core regression (246 tests), isolation (130 tests), subscription and recovery tests passed. Optional features not enabled in this build retain their normal skips.

Assisted-by: OpenAI:GPT-6

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for XML namespace handling with very long prefixes and URIs.
  * Added tests for varying namespace declaration sizes and ordering.
  * Added validation for malformed namespace declarations and confirmation that subsequent valid operations still succeed.


hanjianqiao:
Thanks for your patch!

PR #1984, opened 2026-09-10
Fix TupleDesc attribute array alignment

### LINKED ISSUE #1983: TupleDesc trailing FormData_pg_attribute array can be misaligned after tdhasrowid
﻿## What happened?

`TupleDescData` can place its trailing `FormData_pg_attribute` array at a misaligned address after the ROWID metadata field was added and `CompactAttribute` was reduced to 8 bytes.

On current master `911b66340c437f4b98d21b7fae256c189424444b`, `offsetof(struct TupleDescData, compact_attrs)` is 34 bytes on 64-bit Linux. For `natts = 1`, `TupleDescAttrAddress()` computes `34 + 1 * 8 = 42`. That address is not divisible by 4, while `FormData_pg_attribute` requires at least 4-byte alignment. Casting it to `Form_pg_attribute` and accessing a member is undefined behavior. With alignment/undefined-behavior sanitizer enabled, bootstrap-time tuple descriptor access can terminate `initdb` with a misaligned `FormData_pg_attribute` access.

`initdb` should complete without undefined behavior under an alignment sanitizer. The trailing `FormData_pg_attribute` array must start at an address aligned for that type for every valid `natts`.

Build current `master` on 64-bit Linux with assertions/debug enabled and:

```sh
CFLAGS='-O1 -ggdb -fno-sanitize-recover=all -fsanitize=alignment,undefined'
LDFLAGS='-fsanitize=alignment,undefined'
```

Then run in an empty directory:

```sh
initdb -D data --no-clean --no-sync --auth=trust --lc-messages=C
```

Static layout:

```text
offsetof(TupleDescData, compact_attrs) = 34
sizeof(CompactAttribute) = 8
TupleDescAttrAddress() offset for natts = 1 = 42
42 % alignof(FormData_pg_attribute) = 2
```

Relevant code: `src/include/access/tupdesc.h` (`tdhasrowid`, `TupleDescAttrAddress()`, `TupleDescSize()`) and `src/backend/access/common/tupdesc.c` (`CreateTemplateTupleDesc()`). The issue results from the interaction between ROWID field commit `71933846be8` and `CompactAttribute` size reduction `533ad1c235d`.

## IvorySQL version

master (`911b66340c437f4b98d21b7fae256c189424444b`)

## Compatible mode

Both

## Server encoding

UTF8

Please align the base of the trailing `FormData_pg_attribute` array using the project alignment macros and update `TupleDescAttrAddress()` and `TupleDescSize()` consistently. A compile-time alignment assertion and sanitizer `initdb` regression test would prevent regressions.

I searched existing issues for `tdhasrowid`, `TupleDescAttrAddress`, and alignment failures; no duplicate was found.

## OS version

Ubuntu 22.04 x86_64; GCC 11.4.

- [x] I searched existing issues and found no duplicate

R0CKing666:
﻿## Summary

- Align the trailing `FormData_pg_attribute` array after `CompactAttribute` entries.
- Use the shared aligned offset for descriptor allocation and size calculation.
- Assert the attribute-array alignment in cassert builds.

## Testing

- Reproduced the failure before this change with Ubuntu 22.04, GCC 11.4, `--enable-debug --enable-cassert`, and `-fsanitize=alignment,undefined`: `initdb` failed during bootstrap with a misaligned `FormData_pg_attribute` access in `tupdesc.c`.
- Rebuilt after this change with the same configuration. `initdb` completed successfully and emitted no alignment/UBSan diagnostics.
- `make -C src/test/regress check` could not reach execution because its temporary installation is blocked by the existing `contrib/ivorysql_ora` `gensql.pl` dependency (`ivorysql_ora_merge_sqls` is absent).

Fixes #1983

## Summary by CodeRabbit

* **Refactor**
  * Improved internal memory alignment handling for tuple descriptor data.
  * Standardized calculation of internal data offsets and added alignment validation.
  * Allocation size and runtime behavior remain unchanged.

* **Compatibility**
  * No changes to exported functionality or user-visible behavior.


hanjianqiao:
Thanks for your patch!

PR #2005, opened 2026-09-11
Fix ROWID implicit index lifecycle

### LINKED ISSUE #2004: ROWID implicit indexes are missing after ALTER and duplicated by LIKE
## What happened?

For ordinary ROWID tables, the implicit single-column B-tree index on the ROWID system attribute is not maintained consistently across DDL paths on current `master`:

- `ALTER TABLE ... SET WITH ROWID` adds the ROWID attribute but does not create the implicit ROWID index.
- `CREATE TABLE ... (LIKE source INCLUDING ALL) WITH ROWID` creates the target's implicit index and also clones the source's implicit index, leaving two equivalent default ROWID indexes.

Every ordinary, non-partitioned table with ROWID should have exactly one implicit default B-tree index on `rowid`, regardless of whether ROWID was specified at creation or added with `ALTER TABLE`.

`LIKE ... INCLUDING ALL` should not clone an index that is only the source table's automatically generated default ROWID index. Explicitly customised ROWID indexes (for example `rowid DESC` or a partial index) should continue to be copied by the normal LIKE semantics.

## How can we reproduce it?

```sql
CREATE TABLE rowid_altered (id int) WITHOUT ROWID;
ALTER TABLE rowid_altered SET WITH ROWID;

SELECT c.relhasrowid, count(i.indexrelid) AS rowid_indexes
FROM pg_class AS c
LEFT JOIN pg_index AS i ON i.indrelid = c.oid
                         AND i.indnkeyatts = 1
                         AND i.indkey::smallint[] = ARRAY[-7]::smallint[]
WHERE c.oid = 'rowid_altered'::regclass
GROUP BY c.relhasrowid;
-- actual on current master: t | 0
-- expected:                  t | 1

CREATE TABLE rowid_source (id int) WITH ROWID;
CREATE TABLE rowid_like (LIKE rowid_source INCLUDING ALL) WITH ROWID;

SELECT count(*) AS rowid_indexes
FROM pg_index AS i
WHERE i.indrelid = 'rowid_like'::regclass
  AND i.indnkeyatts = 1
  AND i.indkey::smallint[] = ARRAY[-7]::smallint[];
-- actual on current master: 2
-- expected:                 1
```

A recursive `ALTER TABLE ... SET WITH ROWID` has the same missing-index behavior for each ordinary child table.

## IvorySQL version

`master` at `03b24b1c4688266f738394f443f63d4d80ab6379` (current upstream `master` during reproduction).

## Compatible mode

Both (the catalog/index lifecycle is independent of the compatible mode).

## Server encoding

UTF8.

Reproduced in a debug/cassert build on Ubuntu 22.04 with GCC 11.4. A patch and regression coverage are ready. The intended scope is new DDL only; existing inconsistent tables should not be silently changed during upgrade.

- [x] I confirmed the problem still happens on current `master`.

R0CKing666:
## Summary

Fixes #2004.

- Centralize construction of the implicit default ROWID B-tree `IndexStmt`.
- Queue that index after an `ALTER TABLE ... SET WITH ROWID` rewrite, including ordinary inherited child tables, so every newly enabled ROWID table receives one default index.
- Skip cloning only the source table's default-form ROWID index for `LIKE ... INCLUDING ALL`; explicitly customized ROWID indexes, such as `rowid DESC`, continue to be cloned.
- Add Oracle regression coverage for create, alter, toggling, inherited children, and LIKE with default and customized ROWID indexes.

## Existing tables

This deliberately changes only new DDL paths. It does not create or remove indexes on existing tables during an upgrade. Administrators can use this read-only baseline audit to find ordinary ROWID tables whose count of one-key ROWID indexes is not one:

```sql
WITH rowid_index_counts AS (
    SELECT i.indrelid,
           count(*) FILTER (
               WHERE i.indnkeyatts = 1
                 AND i.indkey::smallint[] = ARRAY[-7]::smallint[]
           ) AS rowid_indexes
    FROM pg_index AS i
    GROUP BY i.indrelid
)
SELECT n.nspname, c.relname, COALESCE(r.rowid_indexes, 0) AS rowid_indexes
FROM pg_class AS c
JOIN pg_namespace AS n ON n.oid = c.relnamespace
LEFT JOIN rowid_index_counts AS r ON r.indrelid = c.oid
WHERE c.relkind = 'r'
  AND c.relhasrowid
  AND NOT c.relispartition
  AND COALESCE(r.rowid_indexes, 0) <> 1
ORDER BY n.nspname, c.relname;
```

## Validation

- `git diff --check`
- CAssert backend compilation: `meson compile -C /tmp/rowid-build-clean postgres`
- Debug/CAssert `initdb` and direct SQL checks covering every newly added regression scenario, including index counts and the `rowid DESC` preservation case.

The repository-wide `oracle-check` was also run. Its current `parallel_schedule` does not include `ora_rowid`, and it reports unrelated pre-existing suite differences; the new `ora_rowid` result block was checked directly against the expected output.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved automatic ROWID index creation when enabling ROWID support or creating ROWID-enabled tables.
  - Ensured ROWID indexes are created correctly after table rewrites.
  - Prevented duplicate default ROWID indexes when using `LIKE ... INCLUDING ALL`.
  - Preserved explicitly customized ROWID indexes during table cloning and inheritance.

- **Tests**
  - Added coverage for ROWID index behavior across table creation, alteration, cloning, inheritance, and custom index definitions.


bigplaice:

PR #2009, opened 2026-09-11
fix: mark sys.regexp_instr/replace/like/count IMMUTABLE

89799969:
## Summary

- Mark every `sys.regexp_instr`, `sys.regexp_replace`, `sys.regexp_like`, and `sys.regexp_count` overload as `IMMUTABLE` (28 overloads), matching the already-correct `sys.regexp_substr` family member.
- These C functions (`ora_regexp_replace`, `ora_regexp_instr`, `ora_regexp_like`, `ora_regexp_like_no_flags`, `ora_regexp_count`) only read their arguments and `PG_GET_COLLATION()`. They do not consult GUCs, transaction state, or the clock.
- This aligns IvorySQL with:
  - **Oracle**: `REGEXP_INSTR` / `REGEXP_REPLACE` / `REGEXP_LIKE` / `REGEXP_COUNT` are all `DETERMINISTIC`.
  - **PostgreSQL core**: `pg_catalog.regexp_replace` / `regexp_match` / `regexp_matches` / `regexp_split_*` are all `IMMUTABLE`.
- `STABLE` blocked constant folding and refused expression indexes over four members of an otherwise-immutable family (`functions in index expression must be marked IMMUTABLE`).
- Existing `PARALLEL SAFE` and `STRICT` attributes are unchanged. Other `STABLE` functions in the same file (for example the NLS-dependent `sys.lengthc(datetime)` overloads) are intentionally left alone.

## Testing

Local verification is limited: this environment does not have a full IvorySQL build toolchain, so `make -C contrib/ivorysql_ora oracle-check` / `make installcheck-world` were **not** run. CI is expected to execute the suite.

Focused regression coverage added in `ora_character_datatype_functions`:

1. Catalog check that no `sys.regexp_*` overload (`regexp_instr` / `regexp_replace` / `regexp_like` / `regexp_count` / `regexp_substr`) remains non-`IMMUTABLE`.
2. `CREATE INDEX` acceptance for `sys.regexp_like(...)` and `sys.regexp_instr(...)` (both previously `STABLE`, so previously refused as index expressions).

`git diff --check` is clean.

## Related

Fixes #2008

Assisted-by: MiMo (Xiaomi MiMo)
Percentage of AI-generated code: 100

## Summary by CodeRabbit

* **Improvements**
  * Regular expression functions now provide consistent immutable behavior across supported overloads.
  * Expression indexes can be created using `regexp_like`, `regexp_instr`, and `regexp_replace`.

* **Tests**
[...]

PR #2017, opened 2026-09-12
fix: report correct match positions from ora regexp_instr for short matches

### LINKED ISSUE #2016: sys.regexp_instr returns 0 for single-character matches, empty matches, and is off by one for ret_opt with subexpr
### IvorySQL Version

`master` @ `03b24b1c468` (Oracle compatible mode)

### Current Behavior

```sql
SELECT regexp_instr('abcabc','c',1);            -- 0   (should be 3)
SELECT regexp_instr('abcabc','c',1,2);          -- 0   (should be 6)
SELECT regexp_instr('abcabc','c',4,1);          -- 0   (should be 6)
SELECT regexp_instr('abc','c*',1,1);            -- 0   (should be 1)
SELECT regexp_instr('abcabc','(a)(b)',1,1,1,'i',2); -- 4   (should be 3)
```

With 3+ arguments the call binds to `sys.regexp_instr` (identical `(text,text,integer)` signature
is preferred over `pg_catalog` in Oracle mode), whose result builder
`ora_build_regexp_instr_matches_result()` (src/backend/utils/adt/regexp.c) decrements the
whole-match end before comparing it with the start, so every one-character match is treated as
degenerate. The subexpression branch returns `eo + 2` for ret_opt = 1, one position too far.

Note: #1656 reported the single-character case but was closed as non-reproducible; that
verification tested `pg_catalog.regexp_instr` explicitly, which bypasses the Oracle-mode binding
to `sys.regexp_instr`. The sys path does reproduce it on master (queries above).

### Expected Behavior (Oracle 23ai Free, gvenzl/oracle-free:23-slim)

```sql
SQL> SELECT regexp_instr('abcabc','c',1,1) FROM dual;  -- 3
SQL> SELECT regexp_instr('abcabc','c',1,2) FROM dual;  -- 6
SQL> SELECT regexp_instr('abc','c*',1,1) FROM dual;    -- 1
SQL> SELECT regexp_instr('abcabc','(a)(b)',1,1,1,'i',2) FROM dual; -- 3
```

### Steps to reproduce

Run the SELECTs in Oracle compatible mode and compare with the Oracle session above.

### Impact

Every one-character regexp pattern - e.g. a delimiter lookup like regexp_instr(str, ',') - returns 0 (no match) instead of the position, silently breaking string parsing, filtering and joins that use the function with 3+ arguments in Oracle mode.

### Acceptance criteria

- The statements shown under Current Behavior return the Oracle values above in oracle compatible mode.
- The oracle_test regression suites pass with the new expectations.

### Corresponding PR

#2017

### Environment

Built from source (gcc, WSL2); Oracle values from a live Oracle 23ai Free container.
- Compatible mode: `oracle` (`SHOW ivorysql.compatible_mode`)
- Server encoding: `UTF8` (`SHOW server_encoding`)
- OS: Linux (WSL2 kernel 6.6.87.2-microsoft-standard-WSL2, x86_64)

unbridled-41:
## Problem

In Oracle compatible mode, `regexp_instr()` returns 0 ("no match") for every **single-character**
match whenever the call has 3 or more arguments, and it mishandles empty matches and
`ret_opt` with subexpressions.

### Evidence (master @ 03b24b1c468, oracle mode)

```sql
SELECT regexp_instr('abcabc','c',1);                 -- 0   (Oracle: 3)
SELECT regexp_instr('abcabc','c',1,2);               -- 0   (Oracle: 6)
SELECT regexp_instr('abcabc','c',4,1);               -- 0   (Oracle: 6)
SELECT regexp_instr('abcabc','c',1,2,1);             -- 0   (Oracle: 7)
SELECT regexp_instr('abc','c*',1,1);                 -- 0   (Oracle: 1)
SELECT regexp_instr('abc','c*',1,1,1);               -- 0   (Oracle: 1)
SELECT regexp_instr('abcabc','(a)(b)',1,1,1,'i',2);  -- 4   (Oracle: 3)
```

The two-argument call (`regexp_instr('abcabc','c')`) happens to bind `pg_catalog.regexp_instr`
and returns 3 — which is why the problem is easy to miss. With 3+ arguments the call binds
`sys.regexp_instr` (identical `(text,text,integer)` signature preferred over `pg_catalog` in
Oracle mode; confirmed via `pg_typeof(regexp_instr('abcabc','c',1))` = `varchar2`).

This was previously reported in #1656 / #1657, but the issue was closed as "does not reproduce"
with a verification that explicitly queried `pg_catalog.regexp_instr`, bypassing the Oracle-mode
binding to `sys.regexp_instr`. On the `sys` path the bug reproduces on master (queries above),
and #1657 was withdrawn without being merged.

## Root cause / Fix

`ora_build_regexp_instr_matches_result()` in `src/backend/utils/adt/regexp.c`:

- the whole-match branch decremented the exclusive match end
  (`eo = match_locs[...] - 1`) and then treated `so == eo` as "degenerate, no match" —
  so **every 1-character match** returned false (and an empty match at position 0 hit it too);
- the ret_opt=1 result was computed from the inclusive end there (`eo + 2` with the decremented
  value), while the subexpression branch added 2 to the *exclusive* end — one position too far.

The fix keeps `so`/`eo` as the zero-based exclusive bounds in both branches and returns
`so + 1` (ret_opt 0) / `eo + 1` (ret_opt 1). Verified against Oracle 23ai:

| query | before | after | Oracle 23ai |
|---|---|---|---|
| `regexp_instr('abcabc','c',1,1)` | 0 | 3 | 3 |
| `regexp_instr('abcabc','c',1,2)` | 0 | 6 | 6 |
| `regexp_instr('abcabc','c',4,1)` | 0 | 6 | 6 |
| `regexp_instr('abcabc','c',1,2,1)` | 0 | 7 | 7 |
| `regexp_instr('abc','c*',1,1)` | 0 | 1 | 1 |
| `regexp_instr('abc','c*',1,1,1)` | 0 | 1 | 1 |
| `regexp_instr('xabc','c*',1,2)` | 0 | 2 | 2 |
| `regexp_instr('abcabc','(a)(b)',1,1,1,'i',2)` | 4 | 3 | 3 |
| `regexp_instr('abcabc','(a)(b)',1,1,0,'i',2)` | 2 | 2 | 2 |
| `regexp_instr('abcabc','cc',1,1)` (no match) | 0 | 0 | 0 |

`regexp_substr` is unaffected (it uses a separate result builder); multi-character matches and
ret_opt=1 without subexpr already returned correct values and remain unchanged.

## Priority / scoring

PRIORITY 80 = 影响 32 (a very common function silently reports "no match" for single-character
patterns — e.g. delimiter lookup — and off-by-one positions corrupt downstream string parsing) +
波及 13 (one shared result builder for the whole sys.regexp_instr family) + 可复现性 20
(deterministic one-liners, verified against real Oracle) + 维护价值 15. FIX_CONFIDENCE 85
(small, contained change; every behaviour cross-checked on Oracle 23ai).

## Tests

- `contrib/ivorysql_ora/sql/ora_character_datatype_functions.sql`: 12 new regression cases
  (single-character matches, empty matches, occurrence/start/ret_opt combinations,
  subexpression positions, no-match controls), all values cross-checked against a live
  Oracle 23ai Free instance; `expected/ora_character_datatype_functions.out` updated.
- `src/oracle_test/regress/expected/strings.out` and `strings_1.out`: the six
  `ret_opt = 1` subexpression expectations encoded the off-by-one (5/10/8/10 → 4/9/7/9);
  verified against Oracle 23ai.
- `make oracle-check` (contrib, 29 suites): **all 29 passed**.
- `make -C src/oracle_test/regress oracle-check` (256 tests incl. strings): **all 256 passed**.

Fixes #2016

## Risk

Low-medium: the change alters only the position computation of the Oracle-mode
`sys.regexp_instr` family. Existing expectations that encoded the buggy values were updated
(strings.out / strings_1.out). Core `pg_catalog.regexp_instr` and `regexp_substr` /
`regexp_count` / `regexp_like` paths are untouched. The return type remains `varchar2` as today
(Oracle returns NUMBER — changing the declared return type is out of scope here and would touch
the same declaration lines as open PR #2009).

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected `regexp_instr` position reporting for whole matches and subexpressions when requesting match-end positions.
  - Fixed handling of single-character matches, empty matches, occurrence and start-position options, and return options involving subexpressions.
  - Results now consistently report positions according to the documented matching behavior, including degenerate and zero-length matches.

- **Tests**
  - Added regression coverage for the updated `regexp_instr` scenarios and adjusted expected results accordingly.


unbridled-41:
Pre-fix failure evidence, captured as a reproducible regression run
====================================================================

Method: master code (`03b24b1c468`, binary without this fix) + this PR regression files, `make oracle-check ORA_REGRESS=ora_character_datatype_functions`. All 12 new cases fail. Actual pre-fix values (`-` = Oracle-correct expected, `+` = pre-fix result):

```diff
- 3      regexp_instr('abcabc','c',1,1)
+ 0
- 6      regexp_instr('abcabc','c',1,2)
+ 0
- 6      regexp_instr('abcabc','c',4,1)
+ 0
- 7      regexp_instr('abcabc','c',1,2,1)
+ 0
- 1      regexp_instr('abc','c*',1,1)
+ 0
- 1      regexp_instr('abc','c*',1,1,1)
+ 0
- 2      regexp_instr('xabc','c*',1,2)
+ 0
- 3      regexp_instr('abcabc','(a)(b)',1,1,1,'i',2)
+ 4
```

(controls already passing pre-fix and unchanged post-fix: no-match → 0, two-character occurrences, ret_opt without subexpr.)

Post-fix, same files: 1/1 passed; full `make oracle-check` 29/29; `make -C src/oracle_test/regress oracle-check` 256/256 (strings.out / strings_1.out ret_opt expectations aligned with Oracle).

pg mode: `make installcheck REGRESS=strings` passed on the final build - `pg_catalog.regexp_instr` is untouched by this change, and the fixed functions only bind in oracle mode.

Fixes #2016

bigplaice:

PR #2025, opened 2026-09-12
test: cover sys.uid (incl. UID pseudo-column) and to_clob/to_blob

### LINKED ISSUE #2024: test: no coverage for sys.uid (incl. UID pseudo-column) and to_clob/to_blob
### Summary

Coverage analysis of contrib/ivorysql_ora shows three user-facing functions
with no regression coverage in sql/*.sql:

- sys.uid() -- never called (0 occurrences); the oracle-mode UID
  pseudo-column is likewise untested, even though oracle parser mode binds
  UID to sys.uid()
- to_clob(varchar2) / to_blob(varchar2) -- only mentioned inside comments,
  never actually invoked

### Proposed coverage

- sys.uid(): assert that sys.uid() equals the current role's OID
  (comparing instead of printing keeps the result stable across
  environments); in oracle mode, assert UID (pseudo-column) matches
  current_user::regrole::oid and agrees with sys.uid()
- to_clob()/to_blob(): return values and lengths for ascii and multibyte
  input; oracle-mode empty-string-to-NULL behavior (functions are STRICT)

Verified: contrib/ivorysql_ora oracle-check 28/28 (on master).

xuhuicai:
Closes #2024.

Adds regression coverage for three implemented but untested functions:

- sys.uid(): compared against current_user::regrole::oid (stable
  output); oracle-mode UID pseudo-column asserted equal to the same and
  to sys.uid().
- to_clob()/to_blob(): lengths for ascii and multibyte input; STRICT
  empty-string-to-NULL behavior in oracle mode.

No production code changes; tests + expected output only.

Verified: make -C contrib/ivorysql_ora oracle-check -> 28/28.

Assisted-by: Cursor

## Summary by CodeRabbit

- **Tests**
  - Added coverage for text-to-LOB conversions, including ASCII, multibyte text, and Oracle-compatible empty-string handling.
  - Added validation that `sys.uid()` and Oracle-compatible `UID` return the current user’s identifier consistently.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_character_datatype_functions.out
@@ -4074,3 +4074,38 @@ select to_single_byte(NULL);
  
 (1 row)
 
+-- Tests for to_clob() and to_blob() (text -> clob/blob domain conversion)
+SELECT length(to_clob('abc')) AS clob_len;
+ clob_len 
+----------
+        3
+(1 row)
+
+SELECT length(to_blob('abc')) AS blob_len;
+ blob_len 
+----------
+        3
+(1 row)
+
+-- multibyte input: length is in characters
+SELECT length(to_clob('中文')) AS clob_mb_len;
+ clob_mb_len 
+-------------
+           2
+(1 row)
+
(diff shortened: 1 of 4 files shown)


bigplaice:

PR #2027, opened 2026-09-12
test: cover Oracle date/time functions and timezone conversions

### LINKED ISSUE #2026: test: weak coverage for date/time functions (sysdate, systimestamp, timezone conversions)
### Summary

Coverage analysis of contrib/ivorysql_ora/sql/*.sql shows the Oracle
date/time functions are barely exercised:

- sysdate / systimestamp -- no actual calls (only mentioned in comments);
  current_date/current_timestamp/localtimestamp appear once, in a CREATE
  TABLE AS SELECT that never checks the values
- sessiontimezone() -- called once
- from_tz() -- a single offset tested
- sys_extract_utc() -- two inputs, no NULL case

### Proposed coverage (stable output, no wall-clock values)

- sysdate()/systimestamp(): IS NOT NULL assertions
- pg_typeof assertions for current_timestamp, current_timestamp(3),
  localtimestamp and systimestamp
- sessiontimezone() under SET TIME ZONE 'America/New_York'
- from_tz(): +14:00 and -12:00 under a fixed UTC session (stable output),
  plus the out-of-range '99:99' error
- sys_extract_utc(): NULL propagation and an offset input converting to UTC

All cases pin the session time zone or assert rather than print time
values, so the expected output does not depend on the host time zone.

Verified: contrib/ivorysql_ora oracle-check 28/28 (on master).

xuhuicai:
Closes #2026.

Adds regression coverage for the Oracle date/time functions that had
none or one call only:

- sysdate()/systimestamp() non-NULL assertions (their only previous
  mentions were in comments);
- pg_typeof checks for current_timestamp (incl. (3) precision form),
  localtimestamp and systimestamp;
- sessiontimezone() under SET TIME ZONE;
- from_tz() offsets +14:00/-12:00 under a fixed UTC session plus the
  out-of-range '99:99' error;
- sys_extract_utc() NULL propagation and offset conversion.

No production code changes; tests + expected output only. All outputs
are stable across host time zones.

Verified: make -C contrib/ivorysql_ora oracle-check -> 28/28.

Assisted-by: Cursor

## Summary by CodeRabbit

* **Tests**
  * Added coverage for Oracle-compatible datetime functions, including system and current timestamp values and their data types.
  * Added validation for session time zone behavior and time zone conversion, including invalid offset handling.
  * Added checks for UTC extraction, NULL propagation, and offset conversion.

### DIFF
--- a/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
+++ b/contrib/ivorysql_ora/expected/ora_datetime_datatype_functions.out
@@ -2076,3 +2076,66 @@ drop table ds_tb;
 reset NLS_TIMESTAMP_FORMAT;
 reset NLS_TIMESTAMP_TZ_FORMAT;
 reset NLS_DATE_FORMAT;
+-- Tests for sysdate / systimestamp / current_timestamp / localtimestamp
+SET ivorysql.compatible_mode = oracle;
+SELECT sysdate IS NOT NULL AS sysdate_not_null FROM dual;
+ sysdate_not_null 
+------------------
+ t
+(1 row)
+
+SELECT systimestamp IS NOT NULL AS systimestamp_not_null FROM dual;
+ systimestamp_not_null 
+-----------------------
+ t
+(1 row)
+
+SELECT pg_typeof(current_timestamp) = 'timestamp with time zone'::regtype AS cts_type,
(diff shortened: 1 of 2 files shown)


bigplaice:

PR #1881, opened 2026-09-05
feat(ivorysql_ora): add Oracle-compatible DBMS_CRYPTO package

elephone-184:
Close #1962

## Summary

This PR implements the Oracle-compatible `DBMS_CRYPTO` package in the `ivorysql_ora` extension.

`DBMS_CRYPTO` is the standard Oracle package providing cryptographic algorithms, message authentication codes (MAC), and random data generation.

### Implemented Subprograms & Constants:
1. **Hash and MAC Constants**:
   - `HASH_MD4` (1), `HASH_MD5` (2), `HASH_SH1` (3), `HASH_SH256` (4), `HASH_SH384` (5), `HASH_SH512` (6)
   - `HMAC_MD5` (1), `HMAC_SH1` (2), `HMAC_SH256` (3), `HMAC_SH384` (4), `HMAC_SH512` (5)
2. **`HASH(src, typ) RETURN RAW`**:
   - Overloaded for `RAW` and `VARCHAR2` inputs.
   - Computes standard message digests using MD5, SHA1, SHA256, SHA384, or SHA512.
3. **`MAC(src, typ, key) RETURN RAW`**:
   - Overloaded for `RAW` and `VARCHAR2` inputs.
   - Generates keyed HMAC digests using standard cryptographic algorithms.
4. **`RANDOMBYTES(number_bytes IN INTEGER) RETURN RAW`**:
   - Generates a sequence of random bytes up to 32767 bytes.
5. **`RANDOMINTEGER() RETURN INTEGER`**:
   - Generates a random 32-bit signed integer.

### Security & Packaging:
- **Authorization & Security (CWE-862)**: Execution privileges on internal C resolver functions in `sys` are revoked from `PUBLIC`.
- Built on PostgreSQL core cryptographic primitives (`common/cryptohash.h` and `common/hmac.h`).
- Integrated into `ivorysql_ora_merge_sqls`, `Makefile`, and `meson.build`.

## Test plan

- [x] Verify all 11 package constants.
- [x] Verify `HASH` across MD5, SHA-1, SHA-256, SHA-384, and SHA-512 with known digest vectors.
- [x] Verify `HASH` with both `VARCHAR2` and `RAW` inputs.
- [x] Verify `MAC` with HMAC-MD5 and HMAC-SHA256.
- [x] Verify `RANDOMBYTES` and `RANDOMINTEGER` output formats and bounds.
- [x] Verify NULL argument handling.
- [x] Verify error handling on invalid/unsupported algorithm IDs.
[...]

Issue #1025, opened 2025-12-05
oracle compatibility feature: DBMS_CRYPTO


hs-liuxh:
## Feature Request
oracle compatibility feature: DBMS_CRYPTO
**Describe the feature you'd like to propose:**
DBMS_CRYPTO performs cryptographic operations such as encryption, decryption, hashing, and Message Authentication Code (MAC), offering robust and flexible security features.

PR #2030, opened 2026-09-12
Add Oracle-compatible STANDARD_HASH function

muzimu217:
## Summary

- add `STANDARD_HASH(expr [, method ])` to the `ivorysql_ora` extension, Oracle's SQL-level message digest function, returning `RAW`
- built on the PostgreSQL core cryptohash infrastructure (`src/common/cryptohash.h`), the same primitives the DBMS_CRYPTO proposal (#1881) uses, so no new dependency is introduced
- implemented in `src/builtin_functions/misc_functions.c` next to the other datatype-independent built-ins (`VSIZE`, `UID`), declared in `builtin_functions--1.0.sql` as `sys.standard_hash(text, text DEFAULT 'SHA1') RETURNS sys.raw`
- new regression test `sql/ora_standard_hash.sql` registered in `ORA_REGRESS`

This is an independent feature PR: no existing issue or PR tracks the `STANDARD_HASH` *function* (the open DBMS_CRYPTO items #1025/#1962/#1881 cover the package interface only).

## Oracle 23ai behavior matrix (verified on a real 23ai instance)

| input | MD5 | SHA1 | SHA256 | SHA384 | SHA512 |
|---|---|---|---|---|---|
| `'abc'` | `900150983CD24FB0D6963F7D28E17F72` | `A9993E364706816ABA3E25717850C26C9CD0D89D` | `BA7816BF...015AD` | `CB00753F...825A7` | `DDAF35A1...4CA49F` |
| `''` (empty) | `D41D8CD98F00B204E9800998ECF8427E` | `DA39A3EE5E6B4B0D3255BFEF95601890AFD80709` | digest of 0 bytes | digest of 0 bytes | digest of 0 bytes |
| `NULL` (varchar2/number/date) | `D41D8CD98F00B204E9800998ECF8427E` | digest of 0 bytes | ... | ... | ... |
| `'中'` (AL32UTF8) | `AED1DFBC31703955E64806B799B67645` (= md5 of UTF-8 bytes `E4 B8 AD`) | | | | |

Semantics reproduced:

- **default method is SHA1** (`STANDARD_HASH('abc')` = `STANDARD_HASH('abc','SHA1')`)
- **NULL input is not propagated**: Oracle digests a zero-byte input, so `STANDARD_HASH(NULL,'MD5')` = `STANDARD_HASH('','MD5')` = `md5('')`. The SQL declaration is therefore *not* `STRICT`; NULL is handled in the C function.
- **multibyte input digests the database-encoding bytes** (`'中文'` SHA256 = `72726D88...` = digest of UTF-8 bytes), matching an AL32UTF8 database
- **the method name is case-sensitive**: only the exact spellings `'MD5'`, `'SHA1'`, `'SHA256'`, `'SHA384'`, `'SHA512'` are accepted; `'md5'`, `'Sha256'`, `'MD5 '` (trailing blank) and `NULL` all raise `ORA-03052: invalid argument for STANDARD_HASH function` (same message text)
- digest lengths: 16/20/32/48/64 bytes (RAW hex display uppercase on Oracle; byte values identical)

## Scope note

Oracle also accepts non-character scalar types, but it digests them in their proprietary *internal binary* format (e.g. `STANDARD_HASH(123)` = digest of the three bytes `C2 02 18`, `STANDARD_HASH(-1)` = digest of `3E 64 66`), not their textual form. Reproducing that for `NUMBER`/`DATE` is a separate effort; rather than silently returning a digest that would *not* match Oracle, this PR supports character input and requires an explicit cast for other types. The limitation is documented in the function header.

## Test plan

- [x] `make oracle-check` in `contrib/ivorysql_ora`: **30/30 tests passed** under `C.utf8`, `en_US.utf8` and `zh_CN.utf8`
- [x] every digest in the regression test cross-checked against the Oracle 23ai instance output (values above)
- [x] error cases covered: five invalid-method spellings
- [x] DCO signed off

## Summary by CodeRabbit

* **New Features**
  * Added the Oracle-compatible `STANDARD_HASH` function.
  * Supports MD5, SHA1, SHA256, SHA384, and SHA512 algorithms, with SHA1 as the default.
  * Returns digest results as RAW values and supports text, multibyte, empty, and long inputs.
  * Invalid or incorrectly cased algorithm names now produce the corresponding Oracle-compatible error.

* **Tests**
  * Added regression coverage for supported algorithms, digest lengths, NULL handling, byte access, and invalid arguments.


bigplaice:

PR #2035, opened 2026-09-13
Add SYS.V$VERSION version banner view

### LINKED ISSUE #2054: Add SYS.V$VERSION compatibility view
Oracle's `V$VERSION` is the standard first-stop introspection view for version
information: migration smoke tests, third-party tool detection, and support
scripts routinely query `BANNER` from it before anything else. IvorySQL
currently has V$PARAMETER, V$PROCESS, and V$SESSION in `sysview`, but no
`V$VERSION`, so tools that probe it at connect time see an
"relation does not exist" error.

Proposal: add `SYS.V$VERSION` as a pure SQL view following the existing
`sysview` pattern. Oracle 19c exposes `BANNER`, `BANNER_FULL`, `BANNER_LEGACY`,
and `CON_ID`; a minimal first version can provide `BANNER` (derived from
`version()`, not hardcoded) plus `CON_ID`, with the column layout documented in
the PR.

Implementation: PR #2035.

muzimu217:
Hello,

I propose this patch to add `SYS.V$VERSION`, the version banner view. In Oracle it is the standard entry point for version introspection, and migration smoke tests as well as third-party tools commonly query it. IvorySQL currently has no counterpart.

The view follows Oracle's column layout (captured on a live Oracle Database 26ai Free instance):

```
BANNER        VARCHAR2(129)
BANNER_FULL   VARCHAR2(258)
BANNER_LEGACY VARCHAR2(129)
CON_ID        NUMBER
```

Example output:

```
SQL> SELECT banner FROM v$version;
          banner
--------------------------------------
IvorySQL 5beta1 (PostgreSQL 19devel)
(1 row)
```

Design notes:

- The version strings are derived from `version()` and `current_setting('server_version')`, so the view follows the built release instead of hardcoding them.
- `BANNER_FULL` is the banner followed by a `Version <ivorysql>` line, matching the shape observed on Oracle Free. `BANNER_LEGACY` equals `BANNER` because there is no legacy IvorySQL branding to report separately.
- The banner deliberately reports IvorySQL branding: it does not start with `'Oracle'`, so tools probing `V$VERSION` with `LIKE 'Oracle%'` will not match, by design.
- `CON_ID` is 0 because IvorySQL has no multitenant container layout.
- In Oracle every user can read `V$VERSION`, and this view exposes no session data (unlike `V$SESSION`/`V$PROCESS`/`V$PARAMETER`), so `SELECT` is granted to `PUBLIC`.

Regression coverage is added to `ora_sysview` (`\d SYS.V$VERSION` plus a full column select). `make oracle-check` passes all 29 tests under ENCODING=UTF8, LC_ALL=C and LC_ALL=C.utf8.

Thanks and regards

Signed-off-by: muzimu217 <muzimu217@users.noreply.github.com>

## Summary by CodeRabbit

* **New Features**
  * Added the Oracle-compatible `SYS.V$VERSION` view.
  * Exposes version banner details, full and legacy banner values, and the container ID.
  * View data is available for querying by all users.

* **Tests**
  * Added coverage verifying the view’s columns and returned IvorySQL version information.

Fixes #2054


muzimu217:
Thanks for the suggestion — I checked the extension's versioning convention and I believe no upgrade script is needed here.

`ivorysql_ora` has a single version: `default_version = '1.0'`, and the tree contains only `--1.0.sql` files (builtin_functions, datatype, xml_functions, sysview, and all packages) with no `--1.0--*` upgrade scripts anywhere. Every existing view in `sysview--1.0.sql` (V$PARAMETER, V$PROCESS, V$SESSION) was added the same way this PR does. Since there is no released 1.0-without-V$VERSION state to upgrade from — any installation of the extension runs the 1.0 script and gets the view — an upgrade path would have nothing to upgrade. One would only become meaningful if the extension version is ever bumped for a future change.

bigplaice:
Please create an issue and associate it with this PR, to satisfy the pr-governance.

muzimu217:
Thanks for the reminder! Created #2054 for this and linked it in the PR description (`Fixes #2054`) to satisfy the pr-governance check.

PR #2039, opened 2026-09-13
fix(oracle): keep the sign when rounding negative NUMTOYMINTERVAL values

### LINKED ISSUE #2038: [Bug] NUMTOYMINTERVAL loses the sign of a negative fractional month
### IvorySQL Version

`master` at `03b24b1c46` (unchanged since; the defect is present on current master).

### Current Behavior

`NUMTOYMINTERVAL()` drops the sign of a negative fractional month, so a caller
that asks for a negative interval gets a positive one.

`contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c`, in
`numtoyminterval()` (the `/* negative */` block):

```c
	/* negative */
	if (interval_val < 0)
	{
		if (interval_val > -0.5 && interval_val < 0)
			interval_val = 0;

		if (interval_val <= -0.5 && interval_val > -1)
			interval_val = 1;      /* <-- not negative */
	}
```

The value that ends up in `interval->month` is `+1`, i.e. an interval that
points *away* from the requested direction:

```sql
select numtoyminterval(-0.7, 'month') from dual;
-- +000000000-01      (one month AFTER the epoch)
select numtoyminterval( 0.7, 'month') from dual;
-- +000000000-01      (identical result)
```

The two arguments are indistinguishable after the call, and
`sysdate + numtoyminterval(-0.7,'month')` moves forward instead of backward.

### Expected behavior

Round away from zero, keeping the sign, exactly like the rest of the function
does: the block immediately below negates the value before the "round to first
place after the decimal point" step and negates it back afterwards, which only
makes sense if a negative argument is still negative at that point.

```sql
select numtoyminterval(-0.7, 'month') from dual;
-- -000000000-01
```

Note that *no* plausible rounding rule turns `-0.7` into `+1`: round, floor,
ceil and truncate (toward zero) all give `-1` or `0`.

### Additional context

The same defect was reported in #1630 together with three unrelated boundary
items; that issue and its PR were withdrawn rather than fixed, with the note
that a focused PR could be resubmitted. This report is that focused item: the
sign fix plus a regression test, nothing else. A PR follows.

btlqql:
## What and why

`NUMTOYMINTERVAL()` rounds a fractional month, but it loses the sign of a
negative fraction, so asking for a negative interval returns a positive one.

## The defect

`contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c`,
`numtoyminterval()`:

```c
	/* negative */
	if (interval_val < 0)
	{
		if (interval_val > -0.5 && interval_val < 0)
			interval_val = 0;

		if (interval_val <= -0.5 && interval_val > -1)
			interval_val = 1;      /* should be -1 */
	}
```

`interval_val = 1` is a positive month count, so `+1` is what reaches
`interval->month`:

```sql
select numtoyminterval(-0.7, 'month') from dual;  -- +000000000-01   (wrong)
select numtoyminterval( 0.7, 'month') from dual;  -- +000000000-01
```

Both arguments produce the *same* interval, and
`sysdate + numtoyminterval(-0.7,'month')` moves forward instead of backward.
No rounding rule turns `-0.7` into `+1` - round, floor, ceil and truncate all
give `-1` or `0` - and the rounding block immediately below negates the value
before rounding and negates it back afterwards, which only makes sense if a
negative argument is still negative at that point.

## The fix

`interval_val = -1;` - one character, in the branch that already exists for
this purpose. The negative case then rounds exactly like its positive
counterpart instead of collapsing onto it.

## The test

`contrib/ivorysql_ora/sql/ora_datetime_datatype_functions.sql`: the existing
`NUMTOYMINTERVAL` block of `datetest` (a single `interval year(3) to month`
column) gains the negative fraction and its positive mirror, so the two rows
must differ in sign:

```sql
insert into datetest values (NUMTOYMINTERVAL(-0.7,'month'));
insert into datetest values (NUMTOYMINTERVAL(0.7,'month'));
```

`expected/ora_datetime_datatype_functions.out` is updated to match:
`-000-01` next to `+000-01` (precision 3 because of the column type; the
`+123-00` / `+019-06` rows and the layout around them are untouched). The
suite is `make -C contrib/ivorysql_ora oracle-check`; the `.sql`/`.out` pair
is part of `ORA_REGRESS`.

## What I could not verify

There is no C toolchain, no `make`, no `bison`/`flex`, and no container or WSL
distro on the machine I worked on, so I could not build IvorySQL or execute
this test. The expected output was derived by reading the implementation:
`numtoyminterval()` (the fixed branch), `EncodeYminterval()`
(`yminterval.c`, which prints `-` then `sprintf(cp, "%0*d-%02d", precision, ...)`
of the negated fields, hence `-000-01` for `month = -1` at precision 3) and
`AdjustYmintervalForTypmod()` (which only validates the leading precision, so
`-1` month in a `year(3) to month` column is accepted). The surrounding
header/separator lines of that block were reused verbatim from the existing
expected file. Please let CI run `oracle_regression`/`pg_regression` to
confirm.

Fixes #2038

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected interval calculations for negative fractional month values.
  * Negative inputs such as `-0.7` months now produce correctly signed results instead of positive intervals.

* **Tests**
  * Expanded date and time function coverage for positive and negative fractional month values.


bigplaice:

muzimu217:
Hi @btlqql, thanks for the focused resubmission — I compared the patch against
current master, re-derived the arithmetic independently, and everything checks
out on my side. Notes below: an independent confirmation of the defect and the
fix, plus two non-blocking observations for you to weigh.

## 1. The defect confirmed independently (arithmetic-level)

I extracted the three rounding blocks of `numtoyminterval()` verbatim from
master (`datetime_datatype_functions.c`, the `/* positive */`, `/* negative */`
and `/* round to first place */` sections) into a standalone C program — same
code, with only the `palloc`/`PG_RETURN` plumbing replaced by printing the
final `month` value — and compiled it with UBSan enabled. Full transparency on
method: I did not have a server build at hand, so this is a source-exact
arithmetic harness, not a live `SELECT`. Since the C function's only float8
input is `PG_GETARG_FLOAT8(0)` and the only sink is `result->month = interval_val`
(an int32 field), the harness output should match what the backend computes.

Results for the affected inputs (master vs. your patch):

```
input                        master   patched
-0.7 month (issue repro)         +1       -1
 0.7 month (issue repro)         +1       +1
-0.5 month (exact half)          +1       -1
-0.99 month                      +1       -1
-0.07 year  (x12 = -0.84 m)      +1       -1
```

So on master every input in `(-1, -0.5]` — including via the `'year'` unit after
the `* 12` — collapses to **+1**, exactly as your issue describes, and no
plausible rounding rule maps `-0.7` to `+1`. The one-character change to `-1`
fixes all of them, and the `(-0.5, 0) → 0` branch and the `|x| >= 1` path (which
negates before the `((int)(x*10)+5)/10` step) are untouched and were already
sign-safe: `-1.5 → -2`, `-1.7 → -2` both before and after the patch, matching the
round-half-away-from-zero the rest of the function implements.

The new expected rows are self-consistent with that: `-000-01` for `-0.7` and
`+000-01` for `0.7` are exactly what the patched code produces, and they also
match the one detailed write-up of Oracle's behavior I could find (Sean
Stuber's intervals article: fractional months are "rounded to the nearest
integer"; his `NUMTOYMINTERVAL(33.7,'month') → +000000002-10` example is
reproduced by the shared `|x| >= 1` rounding step, which is a nice consistency
check for the code path you're building on).

## 2. Observation: rounding is asymmetric around zero, and the positive side may be the odd one

With the fix applied, the two ±branches no longer mirror each other, because
they never did:

- positive: `[0.01, 1) → 1` — anything from 0.01 up rounds **up** to 1;
- negative: `(-0.5, 0) → 0`, `[-1, -0.5] → -1` — only past the half-way point.

Concretely, after this patch: `NUMTOYMINTERVAL(0.4,'month') → +000-01` but
`NUMTOYMINTERVAL(-0.4,'month') → +000-00`, and likewise `0.01 → +1` vs `-0.01 → -0`.
Under the "rounded to the nearest integer" semantics quoted above, one would
expect `0.4 → 0` — i.e. the pre-existing positive threshold may itself be the
compatibility outlier, and this patch (rightly, in my view) doesn't try to
re-litigate it.

I don't have an Oracle instance available, so I can't settle ±0.4 / ±0.01
first-hand. If you can run four probes on a real Oracle
(`NUMTOYMINTERVAL(±0.4,'month')`, `NUMTOYMINTERVAL(±0.01,'month')`), pinning
whichever side is correct with one more fixture pair would immunize this area
against future "simplification". Entirely optional and non-blocking.

## 3. Adjacent edges in the same function (future pass, not for this PR)

While in the neighborhood I noticed the function has no magnitude/NaN guard on
`interval_val`, and the comparisons silently let both through:

- `NaN`: every comparison is false, so `NaN` reaches `result->month =
  interval_val;` — a float8→int32 cast of NaN, which is undefined behavior
  (my UBSan build traps on exactly that line; the arm64 backend happens to
  produce 0).
- `1e30`: sails through the same way, and on my machine the int32 cast of the
  `* 10` intermediate yields `month = -214748364` — a large *negative* month
  count returned silently for a large positive request.

Both are pre-existing, unrelated to your one-line fix, and I'd understand
completely if they're out of scope here; mentioning only in case they're worth
a follow-up issue.

## 4. Small test suggestion

Your fixtures cover the `'month'` unit. Since the `'year'` unit funnels through
the same branch after `* MONTHS_PER_YEAR`, one extra row would pin that path
too — e.g. `NUMTOYMINTERVAL(-0.07,'year')` → `-000-01` (-0.84 months rounds to
-1). Optional; the patch is correct without it.

Nothing blocking from my side — the fix is minimal, in the right place, and
consistent with the function's own rounding convention. Nice find, and thanks
for re-scoping it out of the earlier combined report.

btlqql:
Thanks for the independent confirmation — the source-exact UBSan harness is a stronger check than anything I can run here, and it settles the `(-1, -0.5]` → `+1` collapse exactly as the issue describes.

On your three points:

**1. The asymmetry around zero.** I read it the same way: `[0.01, 1) → 1` on the positive side is the pre-existing threshold, not something this patch introduces — the patch only stops the negative branch from collapsing onto the positive result. I have no Oracle instance available here (this machine has no server build, no container and no WSL either), so I cannot run the four `±0.4` / `±0.01` probes first-hand. If you can settle them, I will add the fixture pair in whichever direction Oracle pins. If you would rather keep this PR as the one-character sign fix, I will leave the positive threshold alone and open a separate issue for the asymmetry so it does not get lost.

**2. The `-0.07 year` row.** Happy to add it (`NUMTOYMINTERVAL(-0.07,'year')` → `-000-01`) as a second row in the same block. I left it out deliberately — the block is a single `interval year(3) to month` column and I wanted the smallest diff that pins the sign — but if you want the `year` path pinned in the same fixture, say so and I will push it.

**3. `NaN` / `1e30`.** Both look real and both are pre-existing (`result->month = interval_val;` with no magnitude check on the way in). Out of scope here, so I will file a follow-up issue carrying your UBSan result instead of bundling it into this PR.

**What is actually blocking this PR: no CI has ever run.** All six regression workflows on the current head (`b0d353874e`) are still `action_required` — queued at `2026-09-20T04:32:21Z` and never started:

- `build`, `contrib_regression`, `meson_build`, `oracle_pg_regression`, `oracle_regression`, `pg_regression`

Only `policy` has run (and passed). Could you approve them, or point me at whoever can? The expected output here was derived by reading `numtoyminterval()`, `EncodeYminterval()` and `AdjustYmintervalForTypmod()` rather than from a real run, so `oracle_regression` is the check that actually matters for this patch.

Partial evidence while that is pending: this exact head is green on my fork's copy of the repository CI — https://github.com/btlqql/IvorySQL/actions/runs/35489395857 — which is `pg-ci`, so it shows the tree builds and the core PostgreSQL tests are unaffected, but it does not run `contrib/ivorysql_ora`'s `oracle-check`. If it is more convenient than approving the upstream runs, I can also run `make -C contrib/ivorysql_ora oracle-check ORA_REGRESS='ora_datetime_datatype_functions'` on a branch in my fork and paste the output here.

PR #2041, opened 2026-09-13
fix(oracle): apply the CHAR length limit to CHAR(n CHAR) and CHAR(n BYTE)

### LINKED ISSUE #2040: [Bug] CHAR(n CHAR) and CHAR(n BYTE) bypass the 2000 length limit in Oracle mode
### Summary

In Oracle-compatible mode, `CHAR(n CHAR)` and `CHAR(n BYTE)` skip the 2000 limit that
IvorySQL already applies to the unqualified `CHAR(n)` spelling, so DDL that Oracle rejects
with `ORA-00910` is accepted.

### Steps to reproduce

In a database whose `ivorysql.database_mode` is `oracle`:

```sql
CREATE TABLE t1 (c char(2001));       -- correctly rejected
CREATE TABLE t2 (c char(2001 char));  -- accepted (bug)
CREATE TABLE t3 (c char(2001 byte));  -- accepted (bug)
```

Actual result:

```
CREATE TABLE t1 (c char(2001));
ERROR:  specified length too long for its datatype
CREATE TABLE t2 (c char(2001 char));   -- no error, table is created
CREATE TABLE t3 (c char(2001 byte));   -- no error, table is created
```

Expected result: all three statements fail with `ORA-00910 specified length too long for
its datatype`, whose documented cause is "for datatypes CHAR and RAW, the length specified
was > 2000". `char(32767 char)` and `char(32767 byte)` are accepted today as well.

### Root cause

`src/backend/oracle_parser/ora_gram.y` checks the limit only for the spelling without a
length qualifier (rule `CharacterWithLength`, line 17175):

```c
if ($3 > CHAR_TYPE_LENGTH_MAX)
    ereport(ERROR,
            (errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
             errmsg("specified length too long for its datatype"),
             parser_errposition(@3)));
```

whereas the alternative spellings are produced by the `OracleCharacter` rule (line 17260),
which builds the same `sys.oracharchar` / `sys.oracharbyte` types and never looks at the
length:

```c
OracleCharacter: character '(' Iconst CHAR_P ')'
                    {
                        if (strcmp($1, "bpchar"))
                            $1 = "oravarcharchar";
                        else
                            $1 = "oracharchar";
                        ...
                        $$->typmods = list_make1(makeIntConst($3, @3));
```

The typmodin function of those types only rejects lengths above 32767, so nothing else
catches the oversized value.

The same 2000 limit is applied to *both* variants elsewhere in the code base: the psql
`VARIABLE` command rejects `variable x char(2001 char)` and `variable x char(2001 byte)`
with "Bind variable length cannot exceed 2000."
(`src/bin/psql/variables.c:768`, `src/bin/psql/psqlplus.h:64`), and
`contrib/ivorysql_ora/expected/ora_character.out` already shows `char(32768 byte)` /
`char(32768 char)` being rejected by the server at 32767.

### Notes

Found by reading the source; there is no C toolchain/container available in this
environment, so the fix and the regression expectations have not been executed locally.

btlqql:
## What and why

In Oracle mode the 2000 limit for `CHAR` is enforced for the unqualified `CHAR(n)`
spelling only, so `CHAR(n CHAR)` and `CHAR(n BYTE)` silently accept lengths that Oracle
rejects with `ORA-00910`. This makes the three spellings of the same declared type
disagree, and it disagrees with IvorySQL's own client-side rule as well.

Reproduction (database created in Oracle mode):

```sql
CREATE TABLE t1 (c char(2001));       -- correctly rejected
CREATE TABLE t2 (c char(2001 char));  -- accepted (bug)
CREATE TABLE t3 (c char(2001 byte));  -- accepted (bug)
```

`char(32767 char)` / `char(32767 byte)` are accepted too. Oracle's documented cause for
`ORA-00910` is *"for datatypes CHAR and RAW, the length specified was > 2000"*, and
IvorySQL already emits that very message (as `specified length too long for its datatype`)
for `char(2001)` — see `src/oracle_test/regress/expected/strings.out`.

## The defect

`src/backend/oracle_parser/ora_gram.y` checks the limit in the rule that has no length
qualifier (`CharacterWithLength`, line 17175):

```c
if ($3 > CHAR_TYPE_LENGTH_MAX)
    ereport(ERROR,
            (errcode(ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE),
             errmsg("specified length too long for its datatype"),
             parser_errposition(@3)));
```

but the *qualified* spellings are produced by a different rule, `OracleCharacter`
(line 17260), which builds the very same types and never looks at the length:

```c
OracleCharacter: character '(' Iconst CHAR_P ')'
                    {
                        if (strcmp($1, "bpchar"))
                            $1 = "oravarcharchar";
                        else
                            $1 = "oracharchar";

                        $$ = OracleSystemTypeName($1);
                        $$->typmods = list_make1(makeIntConst($3, @3));
```

Nothing downstream catches it: `oracharchartypmodin` / `oracharbytetypmodin` only reject
lengths above `MaxCharSize` = 32767 (`contrib/ivorysql_ora/src/datatype/oracharchar.c:46`,
`oracharbyte.c:43`), and `CHAR_TYPE_LENGTH_MAX` is 2000
(`src/include/utils/ora_compatible.h:35`).
The same 2000 limit is already applied to both of these types by the psql `VARIABLE`
command (`src/bin/psql/variables.c:768`, `src/bin/psql/psqlplus.h:64`), which reports
"Bind variable length cannot exceed 2000." for `variable x char(2001 char)`.

## The fix

Run the same `CHAR_TYPE_LENGTH_MAX` check as `CharacterWithLength` in both alternatives of
`OracleCharacter` (24 added lines, no other change). `VARCHAR2(n CHAR|BYTE)` keeps its
32767 limit, and `RAW`/`LONG RAW` are untouched.

## The test

`contrib/ivorysql_ora/sql/ora_character.sql` gains the two `2001` cases next to the
existing limit cases, with the expected output in
`contrib/ivorysql_ora/expected/ora_character.out`:

```
CREATE TABLE TEST_ORACHAR(a char(2001 byte));
ERROR:  specified length too long for its datatype
LINE 1: CREATE TABLE TEST_ORACHAR(a char(2001 byte));
                                         ^
```

The two pre-existing `char(32768 char|byte)` expectations change from the 32767 typmodin
message to the CHAR error, because the parse-time check now fires first; the caret still
points at the length, as in the already-existing `char(2001)` case of
`src/oracle_test/regress`.

## What I could not verify

There is no C toolchain, no `bison`/`flex`, no `make` and no container on this Windows
host, so IvorySQL cannot be compiled and `oracle_regression` / `pg_regression` cannot be
run here. The diagnosis comes from reading the parser and the type support code, and the
expected output was derived from the implementation (same message, same
`parser_errposition` anchor as the sibling rule that is already covered by
`src/oracle_test/regress/expected/strings.out`). The caret column of the two updated
`char(32768 ...)` expectations is the one thing CI should confirm: the error now comes
from the grammar rule with `parser_errposition(@3)` instead of from the type's typmodin
function, which used to report the position of the type name. Reviewers should let
`oracle_regression` and `pg_regression` run before merging.

Fixes #2040

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced the maximum supported length for Oracle-compatible `CHAR(n CHAR)` and `CHAR(n BYTE)` declarations.
  * Oversized declarations now return a consistent “specified length too long for its datatype” error.
  * Added regression coverage for lengths exceeding the 2,000-character limit.


bigplaice:

PR #2043, opened 2026-09-13
fix(oracle): keep embedded NUL bytes in DBMS_OUTPUT PUT/PUT_LINE

### LINKED ISSUE #2042: [Bug] DBMS_OUTPUT PUT/PUT_LINE silently truncate values at embedded NUL bytes
## Bug report

`DBMS_OUTPUT.PUT_LINE` and `DBMS_OUTPUT.PUT` measure their text argument with
`text_to_cstring()` followed by `strlen()`, so any value that contains an
embedded NUL byte is stored truncated at the first NUL. Everything after the NUL
is silently lost and the stored line comes back shortened from
`DBMS_OUTPUT.GET_LINE` / `GET_LINES`.

A `text`/`VARCHAR2` value can contain an embedded NUL byte even though no SQL
literal can express one -- `convert_from()` over binary data produces one, so
this is reachable from ordinary SQL.

## Location

`contrib/ivorysql_ora/src/builtin_packages/dbms_output/dbms_output.c`

`ora_dbms_output_put_line()`:

```
		text	   *line_text = PG_GETARG_TEXT_PP(0);
		line_str = text_to_cstring(line_text);
		line_len = strlen(line_str);
```

`ora_dbms_output_put()`:

```
	str = text_to_cstring(PG_GETARG_TEXT_PP(0));
	str_len = strlen(str);
```

`add_line_to_buffer()` then copies only `line_len` bytes, and `PUT` appends with
`appendStringInfoString()`, which also stops at the first NUL.

## Current behavior

With a one-line check where the stored line is `'a' || chr(0) || 'b'`:

```sql
SELECT sys.ora_dbms_output_enable();
SELECT sys.ora_dbms_output_put_line(convert_from(decode('610062', 'hex'), 'UTF8'));
SELECT octet_length(line), encode(convert_to(line, 'UTF8'), 'hex')
  FROM sys.ora_dbms_output_get_line();
```

returns `1 | 61` instead of `3 | 610062`: the value is silently truncated.

The same happens for the pending-`PUT` accumulator, and for the merge path that
appends a `PUT_LINE` argument to pending `PUT` text.

## Expected behavior

`PUT`, `PUT_LINE` and `NEW_LINE` should store the value byte for byte: the
length should come from the varlena header (`VARSIZE_ANY_EXHDR`) and the bytes
from `VARDATA_ANY`, with `appendBinaryStringInfo()` used for the accumulator.

## Impact

Silent data loss for any value that legitimately contains a NUL byte (binary
data converted with `convert_from()`), and a wrong `ORU-10028` line-length check
for such values, since the length that is checked is the truncated one.

btlqql:
## What and why

`DBMS_OUTPUT.PUT_LINE` and `DBMS_OUTPUT.PUT` measured their `VARCHAR2`/`TEXT`
argument with `text_to_cstring()` followed by `strlen()`, so a value that
contains an embedded NUL byte was stored truncated at the first NUL, and the
rest of the value was silently dropped from the session buffer.

A `text` value can contain an embedded NUL byte even though no SQL literal can
express one -- `convert_from()` over binary data produces one -- so this is
reachable from ordinary SQL:

```sql
SELECT sys.ora_dbms_output_enable();
SELECT sys.ora_dbms_output_put_line(convert_from(decode('610062', 'hex'), 'UTF8'));
SELECT octet_length(line), encode(convert_to(line, 'UTF8'), 'hex')
  FROM sys.ora_dbms_output_get_line();
```

`'a\0b'` came back as `1 | 61` instead of `3 | 610062`.

## The defect

`contrib/ivorysql_ora/src/builtin_packages/dbms_output/dbms_output.c`,
`ora_dbms_output_put_line()`:

```
		text	   *line_text = PG_GETARG_TEXT_PP(0);
		line_str = text_to_cstring(line_text);
		line_len = strlen(line_str);
```

and `ora_dbms_output_put()`:

```
	str = text_to_cstring(PG_GETARG_TEXT_PP(0));
	str_len = strlen(str);
```

`add_line_to_buffer()` copies exactly `line_len` bytes into the buffer node, and
the pending-`PUT` accumulator used `appendStringInfoString()`, which also stops
at the first NUL -- so the merge path that prepends pending `PUT` text to a
`PUT_LINE` argument truncated as well. The `ORU-10028` line-length check
compared that same truncated length, so an over-long value could pass it.

## The fix

Take the length from the varlena header and the bytes from the varlena data
(`VARSIZE_ANY_EXHDR` / `VARDATA_ANY`) instead of using a NUL-terminated copy,
and append to the accumulator with `appendBinaryStringInfo()`. `varatt.h` is
included explicitly for those macros, exactly as `utl_encode.c` and
`dbms_random.c` already do in this extension.

## The test

`contrib/ivorysql_ora/sql/ora_dbms_output.sql` gains a new "Section 14", with
the matching output in `contrib/ivorysql_ora/expected/ora_dbms_output.out`.
Three cases: `PUT_LINE`, `PUT` + `NEW_LINE`, and `PUT` followed by `PUT_LINE`
(the merge path). Each builds `'a\0b'` with `convert_from()` and reports the
stored line as an octet length plus a hex dump, because psql cannot print a NUL
byte. Before the fix the first two cases report `len=1, hex=61` and the third
reports `len=3, hex=616263`; after it they report `3 | 610062`, `3 | 610062` and
`5 | 6162630064`.

## What I could not verify

This host has no C toolchain, no bison/flex, no container runtime and no WSL
distribution, so `ivorysql_ora` could not be built and the regression suite
could not be run. Correctness is argued from close reading of the C code plus
the varlena and `StringInfo` contracts; CI must run `oracle_regression` (and
`contrib_regression`) to confirm the new expected output. The new test calls the
internal `sys.ora_dbms_output_*` entry points directly rather than the
`dbms_output` package wrappers, so that the NUL-containing value is produced and
consumed entirely inside SQL expressions and never round-trips through a
PL/iSQL variable; the package wrappers themselves are unchanged and remain
covered by the pre-existing sections of the same test file.

Fixes #2042

## Summary by CodeRabbit

- **Bug Fixes**
  - DBMS_OUTPUT now preserves embedded NUL bytes in `PUT_LINE`, `PUT`, and `NEW_LINE` operations.
  - Output retains complete byte lengths and content, including data after embedded NUL characters.
  - Byte-based line limits continue to be enforced correctly.

- **Tests**
  - Added coverage verifying NUL-containing output through multiple DBMS_OUTPUT usage patterns.


btlqql:
Withdrawing this one. My regression test cannot pass, and the reachability claim in the description is wrong.

**The test's input construction raises an error before DBMS_OUTPUT is called.** `convert_from(<bytea>, 'UTF8')` runs `pg_convert`, which calls `pg_verify_mbstr(src_encoding, src_str, len, false)` before doing anything else. For UTF8 that is the shift-based DFA in `src/common/wchar.c`, and `0x00` is not a legal byte in it:

```c
#define	BGN 11
#define	ERR  0
#define ILL ERR
...
static const uint32 Utf8Transition[256] = {
	/* ASCII */
	ILL, ASC, ASC, ASC, ASC, ASC, ASC, ASC,   /* 0x00 is ILL, not ASC */
```

`ILL == ERR`, so the state machine lands in the error state and `pg_verify_mbstr` reports the byte. This repository already pins that behaviour down, for this same code path, in its own regression suite: `src/test/regress/sql/conversion.sql` runs `test_conv(inbytes, 'utf8', 'utf8')` over `('\x66006f', 'NUL byte')`, and `test_conv` -> `test_enc_conversion` calls `pg_encoding_verifymbstr`, i.e. the same DFA. `src/test/regress/expected/conversion.out:119`:

```
 NUL byte                           | \x66       | \x006f       | invalid byte sequence for encoding "UTF8": 0x00
```

So all three cases would fail with `ERROR: invalid byte sequence for encoding "UTF8": 0x00` and `oracle_regression` would go red. The "before the fix the first two cases report `len=1, hex=61`" table in the description is wrong as well — those calls never reach `sys.ora_dbms_output_put_line()`.

**"`convert_from()` can produce one" is not true, and neither is the wider reachability claim.** I went looking for any route that can put an embedded NUL into a `text`/`varchar2` value, and every one of them validates:

- SQL literals — rejected by the lexer.
- `oravarcharbytein()` — takes a C string, which cannot contain a NUL by construction.
- `oravarcharbyterecv()` -> `pq_getmsgtext()` -> `pg_client_to_server()` -> `pg_any_to_server()`, which validates even when no conversion is needed ("we never want to just assume validity"): `pg_verify_mbstr(DatabaseEncoding->encoding, s, len, false)`. So the binary wire format and `COPY ... WITH (FORMAT binary)` are covered too.
- `CREATE CAST (pg_catalog.bytea AS sys.oravarcharbyte)` / `... AS sys.oravarcharchar` (`datatype--1.0.sql:10562`, `:10566`) are `WITH INOUT`, so they go through `byteaout()` and yield the twelve literal characters `\x610062`, not three bytes containing a NUL. There is no `bytea -> text` cast in the tree.
- `UTL_FILE.GET_LINE()` — `utl_file.c:1350` calls `pg_verify_mbstr(encoding, buffer, csize, false)` before building the result.
- `UTL_ENCODE.BASE64_DECODE()` — returns `bytea` (RAW), not text.

So I cannot construct a NUL-bearing `text`, which means I cannot write a regression test for this change, and I cannot show that the truncation is reachable in the first place. The `strlen()` handling is genuinely inconsistent with the surrounding code — `add_line_to_buffer()` is length-delimited and `ora_dbms_output_get_line()` returns the line with `cstring_to_text_with_len(node->data, node->len)` — but "defensive consistency" is a weaker claim than the bug report this PR is based on, and not one I can demonstrate with a test.

Closing rather than leaving a red `oracle_regression` for a maintainer to triage. The diagnosis was wrong, not just the test.

PR #2045, opened 2026-09-13
fix(oracle): honour the n and m match parameters in REGEXP_LIKE

### LINKED ISSUE #2044: REGEXP_LIKE rejects the documented 'm' match parameter and misreads 'n'
`sys.regexp_like()` (the C function behind the Oracle-compatible `REGEXP_LIKE`, `contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c`) parses its `match_parameter` argument with its own `switch`, instead of `ora_parse_re_flags()` which the other five `REGEXP_*` functions in the same file already use. As a result the Oracle option letters do not mean what Oracle documents:

* `'m'` is not handled at all:
  `select sys.regexp_like('X' || chr(10), 'X.', 'm') from dual;`
  -> `ERROR:  invalid option of regexp_like: m` (Oracle returns `FALSE`)
* `'n'` sets `REG_NEWLINE`, i.e. PostgreSQL's newline-sensitive mode, which *excludes* a newline from `"."` and enables the multiline anchors.  Oracle's `'n'` is the opposite: it *allows* `"."` to match a newline and must not turn on the anchors:
  `select sys.regexp_like('x' || chr(10) || 'b' || chr(10) || 'y', '^b$', 'n') from dual;`
  -> `t`; Oracle returns `FALSE` (only `'m'` enables the multiline anchors)
* without any `match_parameter`, `"."` matches a newline, which Oracle only allows with `'n'`:
  `select sys.regexp_like('X' || chr(10), 'X.') from dual;`
  -> `t`; Oracle returns `FALSE`

`ora_regexp_like_no_flags()` (the two-argument registration) has the same third problem, because it also sets `out_flag = REG_ADVANCED` without `REG_NLDOT`.

This is measurable against the neighbouring `REGEXP_COUNT`, `REGEXP_INSTR`, `REGEXP_SUBSTR` and `REGEXP_REPLACE` cases in `contrib/ivorysql_ora/sql/ora_character_datatype_functions.sql`, which already pin the n/m matrix as `0|0|1|0` (dot vs. newline) and `0|0|0|1` (multiline anchors). `REGEXP_LIKE` is the only member of the family that does not follow those rules.

Fix: parse the flags exactly like the rest of the family (a patch is on its way).

btlqql:
## Defect

`REGEXP_LIKE` is implemented by `sys.regexp_like(...)` -> `ora_regexp_like()` in
`contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c`. It parses the
`match_parameter` argument with its own `switch`, while the other five `REGEXP_*` functions in the
same file go through `ora_parse_re_flags()` (`src/backend/utils/adt/regexp.c`). The two have drifted
apart, so the Oracle option letters do not mean what Oracle documents:

```c
	/* parse flag options */
	out_flag = REG_ADVANCED;
	for (i = 0; i <in_flag_len; i++)
	{
		switch (in_flag_p[i])
		{
			case 'i':
				out_flag  |=  REG_ICASE;
				break;
			case 'n':
				out_flag |= REG_NEWLINE;      /* <-- PG newline-sensitive mode */
				break;
			case 'c':
				out_flag  &=  ~REG_ICASE;
				break;
			case 'x':
				out_flag |= REG_EXPANDED;
				break;
			default:
				ereport(ERROR,
						(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
						 errmsg("invalid option of regexp_like: %c",
						 in_flag_p[i])));
				break;
		}
	}
```

(`ora_regexp_like()` lines 791-816 before the patch; `ora_regexp_like_no_flags()` had the same
`out_flag = REG_ADVANCED;` two lines below.)

Three consequences:

1. `'m'` falls into `default:` -> the documented Oracle option raises an error.
2. `'n'` sets PostgreSQL's `REG_NEWLINE` (`REG_NLSTOP | REG_NLANCH`), which *excludes* a newline
   from `"."` and turns on the multiline anchors. Oracle's `'n'` is the opposite: it *allows* `"."`
   to match a newline, and only `'m'` may turn on the anchors.
3. With no `match_parameter`, `"."` matches a newline; Oracle only permits that with `'n'`.
   `ora_regexp_like_no_flags()` (the two-argument registration) is wrong for the same reason.

## How I triggered it

`sys.regexp_like` is the C function behind the package's `REGEXP_LIKE`; in Oracle-compatible mode the
unqualified name reaches it too (that is what the neighbouring `REGEXP_COUNT`/`REGEXP_SUBSTR` cases
in the same regression file rely on), and the qualified form is unambiguous either way:

```sql
-- raises an error today, Oracle returns FALSE
select sys.regexp_like('X' || chr(10), 'X.', 'm') from dual;
ERROR:  invalid option of regexp_like: m

-- '.' must not match a newline without 'n'; Oracle returns FALSE
select sys.regexp_like('X' || chr(10), 'X.') from dual;          -- t today

-- 'n' must not enable the multiline anchors; Oracle returns FALSE
select sys.regexp_like('x' || chr(10) || 'b' || chr(10) || 'y',
                       '^b$', 'n') from dual;                    -- t today
```

The expected values are not invented: the same file already pins the identical n/m matrix for
`REGEXP_COUNT`, `REGEXP_INSTR`, `REGEXP_SUBSTR` and `REGEXP_REPLACE` at `0|0|1|0` (dot vs. newline)
and `0|0|0|1` (multiline anchors), and the test below reproduces exactly those two rows for
`REGEXP_LIKE`.

## Fix

Parse the flags the way the rest of the family does (`character_datatype_functions.c`, 2 hunks, 19
lines): both entry points start from `REG_ADVANCED | REG_NLDOT` (Oracle's default: `"."` does not
match a newline), `'n'` clears `REG_NLDOT` and `'m'` sets `REG_NLANCH`. The unknown-option error
message is unchanged, so the existing `ERROR: invalid option of regexp_like: p` expectation stays
valid.

## Regression test

`contrib/ivorysql_ora/sql/ora_character_datatype_functions.sql` + the matching `expected` file: two
statements appended to the existing match-parameter matrix, using the same `chr(10)` idiom and the
same column aliases as the neighbouring cases (so the expected table widths are identical to rows
that are already in the file).

## What I could not verify

There is no C toolchain, `make`, `bison`/`flex`, `perl` or container in my environment, so I could
not build IvorySQL or run `oracle_regression`/`contrib_regression` locally. **The patch and the
expected output are unverified by execution** - CI must run `contrib/ivorysql_ora`'s
`ora_regression` (the `ora_character_datatype_functions` test) to confirm. Everything else was
checked statically against this tree: `REG_NLDOT`/`REG_NLANCH` come from `src/include/regex/regex.h`
(already included by the file, and `REG_NLANCH` is already used in `ora_regexp_count`),
`ora_parse_re_flags()` in `src/backend/utils/adt/regexp.c` is the intended semantics
(`'n'` -> clear `REG_NLDOT`, `'m'` -> set `REG_NLANCH`), and the expected numbers were taken from the
verified `REGEXP_COUNT` block for the same expressions.

Fixes #2044

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated regular expression matching so `sys.regexp_like` consistently handles newline and multiline options.
  - The `n` option allows dots to match newline characters, while `m` enables multiline `^` and `$` anchors.
  - Default matching behavior now excludes newline characters from dot matches.


bigplaice:

PR #2048, opened 2026-09-14
Fix ROWID sequence identity

### LINKED ISSUE #2037: ROWID sequence lookup breaks after metadata changes and can delete user sequences
## Summary

ROWID tables discover their backing sequence by reconstructing a name from the table's current name, schema, and owner. These values are mutable and are not the sequence identity.

## Reproduction

In a fresh Oracle-mode cluster:

`sql
CREATE SEQUENCE collision_path_rowid_seq START WITH 1000;
CREATE TABLE collision_path (id integer) WITH ROWID;
INSERT INTO collision_path VALUES (1);
ALTER TABLE collision_path SET WITHOUT ROWID;
`

The table creates collision_path_rowid_seq1, but the insert consumes the pre-existing user sequence (
owid.rowno = 1000). Removing ROWID deletes the user sequence and leaves the generated sequence orphaned.

`sql
CREATE TABLE rowid_before_rename (id integer) WITH ROWID;
INSERT INTO rowid_before_rename VALUES (1);
ALTER TABLE rowid_before_rename RENAME TO rowid_after_rename;
INSERT INTO rowid_after_rename VALUES (2), (3);
`

The last two rows have 
owid.rowno = 0. SET WITHOUT ROWID also leaves the original backing sequence behind. The same name/schema/owner lookup is affected by SET SCHEMA and OWNER TO.

## Root cause

RelationGetRowIdSeqId() and the SET WITHOUT ROWID cleanup path search pg_class using reconstructed name, namespace, and owner, rather than resolving the sequence's existing OWNED BY table.rowid dependency.

## Expected behavior

Resolve and remove only the sequence owned by the table's ROWID system attribute, using stable OID-based dependency metadata. Metadata DDL must preserve nonzero, unique ROWID generation.

## Validation

Reproduced on current upstream/master (

R0CKing666:
Fixes #2037

## Summary
- bind ROWID sequences to the ROWID system attribute through OWNED BY metadata
- resolve and remove the backing sequence by stable OID dependency instead of mutable table metadata
- invalidate the relation cache after an OWNED BY change
- add ROWID sequence collision and rename lifecycle regression coverage

## Testing
- Docker Ubuntu 22.04: rebuilt the modified backend modules
- fresh Oracle-mode cluster: collision, ROWID removal, rename, and removal checks
- git diff --check

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ROWID sequence tracking so associations remain correct when tables are renamed or similarly named sequences exist.
  - ROWID removal now targets only the sequence directly associated with the table, avoiding unintended changes to other sequences.
  - Updated sequence ownership handling so cached metadata stays accurate when ownership changes, ROWID is disabled, or inserts occur afterward.

- **Tests**
  - Added regression coverage for sequence name collisions, table renames, disabling ROWID, inserts, and generated-sequence cleanup.


bigplaice:

PR #2050, opened 2026-09-14
Fix strlen-based truncation of varlena with embedded NUL bytes in builtin packages

### LINKED ISSUE #2049: Builtin packages silently truncate data when text input contains embedded NUL bytes
## Enhancement
In PR1369: dbms_session.c
in PR998: dbms_output.c
in PR1194: dbms_lock.c
several builtin package functions use 'text_to_cstring()' followed by 'strlen()' to compute the byte length of a ‘text’ argument. when the input contains an embedded NUL byte, strlen() stops at the first NUL, causing silent data truncation - no error is raised, but the stored/hashed/output data is incomplite.

enkilee:
fix #2049

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of text values containing embedded NUL bytes in DBMS_OUTPUT, preserving complete content in `PUT` and `PUT_LINE`.
  - Updated session context operations to retain exact text values and apply length-aware namespace matching, preventing similarly prefixed namespaces from being cleared unintentionally.
  - Improved lock name processing so distinct names produce distinct lock handles, including names that differ after an embedded NUL byte.


bigplaice:

muzimu217:
Draft comment body (English, first person):

Thanks for this fix — the problem in #2049 is real, and I like that the PUT line-length limit is now enforced on the true byte length, and that `GET_LINE` already reads back via `cstring_to_text_with_len()`, so the DBMS_OUTPUT write/read round trip becomes byte-exact with this change.

While reading the patch against master, I found three follow-up spots that still use NUL-terminated semantics, which I believe sit inside the scope of the same issue. All are concrete and reproducible; I have not run the patched build, so please double-check my reading.

**1. `clear_namespace()` in dbms_session.c still matches with `strcmp()` — possible cross-namespace deletion**

The patch changes `make_key()` to store full-length, upper-cased keys (embedded NUL bytes included), but `clear_namespace()` was left unchanged: it still takes a `const char *ns` (so its argument is already truncated by `text_to_cstring()`), upper-cases only up to the first NUL, and matches victims with `strcmp(entry->key.namespace, ns_up) == 0`. Since `strcmp()` stops at the first NUL, a stored key of `'AB' || chr(0) || 'C'` compares equal to `'AB'`.

So with this patch applied, `SET_CONTEXT('AB' || chr(0) || 'C', ...)` and `SET_CONTEXT('AB', ...)` create two distinct hash keys, yet `CLEAR_CONTEXT('AB')` deletes both. I verified the comparison semantics with a standalone harness that copies the two functions' logic verbatim:

```
stored namespace bytes : AB 00 43 (i.e. 'AB' || chr(0) || 'C')
CLEAR_CONTEXT argument : "AB" (after text_to_cstring)
strcmp(stored, arg)    : 0
=> clear_namespace("AB") deletes the 'AB'||chr(0)||'C' entry : YES
```

Making `clear_namespace()` length-aware (e.g. `strncmp()` against `DBMS_SESSION_NAME_LEN`, plus a length-based guard instead of `strlen()`) looks like it would close this. I can help test that if useful.

**2. The read paths for values and keys still truncate, so the fix is not observable end-to-end**

`set_context()` now stores the value byte-exactly (`palloc` + `memcpy` + explicit terminator — nice), but:

- `get_context()` returns `cstring_to_text(entry->value)`, so `SET_CONTEXT('NS','ATTR','V1' || chr(0) || 'V2')` followed by `GET_CONTEXT('NS','ATTR')` still yields `'V1'`;
- `list_context()` emits namespace, attribute, and value through `CStringGetTextDatum()`, which is also strlen-based.

After this, the stored bytes are exact but every reader truncates at the first NUL. In Oracle, `LENGTH('a' || chr(0) || 'b')` is 3, so embedded chr(0) is ordinary data there; if we want IvorySQL to match, the readers probably need the same `with_len` treatment that `DBMS_OUTPUT.GET_LINE` already has. (Happy to cross-check the exact Oracle behavior for GET_LINE/GET_CONTEXT with chr(0) on a real instance — we have an Oracle comparison environment.)

**3. `get_context()`'s over-long guard now contradicts its own comment**

`get_context()` keeps the strlen-based guard

```c
if (strlen(ns) >= DBMS_SESSION_NAME_LEN || strlen(attr) >= DBMS_SESSION_NAME_LEN)
    PG_RETURN_NULL();
```

and then calls the new length-based `make_key()`, which raises `ERRCODE_INVALID_PARAMETER_VALUE`. For an argument like `'A' || chr(0) || repeat('B', 255)` (strlen 1, true length 257 ≥ 256), the guard passes and the call now errors out. The comment right above it says the read path deliberately returns NULL "without aborting the query on an oversized argument" — before this patch that was true, now the same input aborts the query. Guarding on `VARSIZE_ANY_EXHDR()` instead of `strlen()` would keep the documented contract.

**4. No test coverage in this PR**

All three touched packages have regression suites, and every behavior above is cheap to pin down. Cases I'd suggest (in whichever of the three `ora_*.sql` files fits):

- `PUT_LINE('a' || chr(0) || 'b')` then `GET_LINE` round trip;
- `SET_CONTEXT`/`GET_CONTEXT` with a chr(0) inside namespace, attribute, and value (the value case currently truncates);
- `CLEAR_CONTEXT('AB')` must not remove the `'AB' || chr(0) || 'C'` namespace (currently does, per item 1);
- two `DBMS_LOCK.ALLOCATE_UNIQUE` names differing only after chr(0) producing distinct locks.

One small nit: in `put`/`put_line`, `text_to_cstring()` is still called just to feed `appendBinaryStringInfo()`; using `VARDATA_ANY()` + `VARSIZE_ANY_EXHDR()` on the original datum would skip the copy. Not important.

Side note: one of the "policy" checks on the current head shows FAILURE — probably worth a glance in case it is DCO/CLA related.

Happy to re-review after an update, and I can verify the chr(0) cases against a real Oracle 19c instance if that would help settle the intended semantics.

enkilee:
> Draft comment body (English, first person):
> 
> Thanks for this fix — the problem in #2049 is real, and I like that the PUT line-length limit is now enforced on the true byte length, and that `GET_LINE` already reads back via `cstring_to_text_with_len()`, so the DBMS_OUTPUT write/read round trip becomes byte-exact with this change.
> 
> While reading the patch against master, I found three follow-up spots that still use NUL-terminated semantics, which I believe sit inside the scope of the same issue. All are concrete and reproducible; I have not run the patched build, so please double-check my reading.
> 
> **1. `clear_namespace()` in dbms_session.c still matches with `strcmp()` — possible cross-namespace deletion**
> 
> The patch changes `make_key()` to store full-length, upper-cased keys (embedded NUL bytes included), but `clear_namespace()` was left unchanged: it still takes a `const char *ns` (so its argument is already truncated by `text_to_cstring()`), upper-cases only up to the first NUL, and matches victims with `strcmp(entry->key.namespace, ns_up) == 0`. Since `strcmp()` stops at the first NUL, a stored key of `'AB' || chr(0) || 'C'` compares equal to `'AB'`.
> 
> So with this patch applied, `SET_CONTEXT('AB' || chr(0) || 'C', ...)` and `SET_CONTEXT('AB', ...)` create two distinct hash keys, yet `CLEAR_CONTEXT('AB')` deletes both. I verified the comparison semantics with a standalone harness that copies the two functions' logic verbatim:
> 
> ```
> stored namespace bytes : AB 00 43 (i.e. 'AB' || chr(0) || 'C')
> CLEAR_CONTEXT argument : "AB" (after text_to_cstring)
> strcmp(stored, arg)    : 0
> => clear_namespace("AB") deletes the 'AB'||chr(0)||'C' entry : YES
> ```
> 
> Making `clear_namespace()` length-aware (e.g. `strncmp()` against `DBMS_SESSION_NAME_LEN`, plus a length-based guard instead of `strlen()`) looks like it would close this. I can help test that if useful.
> 
> **2. The read paths for values and keys still truncate, so the fix is not observable end-to-end**
> 
> `set_context()` now stores the value byte-exactly (`palloc` + `memcpy` + explicit terminator — nice), but:
> 
> * `get_context()` returns `cstring_to_text(entry->value)`, so `SET_CONTEXT('NS','ATTR','V1' || chr(0) || 'V2')` followed by `GET_CONTEXT('NS','ATTR')` still yields `'V1'`;
> * `list_context()` emits namespace, attribute, and value through `CStringGetTextDatum()`, which is also strlen-based.
> 
> After this, the stored bytes are exact but every reader truncates at the first NUL. In Oracle, `LENGTH('a' || chr(0) || 'b')` is 3, so embedded chr(0) is ordinary data there; if we want IvorySQL to match, the readers probably need the same `with_len` treatment that `DBMS_OUTPUT.GET_LINE` already has. (Happy to cross-check the exact Oracle behavior for GET_LINE/GET_CONTEXT with chr(0) on a real instance — we have an Oracle comparison environment.)
> 
> **3. `get_context()`'s over-long guard now contradicts its own comment**
> 
> `get_context()` keeps the strlen-based guard
> 
> ```c
> if (strlen(ns) >= DBMS_SESSION_NAME_LEN || strlen(attr) >= DBMS_SESSION_NAME_LEN)
>     PG_RETURN_NULL();
> ```
> 
> and then calls the new length-based `make_key()`, which raises `ERRCODE_INVALID_PARAMETER_VALUE`. For an argument like `'A' || chr(0) || repeat('B', 255)` (strlen 1, true length 257 ≥ 256), the guard passes and the call now errors out. The comment right above it says the read path deliberately returns NULL "without aborting the query on an oversized argument" — before this patch that was true, now the same input aborts the query. Guarding on `VARSIZE_ANY_EXHDR()` instead of `strlen()` would keep the documented contract.
> 
> **4. No test coverage in this PR**
> 
> All three touched packages have regression suites, and every behavior above is cheap to pin down. Cases I'd suggest (in whichever of the three `ora_*.sql` files fits):
> 
> * `PUT_LINE('a' || chr(0) || 'b')` then `GET_LINE` round trip;
> * `SET_CONTEXT`/`GET_CONTEXT` with a chr(0) inside namespace, attribute, and value (the value case currently truncates);
> * `CLEAR_CONTEXT('AB')` must not remove the `'AB' || chr(0) || 'C'` namespace (currently does, per item 1);
> * two `DBMS_LOCK.ALLOCATE_UNIQUE` names differing only after chr(0) producing distinct locks.
> 
> One small nit: in `put`/`put_line`, `text_to_cstring()` is still called just to feed `appendBinaryStringInfo()`; using `VARDATA_ANY()` + `VARSIZE_ANY_EXHDR()` on the original datum would skip the copy. Not important.
> 
> Side note: one of the "policy" checks on the current head shows FAILURE — probably worth a glance in case it is DCO/CLA related.
> 
> Happy to re-review after an update, and I can verify the chr(0) cases against a real Oracle 19c instance if that would help settle the intended semantics.

thx

PR #2052, opened 2026-09-14
fix insertchildxml/insertchildxml2 crash on NULL children pointer from xmlParseDoc

### LINKED ISSUE #2051: insertchildxml/insertchildxml2 crash on NULL children pointer from xmlParseDoc
### What happened?

`insertchildxml` and `insertchildxml2` crash with a segmentation fault when
`xmlParseDoc()` returns a non-NULL document that has no child nodes
(`doc->children == NULL`). This can happen when the input XML string contains
only a DTD declaration, a comment, or is empty after parsing.

The code checks `new_node == NULL` but then dereferences `new_node->children->name`
without checking whether `new_node->children` is NULL.

```c
new_node = xmlParseDoc(nodestring);
if (new_node == NULL)
{
    ereport(ERROR, ...);   /* only checks NULL doc */
}
else
{
    /* new_node->children may still be NULL! */
    if (strcmp(cname, (char *)((xmlNodePtr)new_node->children->name)) != 0)
    /*                                    ^^^^^^^^ NULL dereference */
}
```

`xmlParseDoc()` can return a valid `xmlDocPtr` with `children == NULL` for
inputs like:
- `<?xml version="1.0"?>` (declaration only, no root element)
- `<!-- comment -->`
- Strings that libxml2 parses to a document node with no element children

```sql
-- Crash: XML declaration only, no root element
SELECT insertchildxml(
    '<root><a/></root>'::xmltype,
    '/root'::text,
    'a'::text,
    '<?xml version="1.0"?>'::xmltype
);
```

### IvorySQL version

master

### Compatible mode

oracle

### Server encoding

UTF8

_No response_

### OS version

<details>

```console
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Before submitting

- [ ] I searched existing issues and found no duplicate

enkilee:
fix #2051 

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved XML validation when inserting child XML content.
  - Invalid XML documents without child nodes now return a clear error instead of causing processing failures.


bigplaice:

PR #2059, opened 2026-09-15
Declare the time-zone-bearing sys interval arithmetic STABLE, not IMMUTABLE

### LINKED ISSUE #2055: Interval arithmetic on the time-zone-bearing sys types is IMMUTABLE although it depends on the session TimeZone
Twelve `sys.*` interval-arithmetic functions that operate on a time-zone-bearing type are declared `IMMUTABLE`, although their result depends on the session `TimeZone`. Upstream PostgreSQL marks the equivalents `STABLE`, and the difference is user-visible: because the extension claims immutability, PostgreSQL accepts an expression index on such an expression and then returns wrong answers once the time zone changes.

## Affected declarations

`contrib/ivorysql_ora/src/datatype/datatype--1.0.sql`:

| line | function |
|---|---|
| 7590 | `sys.oratimestamptz_mi_yminterval(sys.oratimestamptz, sys.yminterval)` |
| 7605 | `sys.oratimestamptz_mi_dsinterval(sys.oratimestamptz, sys.dsinterval)` |
| 7638 | `sys.oratimestamptz_pl_yminterval(sys.oratimestamptz, sys.yminterval)` |
| 7654 | `sys.oratimestamptz_pl_dsinterval(sys.oratimestamptz, sys.dsinterval)` |
| 7749 | `sys.oratimestampltz_mi_yminterval(sys.oratimestampltz, sys.yminterval)` |
| 7764 | `sys.oratimestampltz_mi_dsinterval(sys.oratimestampltz, sys.dsinterval)` |
| 7797 | `sys.oratimestampltz_pl_yminterval(sys.oratimestampltz, sys.yminterval)` |
| 7813 | `sys.oratimestampltz_pl_dsinterval(sys.oratimestampltz, sys.dsinterval)` |
| 7912 | `sys.yminterval_pl_oratimestamptz(sys.yminterval, sys.oratimestamptz)` |
| 7928 | `sys.dsinterval_pl_oratimestamptz(sys.dsinterval, sys.oratimestamptz)` |
| 7944 | `sys.yminterval_pl_oratimestampltz(sys.yminterval, sys.oratimestampltz)` |
| 7960 | `sys.dsinterval_pl_oratimestampltz(sys.dsinterval, sys.oratimestampltz)` |

## Why they are not immutable

`oratimestamptz_pl_interval()` (`src/datatype/oratimestamp.c`, called by all `pl`/`mi` variants) needs the session zone to keep wall-clock time across a DST change:

```c
tz = DetermineTimeZoneOffset(tm, session_timezone);     /* line 1270 */
...
tz = DetermineTimeZoneOffset(tm, session_timezone);     /* line 1294 */
```

`oratimestamptz_mi_interval()` (line 1317) and `interval_pl_oratimestamptz()` (line 1372) negate / reorder the argument and call the same function via `DirectFunctionCall2`, so they inherit the dependency.

Upstream marks the equivalents `STABLE` for this reason — `timestamptz_pl_interval`, `timestamptz_mi_interval` and `interval_pl_timestamptz` all have `provolatile = 's'` in `pg_proc.dat`.

## Evidence

Ubuntu 22.04, gcc 12.3, `--enable-cassert --enable-debug`, based on `master` (`03b24b1c46`).

**1. Catalog state** — all twelve report `i`:

```
oratimestamptz_pl_dsinterval   'timestamp with time zone, interval day to second'   i
oratimestamptz_mi_yminterval   'timestamp with time zone, interval year to month'   i
dsinterval_pl_oratimestamptz   'interval day to second, timestamp with time zone'   i
...
```

**2. The expression is genuinely time-zone dependent.** Same input, two session zones:

```
TimeZone=UTC            2024-03-31 12:00:00+00
TimeZone=Europe/Berlin  2024-03-31 13:00:00+02     <- 11:00 UTC, DST switch on 2024-03-31
```

`timestamptz + interval '1 day'` on the native type behaves identically, and that is exactly why upstream refuses to call it immutable.

**3. PostgreSQL trusts the claim.** Same statement, two types:

```
-- native timestamptz
CREATE INDEX ON n_native ((ts + interval '1 day'));

-- sys.oratimestamptz
CREATE INDEX o_ora_expr ON o_ora ((ts + sys.numtodsinterval(1,'day')));
CREATE INDEX
```

**4. The accepted index then returns wrong results.** Row inserted and index built under `UTC`; queried under `Europe/Berlin`, where the expression evaluates to `2024-03-31 11:00+00`:

```
seq scan   finds: 1
index scan finds: 0
```

with the index path chosen:

```
Index Scan using o_ora_expr on o_ora
  Index Cond: ((ts + '...'::interval day to second) = '2024-03-31 13:00:00+00'::timestamp with time zone)
```

The index key was computed with the zone in effect at build time, so a query that is correct on a sequential scan silently misses the row.

## Reproduction

```sql
set ivorysql.compatible_mode = 'oracle';

create table o_ora(ts sys.oratimestamptz);
set TimeZone = 'UTC';
insert into o_ora values ('2024-03-30 12:00:00+00'::sys.oratimestamptz);
create index o_ora_expr on o_ora ((ts + sys.numtodsinterval(1,'day')));
analyze o_ora;

set TimeZone = 'Europe/Berlin';
set enable_seqscan = on;  set enable_indexscan = off;
select count(*) from o_ora where ts + sys.numtodsinterval(1,'day') = '2024-03-31 11:00:00+00'::sys.oratimestamptz;  -- 1
set enable_seqscan = off; set enable_indexscan = on;
select count(*) from o_ora where ts + sys.numtodsinterval(1,'day') = '2024-03-31 11:00:00+00'::sys.oratimestamptz;  -- 0
```

## Expected behavior

Declare the twelve functions `STABLE`. Once that is done PostgreSQL will reject the expression index in step 3 with the same message it already produces for the native type, instead of accepting it and returning stale results.

## Scope notes

- The `oratimestamp_*`, `oradate_*` and non-tz variants of the same operators are also `IMMUTABLE`, and that is correct — a plain `timestamp`/`date` has no zone, matching upstream. They should not be changed.
- No regression test currently builds an expression index on these operators or asserts `provolatile` for them, so the change should not require expected-output updates.
- This is a different set from #1754 (`sys.uid`, `sys.tz_offset`), #1696, #1800, #1830, #1843, #2008 and #2046.

jiangdaoli11:
Fixes #2055

## Summary

Twelve `sys.*` interval-arithmetic functions that take or return a time-zone-bearing
type (`sys.oratimestamptz`, `sys.oratimestampltz`) were declared `IMMUTABLE`, although
their result depends on the session `TimeZone`:

| group | functions |
|---|---|
| `oratimestamptz` | `oratimestamptz_mi_yminterval`, `oratimestamptz_mi_dsinterval`, `oratimestamptz_pl_yminterval`, `oratimestamptz_pl_dsinterval` |
| `oratimestampltz` | `oratimestampltz_mi_yminterval`, `oratimestampltz_mi_dsinterval`, `oratimestampltz_pl_yminterval`, `oratimestampltz_pl_dsinterval` |
| interval-left | `yminterval_pl_oratimestamptz`, `dsinterval_pl_oratimestamptz`, `yminterval_pl_oratimestampltz`, `dsinterval_pl_oratimestampltz` |

They are now declared `STABLE`.

## Why the old marking was wrong

`oratimestamptz_pl_interval()` (`src/datatype/oratimestamp.c`) calls
`DetermineTimeZoneOffset(tm, session_timezone)` to preserve wall-clock time across a
DST change; `oratimestamptz_mi_interval()` and `interval_pl_oratimestamptz()` merely
negate or reorder the arguments and call it again through `DirectFunctionCall2`, so
they inherit the dependency. Upstream marks the equivalents `STABLE` for exactly this
reason (`timestamptz_pl_interval`, `timestamptz_mi_interval`,
`interval_pl_timestamptz` all have `provolatile = 's'` in `pg_proc.dat`).

## Effect of the fix

Because the extension claimed immutability, PostgreSQL accepted an expression index on
such an expression and then returned stale results once the session zone changed — the
index key was computed with the zone in effect at build time, so a query that is
correct on a sequential scan silently misses the row. With `STABLE` the index is
rejected, exactly as it already is for the native type:

```sql
set ivorysql.compatible_mode = 'oracle';
create table o_ora(ts sys.oratimestamptz);
-- native type
CREATE INDEX ON n_native ((ts + interval '1 day'));
-- ERROR:  functions in index expression must be marked IMMUTABLE
-- sys.oratimestamptz (after this patch)
create index o_ora_expr on o_ora ((ts + sys.numtodsinterval(1,'day')));
-- ERROR:  functions in index expression must be marked IMMUTABLE
```

## Scope

- Only the twelve time-zone-bearing declarations change; no C code is touched.
- The variants that carry no zone (`oratimestamp_*`, `oradate_*`) stay `IMMUTABLE`,
  matching upstream.
- `PARALLEL SAFE` is retained (`STABLE` + `PARALLEL SAFE` is valid, and upstream's
  `timestamptz_pl_interval` is marked the same way).
- No regression test references these functions or asserts `provolatile` for them, so
  no expected-output updates are required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected timezone-aware date and interval arithmetic to properly account for the session time zone.
  * Updated addition and subtraction operations involving year-month and day-time intervals to produce consistent results when session time zone settings affect calculations.

* **Compatibility**
  * Existing function signatures, return types, and arithmetic operations remain unchanged.


bigplaice:

PR #2061, opened 2026-09-15
fix(oracle): stop ASCIISTR from escaping the ASCII backslash

### LINKED ISSUE #2060: ASCIISTR escapes the ASCII backslash character, unlike Oracle
Oracle `ASCIISTR` converts only **non-ASCII** characters to `\xxxx`. The backslash (U+005C) is ASCII and must pass through unchanged.

```sql
-- Oracle:
SELECT ASCIISTR('Hello\nWorld') FROM DUAL;
-- Hello\nWorld

-- IvorySQL (wrong on master):
SELECT sys.asciistr('Hello\nWorld') FROM DUAL;
-- Hello\005CnWorld
```

Root cause: `ora_asciistr` in `contrib/ivorysql_ora/src/builtin_functions/character_datatype_functions.c` special-cases `c == '\\'` and emits `appendUTF16Escape(..., 0x005C)` instead of treating the backslash as ordinary ASCII.

The regression expected file `ora_character_datatype_functions.out` currently encodes the wrong behavior (`Hello\005CnWorld`).

89799969:
## What

Stop `ASCIISTR` / `sys.asciistr` from escaping the ASCII backslash character.

## Why

Oracle's ASCIISTR converts only non-ASCII characters to `\xxxx`. The backslash is ASCII and must pass through unchanged. IvorySQL previously emitted `\005C` for every backslash, so `Hello\nWorld` became `Hello\005CnWorld`.

Fixes #2060

## How

- Remove the `c == '\\\\'` special case so the backslash follows the ASCII passthrough path.
- Update the regression expected file that locked in the old wrong output.

## Verification

- Source change is local to `ora_asciistr`.
- Full `oracle-check` / `installcheck` requires a Linux PostgreSQL build environment (not available on this Windows machine).

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected `ASCI​​STR` handling for backslash characters.
  - Backslashes are now preserved as ordinary ASCII characters instead of being converted into UTF-16 escape sequences.
  - Updated output expectations ensure strings containing backslashes, such as `Hello\nWorld`, return the original backslash representation.


bigplaice:
The issue that this PR addressed seems wrong.

89799969:
Closing: the underlying assumption in #2060 is incorrect.

@bigplaice verified on Oracle Database 23ai Free (23.4.0.24.05):

```
SELECT ASCIISTR('Hello\nWorld') FROM DUAL;
-- Hello\005CnWorld
```

Oracle **does** escape the ASCII backslash to `\005C`. IvorySQL master already matches Oracle; this PR would have introduced an incompatibility.

Thanks for the careful check.

PR #2065, opened 2026-09-16
Fix MONTHS_BETWEEN dropping the time difference for same-day inputs

### LINKED ISSUE #2064: MONTHS_BETWEEN ignores the time-of-day difference for same-day inputs
## Bug Report

**Describe the bug:**
`sys.months_between` (`contrib/ivorysql_ora/src/builtin_functions/datetime_datatype_functions.c`)

Oracle's `MONTHS_BETWEEN` incorporates the time of day of **both** dates into the
fractional-month computation:

```
(y1-y2)*12 + (m1-m2) + (day1-day2)/31 + (time1-time2)/31
```

When both arguments fall on the **same calendar day** (`d1 == d2`), currently the
time-of-day term is gated out by `if (d1 != d2)`, so the time difference is silently
dropped and the function returns 0.

**Reproduction:**
```sql
SELECT MONTHS_BETWEEN(timestamp '2018-08-09 08:30', timestamp '2018-08-09 09:00');
-- Oracle: -0.000672...  (30 minutes / 31 days)
-- current IvorySQL: 0
```

**Fix:**
Always add the time-of-day term in the fractional branch.  The integer shortcut
when both dates are the last day of their months is unchanged.

**Tests:**
7 regression fixtures appended to `ora_datetime_datatype_functions`: same-day
±30 minutes against the identical microsecond-scale arithmetic (deterministic,
no floating-point display dependence), identical instants, sign checks,
cross-month day-change invariance, and an integer month difference.  Existing
pinned `MONTHS_BETWEEN` expectations are unaffected.

jiangdaoli11:
Fixes #2064

## Summary

Oracle's `MONTHS_BETWEEN` incorporates the time of day of both dates into the
fractional-month computation:

```
(y1-y2)*12 + (m1-m2) + (day1-day2)/31 + (time1-time2)/31
```

`months_between()` (in `datetime_datatype_functions.c`) gated the time term on
`d1 != d2`, so when both arguments fall on the same calendar day the time
difference was silently dropped:

```sql
SELECT MONTHS_BETWEEN(timestamp '2018-08-09 08:30', timestamp '2018-08-09 09:00');
-- Oracle: -0.000672...  (30 minutes / 31 days)
-- before this fix: 0
```

The time term is now always added.  The integer shortcut when both dates are
the last day of their months is unchanged.

## Changes

- `src/builtin_functions/datetime_datatype_functions.c`: always add the
  time-of-day term in the fractional branch
- `sql/ora_datetime_datatype_functions.sql` / `expected/*.out`: 7 fixtures —
  same-day ±30 minutes against the identical microsecond-scale arithmetic
  (deterministic, no floating-point display dependence), identical instants,
  sign checks, cross-month day-change invariance, integer month difference

## Verification

Existing pinned `MONTHS_BETWEEN` expectations (e.g.
`-333.483870967742`, `0.25739247311828`) are unaffected; the new fixtures are
pure boolean assertions that pass with the fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `MONTHS_BETWEEN` calculations to account for time-of-day differences, including timestamps on the same calendar day.
  - Corrected fractional-month results, including positive and negative differences, equal timestamps, cross-month comparisons, and exact whole-month intervals.
  - Added regression coverage to help ensure consistent and accurate date-time calculations.


bigplaice:

muzimu217:
Thanks for this fix and for the clean reproduction in #2064. The defect is real — I confirmed on a local master build (oracle compatible mode) that `months_between(timestamp '2018-08-09 08:30', timestamp '2018-08-09 09:00')` currently returns `0`, and that the time-of-day term is genuinely reachable through the `months_between(sys.oradate, sys.oradate)` signature (e.g. `months_between(timestamp '2018-08-10 09:00', timestamp '2018-08-09 08:30')` = `0.03293010752688172` = `1/31 + 1800s/31 days` today, so the newly reachable branch will actually be exercised by these fixtures). I also like the display-independent fixture style — pinning equality instead of raw float output is the right instinct. Two follow-ups, one of which I believe will make the new fixtures fail as committed:

**1. Two of the seven new fixtures should emit `f`, not `t` — the RHS is numeric arithmetic, not "the identical arithmetic"**

With the patch applied, the function computes the same-day case as

```c
result += ((time1 - time2) / 31.0) / USECS_PER_DAY;
```

i.e. two successively rounded float8 divisions. The fixture RHS, however,

```sql
= -1800000000.0 / 31.0 / 86400000000.0
```

is a chain of untyped **numeric** literals, so PostgreSQL evaluates it in numeric arithmetic and only converts to float8 when resolving `float8 = numeric`. For these operands the two paths disagree in the last bit. I verified the mechanics on the master build, and cross-checked the C expression's exact IEEE-754 evaluation independently in Python (the two divisions are separate statements, so no FMA contraction applies on our builds):

```
C-order two-step float8   : 0xBF46058160581606  (-0.0006720430107526882)
numeric-literal RHS       : 0xBF46058160581605
equality                  : f
```

and symmetrically for the positive fixture (`0x3F46058160581606` vs `0x3F46058160581605`). So the ±30-minute equality fixtures should print `f` against the committed expected `t`. To be explicit about my own uncertainty: I have not run the patched build, so this is derived from the expression's semantics rather than observed end-to-end — but since it's deterministic, a CI run should confirm it immediately.

The minimal fix is to make the RHS the same float8 arithmetic, e.g.

```sql
select months_between(timestamp '2018-08-09 08:30', timestamp '2018-08-09 09:00')
       = -1800000000.0::float8 / 31.0::float8 / 86400000000.0::float8 as ok from dual;
```

which I verified compares bitwise-equal (`t`), or to use a tolerance comparison (`abs(...) < 1e-15`). The other five fixtures don't mix a float8 LHS with a numeric RHS and are mechanically fine as written.

**2. The last-day shortcut still drops the time-of-day — is the resulting asymmetry intended? If so, worth pinning**

The both-last-days branch is unchanged, so on master today:

```sql
select months_between(timestamp '2018-08-31 08:30', timestamp '2018-08-31 09:00');  -- 0
select months_between(timestamp '2018-08-31 23:00', timestamp '2018-07-31 01:00');  -- 1
```

After this fix, an identical 30-minute gap therefore yields `-0.000672...` on Aug 30 but `0` on Aug 31, and a 22-hour gap between two month-end days yields a bare integer. The Oracle 19c docs say "if date1 and date2 are either the same day of the month or both last days of months, then the result is always an integer", which supports keeping the shortcut for the last-days case (interestingly, the same sentence would predict an integer for same-day inputs too, while actual Oracle — the basis of #2064 and this patch — returns fractional there, so the doc wording is clearly not literal behavior). If your understanding from real Oracle is that both-last-days is integer regardless of time, I think one small fixture pinning that (e.g. `months_between(timestamp '2018-08-31 08:30', timestamp '2018-08-31 09:00') = 0`) would document the asymmetry as intentional and protect it from being "fixed" by a future patch. If you've observed fractional behavior there on real Oracle instead, I'd genuinely like to know, since that would suggest the shortcut branch wants the same treatment.

Happy to help test either follow-up — I have a comparison environment set up for exactly this function.

PR #2066, opened 2026-09-16
fix: avoid platform-sized overflow in DBMS_LOCK.SLEEP

### LINKED ISSUE #2069: DBMS_LOCK.SLEEP overflows long delays on Windows
### What happened?

`DBMS_LOCK.SLEEP` stores the requested delay in microseconds using C `long`. On 64-bit Windows, `long` remains 32 bits, so valid delays above roughly 2147 seconds overflow before the wait loop. A NaN input is also converted to an integer without an explicit validity check.

Valid delays within the package's accepted range should remain interruptible and should behave consistently on 32-bit-long and 64-bit-long platforms. NaN should be rejected explicitly.

On a 64-bit Windows build, execute `SELECT sys.dbms_lock_sleep(2200);` with a short `statement_timeout`. The microsecond countdown exceeds `LONG_MAX`; it should instead remain active until the timeout cancels it. `SELECT sys.dbms_lock_sleep('NaN'::float8);` should return a clear validation error.

### IvorySQL version

### Compatible mode

oracle

### Server encoding

UTF8

The issue is caused by the platform width of `long`; PostgreSQL's `int64` is appropriate for the total countdown while each `pg_usleep` chunk can remain a bounded `long`.

### Before submitting

btlqql:
DBMS_LOCK.SLEEP stores its microsecond countdown in long. On Windows, long is 32 bits, so valid delays longer than about 2147 seconds overflow before the wait loop. Use int64 for the full countdown, retain bounded long chunks for pg_usleep, and reject NaN before converting it to an integer.

Add a statement-timeout regression for a 2200-second request and a NaN regression. The timeout prevents the test from waiting for the full delay.

Validation: Linux build/install passed; targeted `dbms_lock` Oracle regression passed in [run 35060670431](https://github.com/btlqql/IvorySQL/actions/runs/35060670431).

Assisted-by: OpenAI:GPT-6

Closes #2069


bigplaice:

PR #2067, opened 2026-09-16
fix: detach overlapping DELETEXML matches before freeing nodes

### LINKED ISSUE #2070: DELETEXML can free overlapping XPath matches twice
### What happened?

`DELETEXML` frees each XPath match immediately. If a union expression selects an ancestor and its descendant, freeing the ancestor also frees the descendant that remains in the XPath node array, so a later loop iteration accesses an already freed node.

All selected nodes should be removed safely, including overlapping ancestor/descendant selections.

```sql
SELECT deletexml(XMLType('<r><a><b/></a><keep/></r>'), '//a | //b');
```

The result should be `<r><keep/></r>` and must not access the descendant after its ancestor subtree is freed. The same applies to repeated subtrees and namespace-qualified expressions.

### IvorySQL version

### Compatible mode

oracle

### Server encoding

UTF8

### Before submitting

btlqql:
DELETEXML frees each selected node immediately. When an XPath union selects both an ancestor and its descendant, freeing the ancestor also frees the descendant that the next iteration still accesses.

Detach all selected nodes before freeing any of them, so each selected subtree is released once. Add nested-node, repeated-subtree and namespace-qualified regression cases shared by both overloads.

Validation: Linux build/install passed; targeted `ora_xml_functions` Oracle regression passed in [run 35060670431](https://github.com/btlqql/IvorySQL/actions/runs/35060670431).

Assisted-by: OpenAI:GPT-6

Closes #2070


bigplaice:

PR #2068, opened 2026-09-16
fix: preserve pending DBMS_OUTPUT text after buffer overflow

### LINKED ISSUE #2071: Failed DBMS_OUTPUT.PUT_LINE mutates pending output
### What happened?

When `DBMS_OUTPUT.PUT_LINE` combines pending `PUT` text with its argument, it appends the argument before checking the completed buffer limit. If the capacity check raises `ORU-10027`, the failed argument remains in the pending line and is emitted by a later `NEW_LINE`.

A failed `PUT_LINE` should not mutate the pending `PUT` text.

```sql
SELECT sys.ora_dbms_output_enable(2000);
SELECT sys.ora_dbms_output_put_line(repeat('x', 1998));
SELECT sys.ora_dbms_output_put('a');
SELECT sys.ora_dbms_output_put_line('bc'); -- ORU-10027
SELECT length((sys.ora_dbms_output_get_line()).line); -- drain 1998 bytes
SELECT sys.ora_dbms_output_new_line();
SELECT (sys.ora_dbms_output_get_line()).line;
```

The final line should be `a`; current `master` leaves the failed `bc` appended.

### IvorySQL version

### Compatible mode

oracle

### Server encoding

UTF8

### Before submitting

btlqql:
With 1998 bytes already buffered under a 2000-byte limit, PUT('a') followed by PUT_LINE('bc') raises ORU-10027 but leaves 'abc' in the pending line. Draining the completed output and calling NEW_LINE then emits text from the failed operation.

Check the completed line's capacity before appending to the pending PUT text. Reuse that capacity check when adding a completed line and cover the failure/recovery sequence in the DBMS_OUTPUT regressions.

Validation: Linux build/install passed; targeted `ora_dbms_output` Oracle regression passed in [run 35060670431](https://github.com/btlqql/IvorySQL/actions/runs/35060670431).

Assisted-by: OpenAI:GPT-6

Closes #2071


bigplaice:

PR #2073, opened 2026-09-16
Fix out-of-bounds read in ora_asciistr on truncated UTF-8 input

### LINKED ISSUE #2072: ora_asciistr out-of-bounds read on truncated multi-byte UTF-8 sequence
### What happened?

`ora_asciistr` reads past the end of the input text buffer when the input
contains a truncated multi-byte UTF-8 sequence. The loop condition only checks
`str < end`, but the two/three/four-byte branches dereference `str[1]`,
`str[2]`, `str[3]` without verifying that enough bytes remain before `end`.

For example, a leading byte that starts a 4-byte sequence (0xF0..0xF4) followed
by only one continuation byte (then end-of-buffer) causes `str[2]` and `str[3]`
to be read from whatever memory lies past the text datum — a heap/stack
out-of-bounds read.

- **Out-of-bounds read** past the text datum buffer. Under ASAN/valgrind this is
  flagged as a heap-buffer-overflow; in production it reads adjacent memory
  (other palloc chunks, freed memory, etc.).
- The leaked bytes are mixed into the `codePoint` and emitted as a `\xxxx`
  escape, so the function returns garbage instead of erroring.
- Any user with EXECUTE privilege on `ora_asciistr` can trigger this, making it
  a denial-of-service / information-disclosure vector.

The input must be a `text`/`varchar2` value whose byte length does not match a
complete UTF-8 sequence. Because the server encoding is usually UTF-8, the
truncation has to be constructed at the byte level, e.g. via `convert_from` with
a raw bytea whose last code point is incomplete:

```sql
-- 4-byte leader (U+1F600 grin) followed by only 1 continuation byte
SELECT ora_asciistr(convert_from(
    E'\\xF0\\x9F\\x98'::bytea,  /* truncated: missing 4th byte */
    'UTF8'));
-- Expected: error "Invalid bytes"
-- Actual (before fix): out-of-bounds read, returns garbage \xxxx
```

### IvorySQL version

master

### Compatible mode

oracle

### Server encoding

UTF8

_No response_

### OS version

<details>

```console
$ cat /etc/os-release
# paste output here
$ uname -a
# paste output here
```

</details>

### Before submitting

- [x] I searched existing issues and found no duplicate

enkilee:
fix #2072 

Fixes an out-of-bounds buffer read in `ora_asciistr` when the input text
contains a truncated multi-byte UTF-8 sequence. The two/three/four-byte
branches now verify that enough bytes remain before `end` before dereferencing
`str[1]`/`str[2]`/`str[3]`.

fix:
Added a remaining-bytes check at the top of each multi-byte branch:

```c
} else if ((c & 0xE0) == 0xC0) {
    /* Two-byte UTF-8 sequence */
    if (end - str < 2)
        ereport(ERROR,
                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
                 errmsg("Invalid bytes")));
    codePoint = ((c & 0x1F) << 6) | (str[1] & 0x3F);
    ...
} else if ((c & 0xF0) == 0xE0) {
    /* Three-byte UTF-8 sequence */
    if (end - str < 3)
        ereport(ERROR,
                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
                 errmsg("Invalid bytes")));
    ...
} else if ((c & 0xF8) == 0xF0) {
    /* Four-byte UTF-8 sequence */
    ...
    if (end - str < 4)
        ereport(ERROR,
                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
                 errmsg("Invalid bytes")));
    ...
}
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of incomplete UTF-8 sequences in character conversion.
  - Invalid input that ends prematurely now returns a clear “Invalid bytes” error instead of being decoded incorrectly.


bigplaice:

PR #2080, opened 2026-09-16
fix(build): make headerscheck and cpluspluscheck pass on Oracle headers

### LINKED ISSUE #2075: headerscheck/cpluspluscheck fail due to Oracle header dependency omissions
## What happened?

On current `master` (`03b24b1c468`), the official header-quality targets both exit non-zero in a Docker Ubuntu 22.04 build environment.

`make headerscheck` reports:

```text
In file included from tmp_headerscheck_c/src_bin_psql_mainloop_h.c:2:
src/bin/psql/mainloop.h:15:14: error: unknown type name 'Ora_psqlScanCallbacks'
   15 | extern const Ora_psqlScanCallbacks Ora_psqlscan_callbacks;
```

`mainloop.h` includes `fe_utils/psqlscan.h`, but `Ora_psqlScanCallbacks` is declared by `oracle_fe_utils/ora_psqlscan.h`.

The same target also reports an Oracle parser generated-header dependency failure:

```text
In file included from tmp_headerscheck_c/src_backend_oracle_parser_ora_gramparse_h.c:2:
src/backend/oracle_parser/ora_gramparse.h:32:10: fatal error: ora_gram.h: No such file or directory
   32 | #include "ora_gram.h"
```

`headerscheck` runs `submake-generated-headers` first, but that prerequisite does not generate `src/backend/oracle_parser/ora_gram.h`.

`make cpluspluscheck` reports:

```text
In file included from tmp_headerscheck_cpp/src_include_commands_packagecmds_h.cpp:3:
src/include/commands/packagecmds.h:54:59: error: expected ',' or '...' before 'namespace'
   54 | extern void IsTherePackageInNamespace(char *new_name, Oid namespace);
```

`namespace` is a C++ keyword. `cpluspluscheck` also reproduces the `mainloop.h` unknown-type error.

These confirmed errors are sufficient for both official targets to fail. The run emitted further independent header-check errors; they are outside this report's scope.

`make headerscheck` and `make cpluspluscheck` should complete successfully on current master after configuration.

Headers checked as standalone should include or forward-declare the types they expose. The generated-header prerequisite should produce every generated header needed by a checked source header, or intentionally non-standalone headers should be explicitly excluded. Headers tested as C++ should not use C++ reserved words as identifiers.

In the project Docker development image on Linux:

```bash
./configure --without-readline
make headerscheck
make cpluspluscheck
```

For a Windows checkout mounted into Docker, normalize CRLF only in a temporary copy before configuring, because Linux cannot execute CRLF shell scripts directly.

Observed environment:

- Docker Engine: Linux/x86_64
- Build image: `ivorysql-dev:latest` (Ubuntu 22.04)
- Compiler: GCC/G++ 11.4.0
- Commit: `03b24b1c468` (`master`)

## IvorySQL version

master

## Compatible mode

Both. This is a build-time header check; no server was started.

## Server encoding

Not applicable; no server was started.

Related but not duplicate:

- #1387 / #1388 only add generated build outputs to `.gitignore`; they do not fix the header-check targets.
- PR #1209 addresses a generated `psqlplusparse.c` warning, not standalone generated headers.

- [x] I confirmed the problem still happens on current `master`.

89799969:
## Summary

Fixes #2075.

`make headerscheck` and `make cpluspluscheck` both fail on current master due to three independent Oracle header issues:

1. **`mainloop.h` missing include**: declares `Ora_psqlScanCallbacks` but only includes `fe_utils/psqlscan.h`. The type lives in `oracle_fe_utils/ora_psqlscan.h`.

2. **`packagecmds.h` C++ keyword**: `IsTherePackageInNamespace(char *new_name, Oid namespace)` uses `namespace` as a parameter name, which is a C++ reserved word and breaks `cpluspluscheck`.

3. **`ora_gram.h` not generated for headerscheck**: `ora_gramparse.h` includes the bison-generated `ora_gram.h`, but `submake-generated-headers` only produces `parser/gram.h`, not `oracle_parser/ora_gram.h`. The `headerscheck` script also lacks the oracle_parser build-dir include and exclusion entries that `parser/` already has.

## Changes

- `src/bin/psql/mainloop.h`: include `oracle_fe_utils/ora_psqlscan.h`
- `src/include/commands/packagecmds.h`: rename `namespace` parameter to `namespaceId`
- `src/backend/Makefile`: add `oracle_parser/ora_gram.h` to `generated-headers`
- `src/tools/pginclude/headerscheck`: exclude `ora_gram.h` (bison output) and add `src/backend/oracle_parser/*` to `EXTRAINCLUDES`

## Verification

The three error messages quoted in #2075 each map to one of the above fixes. After applying them:

- `mainloop.h` standalone include resolves `Ora_psqlScanCallbacks`
- `packagecmds.h` compiles as C++ (`namespace` no longer used as an identifier)
- `ora_gramparse.h` can find the generated `ora_gram.h` via the oracle_parser build dir

## Summary by CodeRabbit

- **Chores**
  - Improved build dependency tracking for Oracle-compatible parsing components.
  - Updated header validation to recognize Oracle parser headers and apply the appropriate include paths.
  - Refined internal declarations and callback header integration to support consistent compilation across the project.


bigplaice:

89799969:
Checking in — still relevant and ready for review. Happy to rebase or adjust if maintainers prefer.

PR #2094, opened 2026-09-16
fix(utl_file): preserve multibyte characters in PUTF formats

### LINKED ISSUE #2092: UTL_FILE.PUTF can corrupt multibyte format strings in non-UTF8 target encodings
## What happened?

`UTL_FILE.PUTF` scans the encoded format string byte-by-byte and interprets the byte pair `0x5c 0x6e` as the `\n` format escape without first checking character boundaries.

That breaks valid multibyte characters in encodings where a character can contain `0x5c` as a trailing byte. For example, `衆n` encoded as GB18030 is `d0 5c 6e`: the current scanner can treat the last byte of `衆` plus the following ASCII `n` as `\n`, corrupting the original character. The same class of problem is reproducible with GBK, SJIS, and BIG5 target encodings.

PUTF should interpret `%s`, `%%`, and `\n` only at character boundaries. Multibyte characters should be written unchanged in the target file encoding.

For a GB18030 file, writing the format string `衆n` should produce:

```text
d05c6e
```

Reproduced on upstream `master` at `03b24b1c4688266f738394f443f63d4d80ab6379` before the fix, in Oracle mode with UTF8 server encoding:

```sql
INSERT INTO sys.utl_file_directory(dirname, dir)
SELECT 'putf_repro', current_setting('data_directory');

SELECT sys.ora_utl_file_fopen(
  'putf_repro', 'putf-gb18030.dat', 'w', 1024, 'GB18030'
) AS fd \gset

SELECT sys.ora_utl_file_putf(:fd, '衆n');
SELECT sys.ora_utl_file_fclose(:fd);
SELECT encode(pg_read_binary_file('putf-gb18030.dat'), 'hex');
```

The problematic scanner is in `contrib/ivorysql_ora/src/builtin_packages/utl_file/utl_file.c`: after `encode_text()` it advances one byte at a time rather than one encoded character at a time.

## IvorySQL version

master

## Compatible mode

oracle

## Server encoding

UTF8

The target file encoding is explicitly GB18030; the server encoding remains UTF8. Regression coverage should also include other supported multibyte encodings, `%s`/`%%` interactions, four-byte GB18030 input, and byte-limit boundaries.

## Before submitting

- [x] I confirmed the problem still exists in current `master` source.

zmuxuny:
Fixes #2092

## Summary

`UTL_FILE.PUTF` currently scans the already-encoded format buffer one byte at a time. In multibyte encodings, a trailing byte from one character can therefore be mistaken for the start of an ASCII format escape.

This change makes PUTF walk the format string by encoded character boundaries while preserving the existing ASCII formatting semantics.

Specifically, it:

- preserves multibyte characters in GB18030, GBK, SJIS, BIG5, and other supported target encodings;
- interprets `\n`, `%s`, and `%%` only when they begin at a character boundary;
- keeps byte-based `max_linesize` accounting;
- refuses to partially write a multibyte character when the remaining byte budget is too small;
- adds a focused regression suite that checks the actual bytes written to disk.

## Root cause

`ora_utl_file_putf()` calls `encode_text()` first and then advances through the converted buffer with `fpt++`. The parser can therefore examine bytes that are not character starts.

For example, `衆n` encoded as GB18030 is:

```text
d0 5c 6e
```

The old byte scanner can see `5c 6e` and treat it as `\n`, even though `5c` is the second byte of `衆`. The new scanner uses the target encoding's character length before interpreting formatting tokens.

## Testing

The new `utl_file_putf` regression covers:

- GB18030, GBK, SJIS, BIG5, UTF8, and LATIN1 output;
- literal multibyte text adjacent to ASCII `n`;
- `%s`, `%%`, and real `\n` formatting;
- a four-byte GB18030 character;
- exact and insufficient byte limits;
- no partial multibyte write at the line-size boundary.

Validation on the fork:

- focused `utl_file_putf` regression: **1/1 passed**;
- complete `contrib/ivorysql_ora` Oracle regression: **30/30 passed**;
- `git diff --check`: passed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved formatted file output for multibyte characters across supported encodings, preventing characters from being misinterpreted as formatting directives.
  - Enforced byte-based buffer limits correctly, including clear errors when output exceeds available space and prevention of partially written characters.
  - Improved handling of format tokens, escaped percent signs, substitution arguments, null values, and four-byte characters.

- **Tests**
  - Added comprehensive regression coverage for multibyte output, encoding compatibility, formatting behavior, byte limits, and temporary-file cleanup.


bigplaice:

PR #1727, opened 2026-08-23
Add UTL_RAW byte-manipulation functions

xuhuicai:
Implement the UTL_RAW functions at or below the "moderate" complexity level, covering the commonly used raw byte operations:

- LENGTH, CONCAT (2-12 args), REVERSE, COPIES, XRANGE, SUBSTR (Oracle position semantics: 0 = 1, negative counts from the end, out-of-range yields NULL)
- BIT_AND, BIT_OR, BIT_XOR, BIT_COMPLEMENT (equal-length arguments)
- COMPARE with optional single-byte pad (-1/0/1)
- TRANSLATE with byte mapping and deletion for unmatched bytes
- CAST_TO_VARCHAR2 (database-encoding interpretation)
- CONVERT (Oracle argument order r, to_charset, from_charset)

Assisted-by: Cursor
Percentage of AI-generated code: 95%

Closes #1726 

## Summary by CodeRabbit

* **New Features**
  * Expanded the `UTL_RAW` package with functions for concatenation, length checks, reversal, repetition, ranges, substrings, bitwise operations, comparison, translation, and text conversion.
  * Added support for character-set conversion based on the database encoding.
  * Improved comparisons with zero-byte padding when the padding value is omitted or null.

* **Bug Fixes**
  * Added validation and clearer error handling for invalid lengths, ranges, padding, encoding, and other inputs.

* **Tests**
  * Added comprehensive coverage for normal, boundary, null, encoding, and error scenarios.

PR #2095, opened 2026-09-16
feat(utl_raw): add binary integer conversion functions

### LINKED ISSUE #2093: Support UTL_RAW binary integer conversion functions
## Feature Request

Add Oracle-compatible `UTL_RAW.CAST_FROM_BINARY_INTEGER` and `UTL_RAW.CAST_TO_BINARY_INTEGER` support to `ivorysql_ora`.

The current UTL_RAW package defines `big_endian`, `little_endian`, and `machine_endian` constants, but does not expose the corresponding binary-integer conversion functions.

Oracle documents the API as:

```sql
UTL_RAW.CAST_FROM_BINARY_INTEGER(
    n          IN BINARY_INTEGER,
    endianess  IN PLS_INTEGER DEFAULT BIG_ENDIAN
) RETURN RAW;

UTL_RAW.CAST_TO_BINARY_INTEGER(
    r          IN RAW,
    endianess  IN PLS_INTEGER DEFAULT BIG_ENDIAN
) RETURN BINARY_INTEGER;
```

At the proposal baseline, IvorySQL did not expose `BINARY_INTEGER` / `PLS_INTEGER` spellings in package declarations; that lower-level compatibility gap is now tracked separately by #2114 / #2115. There is also a call-resolution difference: `numeric -> int4` is assignment-only rather than an implicit function-call cast, so an `INTEGER` source formal rejects fractional `NUMBER` actual arguments before the wrapper can perform Oracle-style integer coercion. That resolver gap is tracked separately by #2112 / #2113.

For that reason the IvorySQL compatibility wrapper uses `NUMBER` only for the `CAST_FROM_BINARY_INTEGER` source argument, then explicitly casts `n::pg_catalog.int4` at the C boundary. This preserves Oracle-observed calls such as `CAST_FROM_BINARY_INTEGER(1.5)` while still enforcing the signed 32-bit domain. `endianess` and the decode result remain `INTEGER` mappings of Oracle's PLS_INTEGER/BINARY_INTEGER types.

Oracle's documented default is the package constant `BIG_ENDIAN`. At the proposal baseline, PL/iSQL did not resolve a same-package constant in this default-expression position when the caller omitted the argument, so the IvorySQL package uses the semantically equivalent literal `1`. The underlying default-scope issue is tracked separately by #2099 / #2100.

Expected behavior is based on Oracle's UTL_RAW documentation and a reference run against Oracle AI Database 26ai Free 23.26.3.0.0. Reference behavior includes:

- default byte order is big-endian;
- selectors `1`, `2`, and `3` represent big-endian, little-endian, and machine-endian;
- signed 32-bit values are emitted as four RAW bytes;
- fractional NUMBER actual arguments are coerced to the signed 32-bit integer domain before encoding;
- one-to-four-byte RAW inputs are accepted by `CAST_TO_BINARY_INTEGER`;
- RAW inputs longer than four bytes raise a conversion error;
- NULL data or an explicitly NULL endian selector returns NULL;
- invalid non-NULL endian selectors raise an error.

Examples:

```text
CAST_FROM_BINARY_INTEGER(16909060, 1) -> 01020304
CAST_FROM_BINARY_INTEGER(16909060, 2) -> 04030201
CAST_FROM_BINARY_INTEGER(1.5, 1)      -> 00000002
CAST_TO_BINARY_INTEGER(01020304, 1)   -> 16909060
CAST_TO_BINARY_INTEGER(01020304, 2)   -> 67305985
CAST_TO_BINARY_INTEGER(010203, 1)     -> 66051
```

This proposal is intentionally limited to the binary-integer conversions and does not duplicate the byte-manipulation functions being developed separately in #1727.

zmuxuny:
Closes #2093

## Problem

IvorySQL's `UTL_RAW` package already defines the Oracle byte-order constants, but it does not implement `CAST_FROM_BINARY_INTEGER` or `CAST_TO_BINARY_INTEGER`. Applications using these Oracle APIs therefore cannot run unchanged.

## Summary

This change adds:

- `UTL_RAW.CAST_FROM_BINARY_INTEGER`
- `UTL_RAW.CAST_TO_BINARY_INTEGER`

The implementation:

- supports big-endian, little-endian, and machine-endian selectors;
- uses explicit unsigned shifts instead of unaligned native integer loads;
- emits a four-byte two's-complement representation for signed 32-bit values;
- accepts one-to-four-byte RAW inputs when decoding;
- treats a full four-byte input as signed and shorter inputs as zero-extended;
- returns NULL for NULL data or an explicitly NULL endian selector;
- rejects unsupported non-NULL selectors and RAW inputs longer than four bytes;
- registers the C source in both Makefile and Meson builds;
- adds focused regression coverage and package-level calls.

## Oracle API mapping

Oracle documents the source argument of `CAST_FROM_BINARY_INTEGER` as `BINARY_INTEGER`, the endian selector as `PLS_INTEGER`, and the decode result as `BINARY_INTEGER`.

At this PR's original base, IvorySQL did not expose `BINARY_INTEGER` / `PLS_INTEGER` spellings in package declarations; #2114 / #2115 address those spellings separately. More importantly for this self-contained implementation, the package resolver at that base used implicit coercion while `numeric -> int4` is assignment-only. Declaring the source argument as `INTEGER` therefore rejected fractional `NUMBER` actual arguments before the wrapper could perform the integer conversion that Oracle accepts; #2112 / #2113 address that resolver behavior separately.

To preserve the observed call semantics without widening the C ABI, only the public `CAST_FROM_BINARY_INTEGER` source parameter uses `NUMBER`; the wrapper immediately applies `n::pg_catalog.int4`, which both performs the Oracle-observed rounding and enforces the signed 32-bit range. The endian selector and `CAST_TO_BINARY_INTEGER` return remain `INTEGER`, with explicit `pg_catalog.int4` C entry points.

Oracle's documented default is the package constant `BIG_ENDIAN`. At this PR's original base, PL/iSQL did not resolve a same-package constant in this default-expression position when the caller omitted the argument; a validation run with `DEFAULT BIG_ENDIAN` reproduced `"big_endian": invalid identifier`. The package therefore uses the semantically equivalent `DEFAULT 1`. The underlying default-scope fix is tracked separately by #2099 / #2100.

## Oracle reference verification

Behavior was compared against Oracle AI Database 26ai Free 23.26.3.0.0 on x86_64. Fixed vectors include:

```text
CAST_FROM_BINARY_INTEGER(16909060, 1) -> 01020304
CAST_FROM_BINARY_INTEGER(16909060, 2) -> 04030201
CAST_FROM_BINARY_INTEGER(1.5, 1)      -> 00000002
CAST_FROM_BINARY_INTEGER(-1.5, 1)     -> FFFFFFFE
CAST_TO_BINARY_INTEGER(01020304, 1)   -> 16909060
CAST_TO_BINARY_INTEGER(01020304, 2)   -> 67305985
CAST_TO_BINARY_INTEGER(010203, 1)     -> 66051
```

The reference run also covered one-to-three-byte RAW inputs, NULL data/selectors, invalid endian selectors, signed boundaries, and overlength RAW input.

## Testing

Validation run:
https://github.com/zmuxuny/IvorySQL/actions/runs/35175531427

Results:

- build and install: passed;
- focused `utl_raw_integer` regression: **1/1 passed**;
- complete `contrib/ivorysql_ora` Oracle regression: **30/30 passed**;
- package regression covers fractional NUMBER coercion and INT32_MIN / INT32_MAX boundaries;
- `git diff --check`: passed.

The validation workflow applies a CI-only exclusion for the unrelated baseline `pg_rewind` linkage failure in its disposable working tree. That exclusion is not part of this feature branch.

The final squashed commit has the same tree hash (`81785d4e65c622cac9eb7b81578a4089b5fd7984`) as the validated pre-squash tree, so the tested source tree is unchanged by the history cleanup.

A later cross-PR validation on upstream `master` `63fb0bfe6159fb91b03012eea09f3db7cd46db8c` stacked this PR together with #2094, #2100, #2113, and the #2115 SETOF fix. Build/install, the combined focused PL/iSQL tests, the full PL/iSQL suite, the full Oracle core suite, and the full `contrib/ivorysql_ora` suite all passed: https://github.com/zmuxuny/IvorySQL/actions/runs/35320091373.

## Meson regression note

The new C source is registered in `contrib/ivorysql_ora/meson.build`.

Since this PR was opened, #1501 has added Oracle-mode Meson test infrastructure. The current harness, however, classifies a test as Oracle-mode only when its source directory is under `src/oracle_test/**`; `contrib/ivorysql_ora` is outside that path, and its Meson file still does not register the existing `ORA_REGRESS` suite. The regression therefore remains registered through the Makefile `ORA_REGRESS` path.

Adding only `utl_raw_integer` as an ordinary Meson `tests +=` regression would still dispatch it through the PostgreSQL-mode harness. Extending Oracle-mode Meson registration to `contrib/ivorysql_ora` is a repository-level follow-up, not part of this feature.

## Scope

This PR intentionally implements only the binary-integer conversions. It does not duplicate the byte-manipulation functions in #1727, although both changes touch the UTL_RAW package definition and may require a small conflict resolution depending on merge order.

## Summary by CodeRabbit

* **New Features**
  * Added `UTL_RAW` support for converting signed 32-bit integers to and from four-byte RAW values.
  * Supports big-endian, little-endian, and machine-endian formats.
  * Handles NULL, empty, short, and boundary values with Oracle-compatible behavior.
  * Added validation for invalid byte-order selectors, oversized RAW values, and out-of-range integers.

* **Tests**
  * Added comprehensive regression coverage for conversions, round trips, boundary cases, and error handling.


bigplaice:

PR #2100, opened 2026-09-17
fix(plisql): evaluate package defaults in declaration scope

### LINKED ISSUE #2099: PL/iSQL package subprogram defaults cannot resolve package-scope references
## Bug

A package subprogram formal parameter cannot currently use a constant or function from the declaring package in its default expression when the caller omits that argument.

Minimal reproduction on upstream `master` at `9cd4f0f112ee8cb2ca201666a7c5867b2637f3df` before the fix:

```sql
CREATE OR REPLACE PACKAGE pkg_default_scope_test IS
  c_default CONSTANT INTEGER := 42;
  FUNCTION get_value(p_value IN INTEGER DEFAULT c_default) RETURN INTEGER;
END pkg_default_scope_test;
/

CREATE OR REPLACE PACKAGE BODY pkg_default_scope_test IS
  FUNCTION get_value(p_value IN INTEGER DEFAULT c_default) RETURN INTEGER IS
  BEGIN
    RETURN p_value;
  END;
END pkg_default_scope_test;
/

SELECT pkg_default_scope_test.get_value();
```

Current result:

```text
ERROR:  "c_default": invalid identifier
```

Passing an explicit value works, so the failure is specific to omitted/defaulted arguments.

## Expected behavior

The default expression should be resolved in the lexical scope of the package/subprogram that declares it. Package constants and package functions should therefore be usable as defaults.

Defaults also need call-time semantics rather than parse-time substitution. For example, if a default calls a package function that increments package state, the function should run once for each call that omits the argument, and should not run when an explicit actual argument is supplied.

## Root cause

`plisql_get_subprocfunc_argdefaults()` reparses the saved formal default and calls `transformExpr()` using the caller's `ParseState`. This loses the declaring package namespace. The package/subprocedure lookup path also expands the default into the caller's argument list before execution, which is too early for defaults that depend on package state or otherwise need per-call evaluation.

This was encountered while implementing Oracle-compatible package APIs such as `UTL_RAW.CAST_FROM_BINARY_INTEGER`, whose documented default is the package constant `BIG_ENDIAN`. The current workaround in #2095 uses the equivalent literal `1` because `DEFAULT BIG_ENDIAN` fails when the argument is omitted.

## Desired scope

- preserve explicit argument behavior, including explicit `NULL`;
- resolve omitted package defaults in the declaring package/subprogram scope;
- evaluate such defaults for each omitted call at execution time;
- preserve named-notation calls where a non-trailing formal is omitted;
- keep ordinary `pg_proc` and nested-subprogram default handling unchanged unless required.

zmuxuny:
Fixes #2099

## Problem

PL/iSQL package subprogram defaults are currently expanded in the caller's parse context. As a result, a formal default that refers to a same-package constant or function fails when the caller omits the argument:

```sql
FUNCTION get_value(p_value IN INTEGER DEFAULT c_default) RETURN INTEGER;
SELECT pkg_default_scope_test.get_value();
```

Current master reports:

```text
ERROR:  "c_default": invalid identifier
```

The same issue affects Oracle-compatible package APIs whose documented defaults reference package constants, such as `UTL_RAW.BIG_ENDIAN`.

## Root cause

`plisql_get_subprocfunc_argdefaults()` reparses and transforms the saved default with the caller's `ParseState`. That loses the declaring package namespace and also evaluates package-dependent defaults too early.

## Fix

For package routine calls only, this change defers omitted formal defaults until the callee starts executing:

- package lookup keeps the call argument list positional with typed NULL placeholders;
- omitted formal positions are recorded in `FuncExpr.default_argnos`;
- the metadata is preserved through planner rewrites and node serialization;
- after package initialization, `plisql_exec_function()` evaluates the formal parameter's saved `default_val` in the callee estate;
- copied formal default expressions are rebound to the actual package subprogram function so package constants/functions resolve against the correct execution state.

Ordinary `pg_proc` defaults and nested-subprogram defaults keep their existing path.

Explicit values, including explicit `NULL`, are not treated as omitted defaults.

## Regression coverage

`ora_package` now covers:

- same-package constant default;
- explicit value and explicit NULL;
- positional omission;
- named notation omission;
- multiple defaults and a non-trailing omitted formal;
- a package function used as a dynamic default;
- per-call re-evaluation of dynamic defaults;
- explicit actual arguments not invoking the default;
- view serialization/deserialization of a package call with an omitted default.

## Validation

Clean baseline reproduction on upstream `master` (`9cd4f0f112ee8cb2ca201666a7c5867b2637f3df`) with the new regression SQL but without this fix reproduces:

```text
SELECT pkg_default_scope_test.get_value();
ERROR:  "c_default": invalid identifier
```

Validation of this branch:

- full `make -j2`: passed;
- `git diff --check`: passed;
- full Oracle regression schedule: **256/256 passed**;
- `ora_package` specifically passed in that full run.

Dynamic-default checks demonstrate call-time behavior:

```text
dynamic_value()  -> 1
counter_value()  -> 1
dynamic_value(9) -> 9
counter_value()  -> 1
dynamic_value()  -> 2
counter_value()  -> 2
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed package subprogram calls with omitted default arguments.
  * Default values now resolve in the declaring package, including package constants and functions.
  * Preserved the distinction between omitted arguments and explicitly provided `NULL` values.
  * Ensured positional and named arguments, views, and repeated dynamic default evaluations work correctly.

* **Tests**
  * Added regression coverage for package default-argument behavior and related execution scenarios.


bigplaice:

PR #2104, opened 2026-09-17
fix(plisql): handle argument types and OUT/INOUT parameters in autonomous transactions

### LINKED ISSUE #2103: PRAGMA AUTONOMOUS_TRANSACTION: procedures with OUT/INOUT parameters fail, and NaN/Infinity / polymorphic arguments are not serialized correctly
### Title

PRAGMA AUTONOMOUS_TRANSACTION: procedures with OUT/INOUT parameters fail, and NaN/Infinity / polymorphic arguments are not serialized correctly

### IvorySQL Version

`master` at `03b24b1c4688266f738394f443f63d4d80ab6379` (feature introduced in PR #986)

### Related

- Original implementation: PR #986
- Feature overview: issue #1069
- Unrelated existing issues (explicitly out of scope): #1625 (privilege bypass), #1128 (COMMIT inside EXCEPTION blocks)

This report covers three independent defects in PL/iSQL autonomous transactions (`PRAGMA AUTONOMOUS_TRANSACTION`, argument serialization and dispatch in `build_autonomous_call()` in `src/pl/plisql/src/pl_autonomous.c`). Test environment: Oracle-mode session with the dblink extension installed; the outer `CALL`/`SELECT` uses the regular syntax, exactly as in a non-autonomous call.

#### 1. Procedures with OUT/INOUT parameters cannot use autonomous transactions (high severity)

```sql
CREATE OR REPLACE PROCEDURE p_out(IN p_id INT, OUT p_result INT) AS $$
PRAGMA AUTONOMOUS_TRANSACTION;
BEGIN
    p_result := p_id * 2;
END;
$$ LANGUAGE plisql;
CALL p_out(9, NULL);
```

Actual errors (unfixed build):

```
ERROR:  function public.p_out(pg_catalog.int4) does not exist
DETAIL:  No function of that name accepts the given number of arguments.
CONTEXT:  while executing query on unnamed dblink connection
SQL statement "SELECT * FROM dblink('dbname=''...'' host=localhost port=...', 'SET ivorysql.compatible_mode = oracle; SET plisql.inside_autonomous_transaction = true; SELECT public.p_out(9);') AS t(result record)"
```

For an INOUT-only procedure:

```
ERROR:  public.tio(pg_catalog.int4) is a procedure
HINT:  To call a procedure, use CALL.
```

Even if the erroneous `SELECT` path is bypassed, `dblink_exec()` rejects a row-returning CALL with `statement returning results not allowed`. In short, such procedures are completely unusable with autonomous transactions. Root causes:

- function/procedure detection uses `prorettype != VOIDOID` instead of `prokind`, so a procedure with OUT/INOUT parameters is misclassified as a function and dispatched via `SELECT`;
- the generated argument list only contains `fcinfo->args` (IN/INOUT input values) and omits the OUT-parameter placeholders — the Oracle-mode parser requires the complete argument list;
- `dblink_exec` cannot handle the OUT record row returned by CALL.

#### 2. NaN / Infinity special numeric values cannot round-trip (medium severity)

```sql
CREATE OR REPLACE PROCEDURE p_num(p_n NUMERIC, p_f FLOAT8) AS $$
PRAGMA AUTONOMOUS_TRANSACTION;
BEGIN
    NULL;
END;
$$ LANGUAGE plisql;
CALL p_num('NaN'::NUMERIC, 'Infinity'::FLOAT8);
```

Actual error:

```
ERROR:  column "nan" does not exist
CONTEXT:  while executing command on unnamed dblink connection
```

`Infinity` / `-Infinity` (FLOAT4/FLOAT8) likewise fail with `column "infinity" does not exist`. Root cause: numeric arguments are embedded in the remote SQL as bare, unquoted output text, and `NaN`/`Infinity` are not valid SQL literals (the parser does not treat them as numeric constants).

#### 3. Polymorphic arguments (ANYELEMENT/ANYARRAY) fail type resolution (medium severity)

```sql
CREATE OR REPLACE FUNCTION f_poly(p_any ANYELEMENT) RETURN TEXT AS $$
DECLARE
    PRAGMA AUTONOMOUS_TRANSACTION;
BEGIN
    RETURN p_any::TEXT;
END;
$$ LANGUAGE plisql;
SELECT f_poly(5);
```

Actual error:

```
ERROR:  cannot display a value of type anyelement
CONTEXT:  PL/iSQL function f_poly(anyelement) during function entry
```

With a string argument (no explicit cast) it fails with `could not determine polymorphic type because input has type unknown`. Root cause: serialization uses the declared pseudo-type (anyelement/anyarray) from `pg_proc` to pick the output function, instead of resolving the concrete type of the actual call argument.

### Expected behavior

Identical to ordinary (non-autonomous) calls:

1. Procedures with OUT/INOUT parameters are dispatched via `CALL` with the complete argument list (OUT placeholders included, e.g. `CALL p_out(9, NULL)`), and output values are returned to the caller;
2. Any numeric argument value (including NaN/±Infinity) round-trips exactly through the autonomous session;
3. Polymorphic parameters use the resolved concrete type for output/input conversion (`ANYELEMENT` resolved to `int4`, etc.).

### Step to reproduce

1. Install dblink: `CREATE EXTENSION IF NOT EXISTS dblink;`
2. Enable Oracle mode in the session: `SET ivorysql.compatible_mode = oracle;`
3. Run each of the three minimal reproductions above.

jiangdaoli11:
## Summary

Fixes #2103 — three defects in PL/iSQL autonomous transactions (`PRAGMA AUTONOMOUS_TRANSACTION`) related to argument handling in `build_autonomous_call()` (`src/pl/plisql/src/pl_autonomous.c`):

1. **Procedures with OUT/INOUT parameters were unusable.** Function/procedure detection used `prorettype != VOIDOID` instead of `prokind`, so such procedures were dispatched via `SELECT` (fails: `... is a procedure` / `function ... does not exist`), the generated call omitted the OUT placeholders required by the Oracle-mode parser, and `dblink_exec()` rejects the row-returning CALL (`statement returning results not allowed`). Procedures with OUT/INOUT parameters are now dispatched via `CALL` with the full argument list (OUT arguments as `NULL` placeholders), and the output-parameter record is read back through `dblink()` + SPI and returned to the caller.

2. **Special numeric values (NaN / ±Infinity) did not round-trip.** Numeric arguments were embedded as bare, unquoted output text, which is not a valid SQL literal for `NaN`/`Infinity` (`column "nan" does not exist`). Every non-NULL argument is now serialized as a quoted literal with an explicit type cast (`'value'::type`), so values round-trip exactly through the remote session's input function.

3. **Polymorphic arguments (ANYELEMENT/ANYARRAY) failed.** The declared pseudo-type from `pg_proc` was used to pick the output function. The concrete type is now resolved from the actual call argument via `get_fn_expr_argtype()`.

## Changes

- `src/pl/plisql/src/pl_autonomous.c`:
  - `build_autonomous_call()`: function/procedure detection via `prokind`; full-argument iteration using `proallargtypes`/`proargmodes` for procedures (OUT as `NULL`); quoted-literal + typed-cast serialization; per-call polymorphic type resolution.
  - New `get_dblink_oid()` helper (shared cached lookup) and new `execute_autonomous_procedure()` running the record-returning CALL through `dblink()` + SPI and building the output record for the caller.
- Regression tests added to `plisql_autonomous` for special numeric values, OUT/INOUT procedures, and polymorphic arguments; the existing expected output is unchanged.

## Validation

- `plisql_autonomous` regression passes with expected output unchanged (plus the new test cases).
- Manual scenarios verified: NaN/±Infinity numeric args, OUT / INOUT / mixed IN+INOUT+OUT procedures (including OUT-first ordering and NULL handling), polymorphic anyelement/anyarray, plus all previously-working argument types (INT/TEXT/BOOL/DATE/TIMESTAMP/VARCHAR/CHAR/UUID/JSONB/BYTEA/ARRAY/DOMAIN/ENUM/MONEY/defaults).

## Summary by CodeRabbit

* **New Features**
  * Autonomous procedures now support `OUT` and `INOUT` parameters, with updated values returned to callers.
  * Autonomous calls correctly preserve special numeric values, including `NaN`, positive infinity, and negative infinity.
  * Polymorphic parameters are now resolved according to the types of the supplied arguments.
  * Procedures with output values can return results through autonomous execution.


bigplaice:

Issue #1128, opened 2025-12-08
Oracle incompatibility: COMMIT inside EXCEPTION-enabled blocks fails


rophy:
### IvorySQL Version

master (commit 7261f1caeb803b3fc926bf03ca4587fc7d063702) as container

### OS Version (uname -a)

```
Linux 6.8.0-87-generic #88-Ubuntu SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
```

### Configuration options (config.status --config)

```
'--prefix=/home/ivorysql/ivorysql' '--enable-debug' '--enable-cassert' '--with-uuid=e2fs' '--with-libxml' '--with-libxslt'
```

### Current Behavior

COMMIT and ROLLBACK statements fail with the error:
```
cannot commit while a subtransaction is active
```
or
```
cannot roll back while a subtransaction is active
```

This occurs when the COMMIT/ROLLBACK is inside a PL/pgSQL or PL/iSQL block that has an EXCEPTION handler, even if the COMMIT is in the main BEGIN block (not inside the EXCEPTION handler itself).

### Expected behavior/code

In Oracle PL/SQL, COMMIT and ROLLBACK work correctly inside procedures that have EXCEPTION handlers. The expected behavior is that COMMIT should succeed in the main block regardless of whether an EXCEPTION handler exists.

Oracle example that works:
```sql
CREATE OR REPLACE PROCEDURE my_proc AS
BEGIN
    INSERT INTO log_table VALUES ('test');
    COMMIT;  -- Works in Oracle
EXCEPTION
    WHEN OTHERS THEN
        ROLLBACK;
END;
/
```

### Steps to reproduce

```sql
-- Setup
DROP TABLE IF EXISTS test_log CASCADE;
CREATE TABLE test_log (id SERIAL, msg TEXT);

-- Test 1: COMMIT without EXCEPTION block - WORKS
DO $$
BEGIN
    INSERT INTO test_log(msg) VALUES ('no exception block');
    COMMIT;
END;
$$;
-- Result: SUCCESS

-- Test 2: COMMIT with EXCEPTION block - FAILS
DO $$
BEGIN
    INSERT INTO test_log(msg) VALUES ('with exception block');
    COMMIT;  -- ERROR: cannot commit while a subtransaction is active
EXCEPTION
    WHEN OTHERS THEN
        NULL;
END;
$$;
-- Result: ERROR

-- Test 3: Procedure with EXCEPTION handler - FAILS
[...]

PR #2106, opened 2026-09-17
fix(plisql): allow COMMIT/ROLLBACK inside EXCEPTION-enabled blocks

### LINKED ISSUE #2105: COMMIT/ROLLBACK inside EXCEPTION-enabled blocks fails
### IvorySQL Version

current master

### Current Behavior

Executing COMMIT or ROLLBACK inside a PL/iSQL block that has an EXCEPTION handler fails with:

```
ERROR:  cannot commit while a subtransaction is active
```

or

```
ERROR:  cannot roll back while a subtransaction is active
```

PL/iSQL implements exception-enabled blocks by wrapping the whole block body in a subtransaction, which makes the transaction-control statements fail even when they appear in the main BEGIN section and no error has occurred. Oracle PL/SQL allows COMMIT/ROLLBACK both in the main section and inside the exception handler.

### Expected behavior/code

The following Oracle-style code should work:

```sql
CREATE OR REPLACE PROCEDURE my_proc AS
BEGIN
    INSERT INTO log_table VALUES ('test');
    COMMIT;                    -- works in Oracle, fails on IvorySQL
EXCEPTION
    WHEN OTHERS THEN
        INSERT INTO log_table VALUES ('error');
        COMMIT;
END;
/
```

### Steps to reproduce

```sql
CREATE TABLE test_tx (a int);

DO LANGUAGE plisql $$
BEGIN
    BEGIN
        INSERT INTO test_tx VALUES (1);
        COMMIT;                -- ERROR: cannot commit while a subtransaction is active
    EXCEPTION
        WHEN OTHERS THEN
            NULL;
    END;
END;
$$;
```

Related: #1128 covers the same symptom from the Oracle-compatibility angle.

jiangdaoli11:
## Summary

Fixes #2105 (related #1128).

`COMMIT`/`ROLLBACK` inside a PL/iSQL block that has an `EXCEPTION` handler used to fail with `cannot commit while a subtransaction is active` / `cannot roll back while a subtransaction is active`, because exception-enabled blocks wrapped their whole body in a single subtransaction. Oracle allows transaction control both in the main section and inside exception handlers.

## Design

Implement Oracle-style **statement-level rollback** for exception-enabled blocks:

- `exec_stmt_block`: exception-enabled blocks no longer open a block-level subtransaction. A new per-invocation mode (`estate->stmt_subxact_level`) tells statement dispatch to isolate statements.
- `exec_stmt` / `exec_stmt_with_stmt_subxact`: while that mode is active, each ordinary statement runs inside its own short-lived subtransaction, so an error only undoes the failed statement before the matching handler runs. `CALL` (the callee manages its own transactions), `COMMIT`/`ROLLBACK` themselves and nested exception blocks are exempt.
- `exec_transaction`: before `SPI_commit`/`SPI_rollback`, release any still-open statement subtransactions (process-global `active_stmt_subxact_depth`, reset at transaction end), so the transaction-control statement is never executed inside a wrapper subtransaction.

Resulting behavior matches Oracle:

- `COMMIT`/`ROLLBACK` work in the main section, in the handler, inside loops, and in nested procedure calls.
- Only the failing statement is rolled back; earlier statements in the block persist (statement-level, not block-level).

## Changes

- `src/pl/plisql/src/pl_exec.c`, `src/pl/plisql/src/plisql.h` — implementation
- `src/pl/plisql/src/sql/plisql_transaction.sql` — new regression cases (commit/rollback in exception blocks, nested blocks, handler errors, called procedures)
- `src/pl/plisql/src/sql/plisql_trap.sql`, `expected/*.out` — update to the new statement-level semantics

## Test plan

- [x] `make oracle-check` in `src/pl/plisql/src`: all 22 PL/iSQL tests pass

## Summary by CodeRabbit

* **New Features**
  * Added Oracle-style statement-level rollback for PL/iSQL exception-enabled blocks.
  * `COMMIT` and `ROLLBACK` now work within exception blocks, handlers, nested blocks, and procedures.
  * When a statement fails, only that statement is rolled back; earlier successful statements remain preserved.
  * Improved handling of transaction boundaries and propagated errors in exception handlers.

* **Tests**
  * Added extensive coverage for transaction control, nested exception handling, and statement preservation.


bigplaice:

Issue #2053, opened 2026-09-14
Add Oracle-compatible package: DBMS_NETWORK_ACL_ADMIN


yuluo-yx:
Follow-up to #1855, as requested in https://github.com/IvorySQL/IvorySQL/pull/1855#issuecomment-5661876122.

Implement the Oracle-compatible `DBMS_NETWORK_ACL_ADMIN` package to manage network ACLs, then integrate host ACL checks into `UTL_INADDR.GET_HOST_ADDRESS` and `UTL_INADDR.GET_HOST_NAME`.

The reviewer requests checking the invoking user's CONNECT privilege in the ACL assigned to the remote host. Confirm the privilege semantics against the Oracle documentation when implementing the package and enforce the applicable host privilege before name/address resolution. Raise `UTL_INADDR.NETWORK_ACCESS_DENIED` (ORA-24247) when access is denied, and retain `AUTHID CURRENT_USER`.

Include regression coverage for permitted and denied invoking users and host ACL matching. #1855 marks the integration points with TODO comments; it does not implement network ACL management.

Reference: https://docs.oracle.com/en/database/oracle/oracle-database/26/arpls/UTL_INADDR.html

bigplaice:
Thanks for bringing new feature to IvorySQL!

PR #2109, opened 2026-09-17
Add Oracle-compatible UTL_TCP package (plain outbound TCP core)

### LINKED ISSUE #1059: oracle compatibility feature: UTL_TCP
## Feature Request
oracle compatibility feature: UTL_TCP
UTL_TCP enables low-level TCP/IP network communication from the database to remote servers (e.g., web services, mail servers, custom protocols).

maoruiqi-hub:
# Add Oracle-compatible UTL_TCP package (plain outbound TCP core)

Fixes #1059

## What this adds

`UTL_TCP` in `contrib/ivorysql_ora`: an Oracle-style client for plain
outbound TCP, implemented as an `AUTHID CURRENT_USER` PL/iSQL package over
a small C module.  The public contract follows the Oracle 19c reference
for the operations covered here:

- `OPEN_CONNECTION`, `CLOSE_CONNECTION`, `CLOSE_ALL_CONNECTIONS`, `FLUSH`
- `WRITE_TEXT`, `WRITE_LINE`, `WRITE_RAW`
- `GET_TEXT`, `GET_LINE`, `GET_RAW`

The eight-field `CONNECTION` record keeps Oracle's field names and order
(`remote_host`, `remote_port`, `local_host`, `local_port`, `charset`,
`newline`, `tx_timeout`, `private_sd`); `private_sd` carries an opaque,
backend-local handle, never an OS socket descriptor, and the record is
otherwise descriptive.  `UTL_TCP.CRLF` is provided, and Oracle argument
names, order and defaults are preserved, including parameters whose
non-default values are unsupported (local bind, buffer sizes, wallets,
`peek`): those are rejected explicitly with `feature_not_supported`
instead of being ignored.

## Security (release condition)

Until #2053 supplies a shared invoking-user host/port `connect` check,
this version is deliberately restricted:

- `OPEN_CONNECTION` requires a superuser, checked in C before any name
  resolution or socket activity.
- PUBLIC has no EXECUTE on the package or on any of the new
  `sys.ora_utl_tcp_*` functions.
- The package is `AUTHID CURRENT_USER`, so a GRANT on the package alone
  cannot bypass the C-level checks; the TAP suite demonstrates that a
  directly granted C wrapper still denies a non-superuser, and that
  `SET ROLE` cannot use another role's handle.

## Implementation notes

- Sockets live in a bounded per-backend registry (max 50, matching
  Oracle) addressed by monotonic handles that are never reused while the
  backend lives.  Ownership is re-validated on every call, failed opens
  leave no slot or socket behind, `before_shmem_exit` closes whatever is
  still open, and connections outlive transactions like `UTL_FILE`'s.
- Connects are nonblocking.  The documented 30 second deadline bounds
  the TCP connect phase and starts **after address resolution**: DNS
  runs through the system resolver and is not bounded by, or
  cancellable through, this deadline.  Numeric-address connects are
  covered by timeout/cancellation tests; nothing here claims to bound
  DNS lookups.  Completion is verified via `SO_ERROR`; transfer
  timeouts bound each read/write operation (NULL = wait indefinitely
  but interruptibly, 0 = never wait, though a single nonblocking
  receive is still attempted for bytes that already arrived); all
  waiting goes through `WaitLatchOrSocket` so backend cancellation and
  postmaster death are handled.
- Text I/O converts between the database encoding and the charset given
  at open time, counts characters, never splits a multibyte character,
  and keeps incomplete trailing bytes buffered.  The `newline` sequence
  is converted to the wire encoding and validated at open time, before
  any socket or slot is allocated.  Raw I/O moves bytes unchanged.

## Review-fix round (commits after b6d52b8988)

A post-push review round fixed reproducible data-loss, timeout and
encoding defects found by re-running the suite and by loopback probes:

- The TAP harness could no longer hang or report a false PASS: its
  persistent `psql` sessions now use `background_psql` with real,
  attached timers, every wait is bounded, and capture-file assertions
  poll (bounded) for the server's asynchronous writes.  Running the
  previously unreached assertions also surfaced and fixed latent test
  regex bugs (literal backspace bytes, message-text mismatches).
- `GET_TEXT` no longer loses buffered input at the 32767-byte value
  limit: complete characters are staged and converted as a whole, the
  wire bytes are consumed only after the return value is built, and the
  longest fitting prefix is returned with the remainder kept buffered
  (8192 four-byte UTF8 characters come back as 8191 + 1; LATIN1-to-UTF8
  expansion behaves the same).
- With `tx_timeout => 0`, data that already arrived is readable by all
  three read entry points; an empty socket still raises `08006`
  immediately.
- `GET_LINE` follows Oracle's `READ_LINE` rules: a lone CR completes a
  line immediately (a following LF is swallowed as part of the same
  terminator instead of producing an empty line), and a timeout returns
  the buffered text as a partial line when it forms whole characters.
- Following Oracle's `READ_TEXT` rules, `GET_TEXT` raises
  `TRANSFER_TIMEOUT` instead of returning a partial value when the
  timeout hits while a character is incomplete; everything stays
  buffered for a retry, and a clean EOF still returns the complete
  characters (the partial one stays available to `GET_RAW`).
- The `newline` is converted to the wire encoding at open time, so a
  UTF8 database with a LATIN1 connection sends `E9` for `CHR(233)`
  instead of the UTF8 bytes; unconvertible or overlong newlines are
  rejected before any resource is allocated.  Negative
  `in_buffer_size`/`out_buffer_size` are rejected with `22023`, and any
  non-NULL `local_port` (including 0) with `0A000`.
- Windows: `pgwin32_send` waits indefinitely for writability unless
  called with `pgwin32_noblock`, which previously bypassed the transfer
  timeout and cancellation entirely for writes.  Every `send()` now
  sets/restores that flag around the call like the `recv()` path.
  **Windows behavior has not been runtime-verified** (no Windows
  environment was available); this is a static review against
  `pgwin32_send`'s would-block branches, with the Linux back-pressure
  tests passing.

## Deviations from Oracle's declaration (current PL/iSQL limits)

- `PLS_INTEGER` does not exist as a type on master; `INTEGER` (the same
  32-bit signed range) is used for every `PLS_INTEGER` field/parameter.
- A package constant cannot be referenced from a parameter `DEFAULT`
  (the default is resolved in the caller's scope at call time), so the
  `newline` default is spelled `CHR(13) || CHR(10)`, which is the value
  of `UTL_TCP.CRLF`.  Both were reproduced with minimal probes.

## Deferred to follow-ups

`AVAILABLE`, the `READ_*` OUT-parameter forms, NCHAR variants,
`SECURE_CONNECTION`/wallets/TLS, I/O buffering and `peek = TRUE` are not
implemented; no UTL_TCP-specific network ACL is introduced.

## Tests

`t/001_utl_tcp.pl` + `t/utl_tcp_server.pl` (TAP, registered for both
GNU Make and Meson) run against a test-owned loopback server with
scripted behaviors: echo, binary echo, hold (never reads), immediate
EOF, capture-to-file, fragmented CRLF, close-after-prefix, and push
(send-on-accept, echoing or closing for clean-EOF coverage).  Coverage:
record fields/defaults, named/positional calls, CRLF/custom newline
including wire-encoding conversion, text/raw round trips with NUL and
high bytes, multibyte conversion in both directions verified
byte-for-byte on the wire, fragmented reads, value-limit boundaries
without data loss, transfer timeouts (read and write side), ready data
at `tx_timeout => 0`, lone-CR/partial-line timeout rules, cancellation
of an unbounded wait, end of input, invalid host/port/options/handles,
the 50-slot bound with never-reused handles, failed opens consuming no
slots, privilege denials (package, direct C grant, `SET ROLE`), and
backend exit observed as EOF by the peer.  The suite forces an
Oracle-mode cluster in both harnesses and finds the Oracle-mode
listener with `SHOW ivorysql.port`.

Check matrix (Linux/WSL, gcc, --enable-cassert --enable-tap-tests
--with-libxml --with-uuid=e2fs), current head:

- `make -C contrib/ivorysql_ora oracle-check`: 29/29 Oracle regression
  tests plus the UTL_TCP TAP suite (134 subtests, no harness timeouts,
  single-digit seconds): PASS.
- `make -C src/pl/plisql/src oracle-check`: 22/22: PASS.
- Meson build plus `meson test ivorysql_ora/001_utl_tcp`: 134 subtests:
  PASS (the Meson registration needed the Oracle-mode Perl modules
  from `src/oracle_test/perl` to take precedence; fixed in this round).
- `git diff --check`: clean.
- Not run: Windows build/tests (no Windows environment; see the
  static-review note above) and a DNS-resolution timing claim (DNS is
  explicitly not covered by the 30 second guarantee).

Commits are split per the contribution guide (contract, C
implementation, tests, build registration plus one small wait-path fix
found by the new tests, followed by the review-fix round).  Per
coding-assistants.md this work was AI-assisted (Assisted-by tags in the
commits); the human submitter should add their own Signed-off-by.

Assisted-by: ZCode:GLM-5.3-Flash

## Summary by CodeRabbit

* **New Features**
  * Added Oracle-compatible `UTL_TCP` support for outbound TCP connections.
  * Supports connection management, text and binary reads/writes, line handling, timeouts, character-set conversion, and configurable newline formats.
  * Added security controls requiring superuser access and enforcing connection ownership.
  * Added validation for connection limits, handles, arguments, unsupported options, and connection timeouts.
* **Tests**
  * Added comprehensive coverage for networking, encoding, timeouts, errors, limits, permissions, and connection cleanup.


bigplaice:

maoruiqi-hub:
Rebased onto current `master` (`63fb0bfe615`) and force-pushed, so the branch is mergeable again.

The only conflict was the top-level `meson.build`: `meson: enable Oracle test suites` (c4471ee2189) now adds `-I src/oracle_test/perl` when a suite lives under `src/oracle_test/`, but this suite lives in `contrib/ivorysql_ora/`, so the `perl_incs` hook is kept on top of that logic instead of replacing it. Nothing else needed manual merging.

Re-verified at the pushed commit:

| Check | Result |
| --- | --- |
| `make -C contrib/ivorysql_ora oracle-check` | 31/31 Oracle regression tests + TAP `t/001_utl_tcp.pl` (134 subtests), PASS (~9 s) |
| `meson test -C build ivorysql_ora/001_utl_tcp` | 134 subtests, PASS |
| `git diff --check` | clean |

The limits documented in the description are unchanged: Windows is static-review only, and DNS resolution is explicitly not covered by the 30-second connect deadline.

@bigplaice would you have a chance to take a look when you get a moment? Thanks!

PR #2111, opened 2026-09-17
feat: allow CALL/EXEC procedures without parentheses

### LINKED ISSUE #2125: Support CALL/EXEC of argument-less procedures and functions without parentheses
## Feature Request
Allow CALL/EXEC procedures without parentheses

Oracle allows invoking procedures that take no arguments without trailing parentheses, using either CALL or EXEC. IvorySQL currently throws an error for these same statements, breaking Oracle compatibility for migrated code.

Example: Oracle (works):

```sql
CALL proc1;        -- Oracle succeeds
EXEC proc1;        -- Oracle succeeds
```
Example: IvorySQL (currently fails):

```sql
CALL proc1;        -- IvorySQL throws error
EXEC proc1;        -- IvorySQL throws error
```

**Expected Behavior**:
The statements above should execute successfully in IvorySQL without requiring () after the procedure name.

yasir-hussain-shah:
Allow CALL/EXEC procedures having no arguments without parentheses, just like Oracle.

- Previously a procedure invocation without a trailing argument list (CALL foo; or exec foo;) was a syntax error, even when they took no arguments. Oracle allows this. 

**Implementation details**:
- pl_gram.y: build_call_expr() now tracks whether the scanned call actually contained a '(' and appends an empty "()" when it did not, so the generated query text is always a valid CALL expression.
- ora_gram.y: route CALL through exec_func_application so the Oracle parser accepts the no-paren form consistently with the PLiSQL path.
- Update regression tests and expected output accordingly.

* **New Features**
  * Added support for Oracle-style procedure calls without parentheses when no arguments are required.
  * Extended no-argument procedure invocation support in top-level statements and `DO` blocks.
  * Preserved support for parenthesized calls and calls with arguments, including `INTO` and optional ordering clauses.

Fixes: https://github.com/IvorySQL/IvorySQL/issues/2125


bigplaice:

src/oracle_test/regress/expected/ora_plisql.out
```diff
@@ -1007,15 +1007,13 @@ raise notice 'protest';
...
+NOTICE:  protest
 CALL protest();
 NOTICE:  protest
 exec protest;      
-ERROR:  syntax error at end of input
+ERROR:  syntax error at or near ";"
 LINE 1: BEGIN protest;      ; END
-                          ^
+                            ^
 QUERY:  BEGIN protest;      ; END
```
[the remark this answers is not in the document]
yasir-hussain-shah: This is not regression of my change. I didn't touch `EXEC` command's flow.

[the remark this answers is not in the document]
bigplaice: @yasir-hussain-shah 

This is an existing issue in src/bin/psql/mainloop.c 

"exec protest; " is packed as anonymous block by psql and so becomes: "BEGIN protest; END;"
But the current code checks if the last character is ';' without removing the trailing spaces. 
If there are trailing spaces like "exec protest;   ", the translated statement becomes: "BEGIN protest;   ; END;". 
The second null statement will cause error. 

I will create an issue to address it. 

As a workaround, could you please remove the trailing spaces from ora_plisql.out so that "exec protest;" can be executed successfully? 

Thanks,
Steven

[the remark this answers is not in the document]
yasir-hussain-shah: @bigplaice Done!

src/oracle_test/regress/expected/ora_plisql.out
```diff
@@ -1284,9 +1272,7 @@ end;
 /
 -- fail
```
bigplaice, line 1273: Now the "fail" shoule be "ok"

src/oracle_test/regress/expected/ora_plisql.out
```diff
@@ -1301,9 +1287,7 @@ end;
 variable x number;
 -- fail
```
bigplaice, line 1288: Now the "fail" shoule be "ok"

src/oracle_test/regress/expected/ora_plisql.out
```diff
@@ -1321,9 +1305,7 @@ end;
 /
 -- fail
```
bigplaice, line 1306: Now the "fail" shoule be "ok"

src/oracle_test/regress/expected/ora_plisql.out
```diff
@@ -1341,9 +1323,7 @@ end;
 variable x number
 --fail
```
bigplaice, line 1324: Now the "fail" shoule be "ok"

PR #2118, opened 2026-09-18
Fix ivorysql_ora SQL generation for out-of-tree builds

### LINKED ISSUE #2117: ivorysql_ora SQL generation depends on the build directory location
### What happened?

The `ivorysql_ora` SQL generator assumes a particular working directory. Meson invokes `../contrib/ivorysql_ora/gensql.pl`, and the script opens its manifest and SQL fragments through the same relative prefix. This works with a build directory immediately below the source root, but prevents SQL generation in an external or more deeply nested build directory.

Using the absolute script path only gets past the first problem: the generator still fails to open `ivorysql_ora_merge_sqls` from an external directory. Make mode similarly looks for its inputs in the working directory.

The generator should locate the manifest and SQL fragments relative to its source location, regardless of the build directory. Meson should capture the generated SQL in its target file, and Make mode should continue to write output in the current build directory.

### How can we reproduce it?

From a checkout of current `master`, with Perl installed:

```sh
src=$(pwd)
build=$(mktemp -d)
(
  cd "$build"
  perl "$src/contrib/ivorysql_ora/gensql.pl" meson 1.0
)
```

Actual result: nonzero exit with `Could not open file ivorysql_ora_merge_sqls: No such file or directory`.

With the normal Meson build dependencies installed, the full target can be reproduced using an external build directory:

```sh
meson setup "$build/meson" "$src" -Dtap_tests=disabled
ninja -C "$build/meson" contrib/ivorysql_ora/ivorysql_ora--1.0.sql
```

The relative script path in the custom target is also used by `contrib/gb18030_2022/meson.build`.

### IvorySQL version

`master`, commit `63fb0bfe6159fb91b03012eea09f3db7cd46db8c`.

### Compatible mode / server encoding

Not applicable: this happens during SQL generation, before starting a database server.

### OS version

macOS 26.6.2, arm64; Perl 5.34.1, Meson 1.12.0, Ninja 1.13.2.

### Before submitting

- [x] Reproduced against the current `master` commit above.
- [x] Searched existing issues and open PRs; no duplicate fix for this path issue found.

pilichoumao:
Fixes #2117.

An external or more deeply nested Meson build cannot generate the Oracle extension SQL: both the custom target command and `gensql.pl` assume that the working directory is immediately below the source root.

Resolve the generator through Meson's `files()` in both existing callers, and resolve its manifest and SQL fragments relative to the script with `FindBin`. Make mode still writes its output to the current build directory. The generated SQL and the existing GB18030 target are otherwise unchanged.

Add a TAP regression test for installation and upgrade SQL, including an absent upgrade fragment, in both output modes. It checks complete SQL contents and output placement from the source directory and a nested external directory, with spaces in the paths. Register it with Meson; the Make TAP harness already discovers tests in this directory.

Validation on macOS arm64:

- `prove -v contrib/ivorysql_ora/t/002_gensql.pl`: all 18 assertions pass. The same test fails 13 assertions with the original generator.
- `meson test -C "$build" --no-rebuild --print-errorlogs --suite ivorysql_ora`: passes with TAP tests enabled.
- Both existing Ninja generator targets succeed in an external build directory containing a space.
- `make -C "$make_build/contrib/ivorysql_ora" gensql` succeeds in a separate out-of-tree Make build.
- The real 439,890-byte extension SQL is byte-for-byte identical between Meson, Make, and the original generator run from its supported directory layout.
- `git diff --check`: passes.

A full Meson build was attempted. It stops at the existing `libpq_check.pl` check because `libpq` references `_exit`; the unmodified `sysv_ivysema.c` contains the corresponding `exit(-1)` calls. Continuing with Ninja's keep-going mode built the remaining targets, but the overall build still fails that check. The full database regression suite and `make installcheck-world` have not been run. The Meson TAP command above uses `--no-rebuild` to run this server-independent test after that build attempt.

Assisted-by: OpenAI:GPT-6

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SQL generation reliability when run from different source and build directories.
  * Ensured generated SQL files and included inputs are resolved consistently relative to the generator script.

* **Tests**
  * Added automated coverage for SQL generation across supported invocation modes and directory layouts.
  * Added validation for versioned SQL outputs, including temporary paths containing spaces.


bigplaice:

PR #2124, opened 2026-09-18
Fix libpq_pipeline oracle mirror: register protocol_version test

### LINKED ISSUE #2123: oracle_test libpq_pipeline omits protocol_version test registration (skips test; breaks -Werror build)
## Summary

The Oracle mirror of the `libpq_pipeline` TAP module — `src/oracle_test/modules/libpq_pipeline/libpq_pipeline.c` — never registers the `protocol_version` test. The test function is fully present but is dead code, which:

1. silently skips the test (the Oracle test list has **12** entries vs. **13** in the PG parallel module), and
2. breaks a strict (`-Werror`) build with `-Wunused-function`, stopping before any test runs.

## Repro

```sh
make -C src/oracle_test/modules/libpq_pipeline
./src/oracle_test/modules/libpq_pipeline/libpq_pipeline tests    # 12 entries, no protocol_version

make -C src/test/modules/libpq_pipeline
./src/test/modules/libpq_pipeline/libpq_pipeline tests           # 13 entries, includes protocol_version
```

Strict build fails:

```sh
COPT=-Werror make -C src/oracle_test oracle-check
```

```
libpq_pipeline.c:1325:1: error: 'test_protocol_version' defined but not used [-Werror=unused-function]
cc1: all warnings being treated as errors
```

## Root cause

Relative to `src/test/modules/libpq_pipeline/libpq_pipeline.c`, the Oracle mirror is missing exactly two lines:

- `printf("protocol_version\n");` in `print_test_list()` (PG: line 2127)
- `else if (strcmp(testname, "protocol_version") == 0) test_protocol_version(conn);` in `main()` (PG: lines 2240–2241)

The `test_protocol_version()` body (line 1325) was synced from PG, but its registration and dispatch were not, leaving it unreachable and therefore unused. The TAP driver `t/001_libpq_pipeline.pl` iterates over the output of `libpq_pipeline tests`, so a test absent from the list is never executed.

## Expected

The Oracle mirror should list and dispatch the same `protocol_version` test as the PG module, and a `-Werror` build should succeed.

## Proposed fix

Add the two missing lines to match the PG parallel module. A PR with this fix accompanies this issue.

R0CKing666:
Fixes #2123

## What

The Oracle mirror of the `libpq_pipeline` TAP module never registered the
`protocol_version` test. `test_protocol_version()` was synced from PG but left
unreachable, so the TAP driver silently skipped it and a `-Werror` build failed
with `-Wunused-function`.

This adds the two missing lines to
`src/oracle_test/modules/libpq_pipeline/libpq_pipeline.c` to match the PG
parallel module `src/test/modules/libpq_pipeline/libpq_pipeline.c`:

- `printf("protocol_version\n");` in `print_test_list()`
- `else if (strcmp(testname, "protocol_version") == 0) test_protocol_version(conn);` in `main()`

## Verification

Built both modules in an Ubuntu 22.04 (GCC 11.4) container:

- `./libpq_pipeline tests` now lists **13** entries on both sides (previously 12 vs 13), and the two lists are now identical.
- `make -C src/oracle_test/modules/libpq_pipeline COPT=-Werror` now succeeds (previously failed with `libpq_pipeline.c:1325:1: error: 'test_protocol_version' defined but not used [-Werror=unused-function]`).

## Summary by CodeRabbit

- **Tests**
  - Added the `protocol_version` test to the available test list.
  - Added support for running the protocol version test by specifying its test name.


NotHimmel:

